PR #6, opened 2017-02-28
added rsa error handling

zhengping12:


lints/lint_rsa_exp_negative.go
```diff
@@ -22,7 +22,11 @@ func (l *rsaExpNegative) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rsaExpNegative) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	pubKey := c.PublicKey.(*rsa.PublicKey).E
+	key, found := c.PublicKey.(*rsa.PublicKey)
+	if !found {
+		return ResultStruct{Result: Error}, nil
+	}
+	pubKey := key.E
```
zakird: This is not the public key. This is the exponent. Please change variable name.

lints/lint_old_root_ca_rsa_mod_less_than_2048_bits.go
```diff
@@ -26,7 +26,11 @@ func (l *rootCaModSize) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rootCaModSize) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	mod := c.PublicKey.(*rsa.PublicKey).N
+	key, found := c.PublicKey.(*rsa.PublicKey)
```
zakird: In golang, it's generally accepted practice to use the variable name `ok` instead of found. I'd prefer to use this:

`key, ok := c.PublicKey.(*rsa.PublicKey)`

lints/lint_old_sub_ca_rsa_mod_less_than_1024_bits.go
```diff
@@ -25,7 +25,11 @@ func (l *subCaModSize) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *subCaModSize) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	mod := c.PublicKey.(*rsa.PublicKey).N
+	key, found := c.PublicKey.(*rsa.PublicKey)
+	if !found {
+		return ResultStruct{Result: Error}, nil
+	}
+	mod := key.N
```
zakird: Not really sure why this is a variable.

lints/lint_old_sub_cert_rsa_mod_less_than_1024_bits.go
```diff
@@ -24,7 +24,11 @@ func (l *subModSize) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *subModSize) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	mod := c.PublicKey.(*rsa.PublicKey).N
+	key, found := c.PublicKey.(*rsa.PublicKey)
+	if !found {
+		return ResultStruct{Result: Error}, nil
+	}
+	mod := key.N
```
zakird: Same here. Not sure why this deserves a variable.

lints/lint_rsa_mod_factors_smaller_than_752.go
```diff
@@ -26,7 +26,11 @@ func (l *rsaModSmallFactor) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rsaModSmallFactor) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	mod := c.PublicKey.(*rsa.PublicKey).N
```
zakird: This file name should probably be `lints/lint_rsa_mod_factors_smaller_than_752_bits.go` to match how all the other files are named.

lints/lint_old_root_ca_rsa_mod_less_than_2048_bits.go
zakird: Also, why do we have this lint? This seems to be a strict subset of `lints/lint_rsa_mod_less_than_2048_bits.go`.

lints/lint_rsa_public_exponent_not_in_range.go
```diff
@@ -27,7 +27,11 @@ func (l *rsaParsedTestsExpInRange) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rsaParsedTestsExpInRange) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	pubKey := c.PublicKey.(*rsa.PublicKey).E
+	key, found := c.PublicKey.(*rsa.PublicKey)
+	if !found {
+		return ResultStruct{Result: Error}, nil
+	}
+	pubKey := key.E
```
zakird: This is an exponent, not a public key.

lints/lint_rsa_public_exponent_not_odd.go
```diff
@@ -26,8 +26,11 @@ func (l *rsaParsedTestsKeyExpOdd) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rsaParsedTestsKeyExpOdd) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	theKey := c.PublicKey.(*rsa.PublicKey)
-	if theKey.E%2 == 1 {
+	key, found := c.PublicKey.(*rsa.PublicKey)
+	if !found {
+		return ResultStruct{Result: Error}, nil
+	}
+	if key.E%2 == 1 {
```
zakird: Has go lint been run on this? I'm surprised this isn't `if key.E % 2 == 1 {`

lints/lint_old_root_ca_rsa_mod_less_than_2048_bits.go
zhengping12: I'll change it to ok.
If you read the description parts in the lints. The former is "In a validity period beginning on or before 31 dec 2010, root CA certificates using RSA public key algorithm must  have 2048 bits of modulus". While the latter is "in validity period beginning after 31 Dec 2010, all certificates using RSA public key algorithm must have 2048 bits of modulus".

zakird: If that is the case, then that lint is not correctly defined. Nowhere in `lints/lint_rsa_mod_less_than_2048_bits.go`, do you check for the validity period.

zakird: Oh! I see, there's an effective date. NVM last comment.

zakird: This is super confusing in general though. Because it looks like `lint_old_root_ca_rsa_mod_less_than_2048_bits.go` checks this in the `CheckApplies` function. Whereas `lint_rsa_mod_less_than_2048_bits.go` defines this in `EffectiveDate`. Is there a reason this isn't done in the same manner for both lints?

zakird:
I mentioned this to @kumarde out of channel, but we also need to be returning a different value for when the lint failed and for when an error occurred while executing the lint. Deepak, can you provide the right values that should be used in both cases here?

kumarde:
Yup, @zhengping12 we should use the Fatal result when something just goes wrong or errors out, and return Error only if the lint doesn't succeed.

lints/lint_old_sub_cert_rsa_mod_less_than_1024_bits.go
zhengping12: fixed

lints/lint_old_sub_ca_rsa_mod_less_than_1024_bits.go
zhengping12: fixed

lints/lint_rsa_mod_factors_smaller_than_752.go
zhengping12: fixed

lints/lint_rsa_public_exponent_not_in_range.go
zhengping12: fixed.

lints/lint_rsa_public_exponent_not_odd.go
zhengping12: It actually doesn't complain, but I fixed it.

zhengping12:
Ok, we'll fix that after the next few PRs get approved.

lints/lint_rsa_exp_negative.go
zhengping12: fixed

zhengping12: We don't have a lint for that situation.

lints/lint_rsa_mod_not_odd.go
```diff
@@ -27,9 +27,12 @@ func (l *rsaParsedTestsKeyModOdd) CheckApplies(c *x509.Certificate) bool {
...
 
 func (l *rsaParsedTestsKeyModOdd) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	theKey := c.PublicKey.(*rsa.PublicKey)
+	key, ok := c.PublicKey.(*rsa.PublicKey)
+	if !ok {
+		return ResultStruct{Result: Error}, nil
+	}
 	z := big.NewInt(0)
-	if (z.Mod(theKey.N, big.NewInt(2)).Cmp(big.NewInt(1))) == 0 {
+	if (z.Mod(key.N, big.NewInt(2)).Cmp(big.NewInt(1))) == 0 {
```
zakird: This should also check that the exponent isn't one.

lints/lint_rsa_public_exponent_too_small.go
```diff
@@ -26,8 +26,11 @@ func (l *rsaParsedTestsExpBounds) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rsaParsedTestsExpBounds) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	theKey := c.PublicKey.(*rsa.PublicKey)
-	if theKey.E >= 3 { //If Cmp returns 1, means N > E
+	key, ok := c.PublicKey.(*rsa.PublicKey)
```
zakird: Can't this lint be merged with `lint_rsa_public_exponent_not_in_range.go` ?

lints/lint_rsa_public_exponent_not_in_range.go
```diff
@@ -27,12 +27,16 @@ func (l *rsaParsedTestsExpInRange) CheckApplies(c *x509.Certificate) bool {
...
+	if !ok {
+		return ResultStruct{Result: Error}, nil
+	}
+	exponent := key.E	
 	lowerBound := 65536 // 2^16 + 1
 	var upperBound big.Int
 	upperBound.Exp(big.NewInt(2), big.NewInt(256), nil)
 
-	if pubKey > lowerBound && upperBound.Cmp(big.NewInt(int64(pubKey))) == 1 {
+	if exponent > lowerBound && upperBound.Cmp(big.NewInt(int64(exponent))) == 1 {
```
zakird: I'm super confused by this lint now that I read it. IIUC, this will fail if exponent is < 65536? There are lots of perfectly fine exponents lower than that. Like 3.

lints/lint_rsa_public_exponent_not_in_range.go
```diff
@@ -27,12 +27,16 @@ func (l *rsaParsedTestsExpInRange) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rsaParsedTestsExpInRange) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	pubKey := c.PublicKey.(*rsa.PublicKey).E
+	key, ok := c.PublicKey.(*rsa.PublicKey)
+	if !ok {
+		return ResultStruct{Result: Error}, nil
+	}
+	exponent := key.E	
 	lowerBound := 65536 // 2^16 + 1
```
zakird: This should be marked const.

lints/lint_rsa_public_exponent_not_in_range.go
```diff
@@ -27,12 +27,16 @@ func (l *rsaParsedTestsExpInRange) CheckApplies(c *x509.Certificate) bool {
...
 	var upperBound big.Int
 	upperBound.Exp(big.NewInt(2), big.NewInt(256), nil)
```
zakird: Why is this being run in every test? This should be run once during initialization and saved.

lints/lint_rsa_mod_not_odd.go
```diff
@@ -27,9 +27,12 @@ func (l *rsaParsedTestsKeyModOdd) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *rsaParsedTestsKeyModOdd) RunTest(c *x509.Certificate) (ResultStruct, error) {
-	theKey := c.PublicKey.(*rsa.PublicKey)
+	key, ok := c.PublicKey.(*rsa.PublicKey)
+	if !ok {
+		return ResultStruct{Result: Error}, nil
+	}
 	z := big.NewInt(0)
```
zakird: It look like `z := big.NewInt(0)` and `big.NewInt(1)` should both be called in init.

zakird:
More comments. Please also add the Lint I requested.

lints/lint_rsa_public_exponent_too_small.go
zhengping12: This is stated in CAB BR 6.1.6
RSA: The CA SHALL confirm that the value of the public exponent is an odd number equal to 3 or more. 
Additionally, the public exponent SHOULD be in the range between 2^16+1 and 2^256‐1

lints/lint_rsa_public_exponent_not_in_range.go
zhengping12: This is stated in CAB BR 6.1.6
RSA: The CA SHALL confirm that the value of the public exponent is an odd number equal to 3 or more.
Additionally, the public exponent SHOULD be in the range between 2^16+1 and 2^256‐1

zhengping12:
By adding the lint, do you mean checking if the the cert has no public key?

zakird:
yeah.

zakird:
I'm mostly afraid if we don't add it now, we'll lose track of it.

zhengping12:
@zakird Hi. I've added the new lint, and made minor changes according to your suggestions.

zakird:
I'll take a look. However, I do note that tests currently fail, and we won't be able to merge until that is resolved.

zakird:
Uff. Can you move the `go fmt` into its own PR?

PR #7, opened 2017-03-01
only the rsa related changes

zhengping12:
All the *_test.go files have one empty newline removed due go fmt. Rest of the files contain actual rsa related fixes.

### DIFF
--- a/lints/base.go
+++ b/lints/base.go
@@ -2,12 +2,14 @@ package lints
 
 import (
 	"github.com/zmap/zgrab/ztools/x509"
+	"math/big"
 	"time"
 )
 
 // global
 var (
-	Lints map[string]*Lint = make(map[string]*Lint)
+	Lints      map[string]*Lint = make(map[string]*Lint)
+	upperBound                  = &big.Int{}
 )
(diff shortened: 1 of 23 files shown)


lints/base.go
```diff
@@ -2,12 +2,14 @@ package lints
...
 	"github.com/zmap/zgrab/ztools/x509"
+	"math/big"
 	"time"
 )
 
 // global
 var (
-	Lints map[string]*Lint = make(map[string]*Lint)
+	Lints      map[string]*Lint = make(map[string]*Lint)
+	upperBound                  = &big.Int{}
```
zakird: I'm really not sure why this is global across the entire zlint. Why can't this be stored on the rsaParsedTestsExpInRange object? AFAICT that is the only place this is used. Each lint has its own Initialization function.

zakird: I'm really not sure why this is global across the entire zlint. Why can't this be stored on the rsaParsedTestsExpInRange object?

zhengping12: You're right, that's better. Fixed now.

PR #11, opened 2017-03-10
rsa related lints changes.

zhengping12:
Lints involving RSA public key now check whether certs have public keys in the CheckApply function.

### DIFF
--- a/lints/lint_old_root_ca_rsa_mod_less_than_2048_bits.go
+++ b/lints/lint_old_root_ca_rsa_mod_less_than_2048_bits.go
@@ -22,7 +22,7 @@ func (l *rootCaModSize) Initialize() error {
 func (l *rootCaModSize) CheckApplies(c *x509.Certificate) bool {
 	issueDate := c.NotBefore
 	_, ok := c.PublicKey.(*rsa.PublicKey)
-	return (c.PublicKeyAlgorithm == x509.RSA && util.IsRootCA(c) && issueDate.Before(util.RsaDate2)) && ok
+	return ok && (c.PublicKeyAlgorithm == x509.RSA && util.IsRootCA(c) && issueDate.Before(util.RsaDate2))
 }
 
 func (l *rootCaModSize) RunTest(c *x509.Certificate) (ResultStruct, error) {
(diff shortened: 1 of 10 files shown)


lints/lint_old_root_ca_rsa_mod_less_than_2048_bits.go
```diff
@@ -21,14 +21,12 @@ func (l *rootCaModSize) Initialize() error {
 
 func (l *rootCaModSize) CheckApplies(c *x509.Certificate) bool {
 	issueDate := c.NotBefore
-	return (c.PublicKeyAlgorithm == x509.RSA && util.IsRootCA(c) && issueDate.Before(util.RsaDate2))
+	_, ok := c.PublicKey.(*rsa.PublicKey)
+	return (c.PublicKeyAlgorithm == x509.RSA && util.IsRootCA(c) && issueDate.Before(util.RsaDate2)) && ok
```
zakird: I'm not sure if I agree with this. Not sure that I don't either though. The check _does_ apply, it's just a malformed certificate. That said, if another test checks for it, then, I could also see why we wouldn't say it's relevant. @kumarde how are we handling this in general?

zakird: If we are keeping this form, then the ok should be the first check in the and statement, or even more preferably, as separate statements.

kumarde: @zakird There's a lint that checks whether or not a public key exists already, so rather than returning a fatal when there's no public key, I think it's fine if we just don't say it's relevant for all the other RSA checks rather than returning fatal

zakird: OK sounds good to me.

lints/lint_old_sub_ca_rsa_mod_less_than_1024_bits.go
```diff
@@ -20,14 +20,12 @@ func (l *subCaModSize) Initialize() error {
 func (l *subCaModSize) CheckApplies(c *x509.Certificate) bool {
 	issueDate := c.NotBefore
 	endDate := c.NotAfter
-	return (util.IsSubCA(c) && issueDate.Before(util.RsaDate2) && endDate.Before(util.RsaDate3))
+	_, ok := c.PublicKey.(*rsa.PublicKey)
+	return (util.IsSubCA(c) && issueDate.Before(util.RsaDate2) && endDate.Before(util.RsaDate3)) && ok
```
zakird: Please change these to all be of the following form:
`ok && util.IsSubCA(c) && issueDate.Before(util.RsaDate2) && endDate.Before(util.RsaDate3)`

zhengping12:
@zakird fixed.

zakird:
It still looks like there are unnecessary parens in there

zhengping12:
Sorry they were there before. Fixed now.

PR #19, opened 2017-03-24
fixing bug in ip checking for common names

kumarde:
string(net.IP) is not correct.


zakird:
Is this correct? Aren't there multiple correct stringified versions of an IPv6 address?

PR #20, opened 2017-03-28
updating FQDN parsing in ZLint

kumarde:
Matching FQDN parsing in ZGrab, removing regular expression stuff.


lints/lint_ext_san_uri_host_not_fqdn_or_ip.go
```diff
@@ -32,9 +32,9 @@ func (l *SANURIHost) CheckApplies(c *x509.Certificate) bool {
 
 func (l *SANURIHost) RunTest(c *x509.Certificate) (ResultStruct, error) {
 	for _, uri := range c.URIs {
-		auth := util.GetAuthority(uri)
-		if auth != "" {
-			host := util.GetHost(auth)
+		//auth := util.GetAuthority(uri)
```
zakird: This looks like it should be removed.

util/fqdn.go
```diff
@@ -2,13 +2,18 @@ package util
...
 	"net"
-	"regexp"
 	"strings"
+	"github.com/asaskevich/govalidator"
 )
 
 func IsFQDN(domain string) bool {
-	RegExp := regexp.MustCompile(`^((([a-zA-Z]{1})|([a-zA-Z]{1}[a-zA-Z]{1})|([a-zA-Z]{1}[0-9]{1})|([0-9]{1}[a-zA-Z]{1})|([a-zA-Z0-9][a-zA-Z0-9-_]{1,61}[a-zA-Z0-9])|\*)\.){2,}([a-zA-Z]{2,6}|[a-zA-Z0-9-]{2,30}\.[a-zA-Z]{2,3})(\.)*$`)
-	return RegExp.MatchString(domain)
+	var ret bool
```
zakird: Is this correct? Can you have `?.?.a.b.c`? If so, this doesn't look like it will work.

kumarde: I don't think you can have that, or it's at the least undefined behavior. I can't seem to find it in certificates either. We can add a check for it but it looks like it wasn't a big deal in this PR. https://github.com/zmap/zgrab/pull/188. thoughts?

zakird: Sure looks like it's possible ;)

kumarde: (╯°□°）╯︵ ┻━┻

kumarde: by this logic, i assume ?.?.?.a.b.c also probably works.

kumarde: or any number of question marks? for that matter? sad. okay. I'll fix it.

kumarde:
@zakird added fix - it's a little annoying but let me know what you think!

zakird:
look good. can you confirm you've run `go fmt` on this?

PR #21, opened 2017-03-28
updating SAN dependent lints

kumarde:


lints/lint_subject_common_name_disallowed.go
```diff
@@ -24,7 +24,7 @@ func (l *BadCommonName) Initialize() error {
 }
 
 func (l *BadCommonName) CheckApplies(c *x509.Certificate) bool {
-	return len(c.Subject.CommonName) != 0
+	return len(c.Subject.CommonName) != 0 && util.IsExtInCert(c, util.SANOID)
```
zakird: This seems like it should actually fail if you consider what the point of this test.

lints/lint_subject_common_name_not_from_san.go
```diff
@@ -22,7 +22,7 @@ func (l *subjectCommonNameNotFromSAN) Initialize() error {
 }
 
 func (l *subjectCommonNameNotFromSAN) CheckApplies(c *x509.Certificate) bool {
-	return c.Subject.CommonName != ""
+	return c.Subject.CommonName != "" && util.IsExtInCert(c, util.SANOID)
```
zakird: Also feels like this one should fail.

kumarde:
Closing as not the right thing to do.

PR #25, opened 2017-04-14
fixing bad hostname checking for URIs

kumarde:


zakird:
Can you please provide a description of what this is actually fixing?

zakird:
Also, you broke certificates in this PR. Please fix.

kumarde:
The previous version of this lint was checking for the existence of "@" exclusively when parsing the hostname of a URI, which would then fail on something like http://startcom.com, which is clearly a correct URI. This change uses the net/url library to extract the host from the URI to perform the lint.

PR #27, opened 2017-04-16
Added one DNS related lint. Fixed magic numbers in DNS related lints.

zhengping12:


lints/lint_ext_san_dns_name_too_long.go
```diff
@@ -0,0 +1,38 @@
...
+		if len(dns) > 253 {
+			return ResultStruct{Result: Error}, nil
+		}
+	}
+	return ResultStruct{Result: Pass}, nil
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_ext_san_dns_name_underscore",
```
zakird: This name seems to be unrelated to the lint.

lints/lint_ext_san_dns_name_too_long.go
```diff
@@ -0,0 +1,38 @@
...
+		Description:   "DNSName must be less than or equal to 253 bytes",
+		Providence:    "aswlabs certlint",
```
zakird: This doesn't look like a valid providence.

lints/lint_ext_san_dns_not_ia5_string.go
```diff
@@ -41,21 +41,22 @@ func (l *SANDNSNotIA5String) RunTest(c *x509.Certificate) (ResultStruct, error)
 	var seq asn1.RawValue
 	var err error
 	if _, err = asn1.Unmarshal(value, &seq); err != nil {
-		return ResultStruct{Result: NA}, err
+		return ResultStruct{Result: Fatal}, err
```
zakird: This looks like duplicate code from above. Please create a helper function.

zhengping12:
Fixed code according to request. Also, for the providence part, I don't really know of an efficient way to find real providence, nor do I know if such providence exists outside of awslabs certlint. I'd like to do this as the last step of the zlint project if possible.

zakird:
Please run `go fmt` on this code. Can you please figure out how to have your environment do this? I've asked for this close to a dozen times now.

zakird:
@kumarde once you've done a pass on this, I'll take a look,.

zhengping12:
@zakird I've ran gofmt on all of them, which file in particular are you not happy with?

zakird:
I haven't gone through in depth, but I can immediately tell you that that is false. To start, `lints/lint_ext_san_dns_not_ia5_string.go` has multiple indentation styles in the same file.

zhengping12:
I'm sorry but I can neither see it by eyeballing, nor does "gofmt" or "go fmt" give me a different file.

zakird:
Please add a test to this repository that runs `go fmt` on all of the files.

util/dns_string_format.go
```diff
@@ -0,0 +1,41 @@
+package util
+
+import (
+	"encoding/asn1"
+	"github.com/zmap/zcrypto/x509"
+)
+
+func DNSHasNonStringIA5(c *x509.Certificate, isIssuer bool) (bool, error) {
+	oid := SANOID
```
kumarde: Don't we have a DNSNames field for both SAN and IAN extensions parsed out now? Why are we doing asn.1 marshalling here?

zhengping12: Only some types are parsed. This way we can examine all types of values.

dadrian:
Merged in https://github.com/zmap/zlint/pull/71

PR #31, opened 2017-04-20
Added new lints for multiple RDN attributes

zhengping12:


lints/lint_issuer_multiple_attr_in_rdn.go
```diff
@@ -0,0 +1,45 @@
...
+		}
+	}
+	return ResultStruct{Result: Pass}, nil
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "w_issuer_multiple_attr_in_rdn",
+		Description:   "Certificates should not have multiple attributes in a single RDN in the subject name",
+		Providence:    "aswlabs certlint",
```
kumarde: Can you make these all awslabs in another PR next (asw is not a thing)

lints/lint_issuer_multiple_attr_in_rdn.go
```diff
@@ -0,0 +1,45 @@
+// lint_issuer_multiple_attr_in_rdn.go
+
```
kumarde: Can you add exactly what this lint is supposed to do?

lints/lint_issuer_multiple_attr_in_rdn.go
```diff
@@ -0,0 +1,45 @@
...
+func (l *IssuerRDNHasMultipleAttribute) Initialize() error {
+	return nil
+}
+
+func (l *IssuerRDNHasMultipleAttribute) CheckApplies(c *x509.Certificate) bool {
+	return true
+}
+
+func (l *IssuerRDNHasMultipleAttribute) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	var issuer pkix.RDNSequence
```
kumarde: If the RDNSequence is important and used in a number of lints we should parse it out in x509.go

lints/lint_issuer_multiple_attr_in_rdn.go
zhengping12: Basically all I'm PRing are from Tab2, or awslabs certlint. Also do you know of any efficient way to find the real providence?

lints/lint_issuer_multiple_attr_in_rdn.go
zhengping12: It's described in the description section :)

lints/lint_issuer_multiple_attr_in_rdn.go
zhengping12: It's already parsed, but pkix.go discards any attribute beyond the first. We don't want to store anything beyond the first, because they are not adhering to the standard.

lints/lint_issuer_multiple_attr_in_rdn.go
kumarde: Nah I was just referring to the typo. We need to at some point look through and see where the awslabs ones are coming from, which will probably just be parsing the BR or RFC.

lints/lint_issuer_multiple_attr_in_rdn.go
kumarde: So is it the case that anything after the first one is a violation? As it stands it looks like you're checking if each RDN has a length > 1, not the other way around.

kumarde:
Closing. Handling in #59.

PR #33, opened 2017-04-20
Added gofmt test for all source code

zhengping12:
This test should fail due to multiple files not currently properly formatted.


kumarde:
The codebase unironically fails the `gofmt` test.

zhengping12:
Ran gofmt on the whole project, the test passes now. @kumarde

PR #35, opened 2017-04-24
Kumarde/zlint refactor

kumarde:
Based on a conversation w/ @zakird , we want to keep tracking of lints in a struct rather than a dynamic map so we catch errors at compile time rather than runtime. 

Looking for feedback on the best way to do this. I don't want to couple the lint code to a specific struct field, so right now I'm just checking the name and updating the appropriate field, but not sure if there's a better to handle this.

### DIFF
--- a/util/fqdn_test.go
+++ b/util/fqdn_test.go
@@ -31,7 +31,7 @@ func TestIsFQDNQuestionMarkFQDN(t *testing.T) {
 }
 
 func TestIsFQDNQuestionMarkIncorrectPlaceFQDN(t *testing.T) {
-	domain := "?.?.abc.?com"
+	domain := "?.?.abc?.com"
 	expected := false
 	actual := IsFQDN(domain)
 	if expected != actual {
--- a/lints/base.go
+++ b/lints/base.go
@@ -10,12 +10,12 @@ var (
 	Lints map[string]*Lint = make(map[string]*Lint)
 )
 
-const ZLintVersion = 1;
+const ZLintVersion = 1
 
 type ZLintResult struct {
-	ZLintVersion	int64
-	ZLints		map[string]string
-	Timestamp 	int64
+	ZLintVersion int64
+	ZLints       map[string]string
+	Timestamp    int64
 }
 
 type LintTest interface {
--- a/lints/base.go
+++ b/lints/base.go
@@ -10,201 +10,214 @@ var (
 	Lints map[string]*Lint = make(map[string]*Lint)
 )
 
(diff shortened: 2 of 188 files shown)


zakird:
This really doesn't seem like it solves any problems, it just moves them
around. But I understand how we arrived here too based on how lints are
registered. Otherwise, we have to do this in a much more static method. But
given that the protobuf is static, it's not like we're gaining anything by
having run time registration, as evidenced by that function....

I'm not 100% sure how one is supposed to do this in golang. @dadrian do you
have thoughts?

zakird:
It's also worth noting that his solution is probably slower. It's O(n^2)
whereas the old solution was O(n) with out any downfalls compared to this
solution.

kumarde:
@zakird yeah you're right. I could use a static map I guess to do this, but I wonder if that's _really_ better than the old solution.

dadrian:
I don't fully have enough context, what type of errors do you want caught at compile-vs-run time? What is the problem we're trying to solve?

kumarde:
@dadrian when ZLint is run over a certificate, it currently returns a `map[string]string`, indexed by lint_name. @zakird mentioned to me that's dangerous because we won't know at compile time whether or not indexing into that map will fail, and so it's better to expose a struct that requires the calling code to directly reference fields rather than indexing into a map. Unfortunately, we register lints at runtime and in order to do that, we just need to do the string compare in the ZLint library rather than the calling application.

So the question: do you have any suggestions on how to do this idiomatically in Golang?

util/fqdn_test.go
```diff
@@ -31,7 +31,7 @@ func TestIsFQDNQuestionMarkFQDN(t *testing.T) {
 }
 
 func TestIsFQDNQuestionMarkIncorrectPlaceFQDN(t *testing.T) {
-	domain := "?.?.abc.?com"
+	domain := "?.?.abc?.com"
```
zakird: These seem equally incorrect. wat.

kumarde: I think govalidator stopped checking the suffix for things; I just pulled the update and a lot of changes were made. _sigh_

zakird: wat?

kumarde: I have literally no idea why this test was all of a sudden breaking. Something changed in the underlying govalidator library that we rely on to check if something is an FQDN.

zakird:
What's the current status of this?

kumarde:
@zakird ready for code review

PR #36, opened 2017-04-24
Add an example of binding Lints to struct fields

dadrian:
This isn't meant to be merged, but is just an example of one way to achieve what I think you're trying to achieve in #35.

There's some wiggle room here with the details, for example the `updateReport` function could become a private interface with a private single function which would have to be defined by each lint. 

Alternatively, you could make `Result` and interface and declare the `updateReport` there. Then each Lint would implement its own `Result` capable of updating a `LintReport` with a `ResultStruct`. This might end up being a lot of duplicate code in each individual Lint, although perhaps embedded structs would make this easier?

### DIFF
--- a/lints/base.go
+++ b/lints/base.go
@@ -1,8 +1,9 @@
 package lints
 
 import (
-	"github.com/zmap/zcrypto/x509"
 	"time"
+
+	"github.com/zmap/zcrypto/x509"
 )
 
 // global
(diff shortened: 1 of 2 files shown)


kumarde:
Cool, this is what I was looking for! I think I like keeping it as a function in a lint because it means each lint abstraction is responsible for updating it's own field without an additional lookup. Thanks!

zakird:
@dadrian do you think it's poor form to have the map as its currently
setup, or am I overreaching?

dadrian:
I think there's a couple things to think about:

1. Client code looking for a specific lint shouldn't be stringly typed. This is fixed by the struct, e.g. `report.EBasicConstraintsNotCritical` instead of `EBasicConstraintsNotCritical, ok := lints["EBasicConstraintsNotCritical"]`.

2. However, in many cases you probably don't care about individual lints, and want to iterate over them and still know their name. I think the current struct design proposed here doesn't lend itself to that, although with some more tinkering, could probably provide some method to get something along the lines of `[]struct{Lint, ResutStruct}`. You certainly don't want to end up with some giant function that has to hardcode out each lint.

3. The `LintReport` struct proposed here should probably store values, not pointers. This prevents also having to check to make sure that each field is not nil. This might take some reevaluating of the zero values of all the structs.

I think a struct solution is important, but the key to making it really solid I think will be good bindings between lints and their respective results, and being able to easily go back and forth from result to lint, to struct field, without having to manually index a map of strings.

kumarde:
@dadrian in order to fill out the protobuf field for Lints, we need to index each lint individually (in the protobuf struct) regardless of a map vs. struct, so even if we had an interface to easily iterate through each Lint I dunno how valuable it is? I don't think there's a better way fill out the protobuf without directly indexing the fields.

dadrian:
You could extend the same update function approach to also work with taking a Lint and updating the proto struct.

kumarde:
True, but I feel like that would couple the ZLint code the Protobuf code and I feel like it could get messy if we need to use something else down the line. Do you envision that as a problem?

zakird:
I do agree that it makes sense to not have the protobuf code be a dependency of zlint. I think you could easily autogenerate this function though.

kumarde:
@zakird yeah, it's easy to auto generate.

kumarde:
Closing. Thanks @dadrian!

PR #42, opened 2017-05-03
Kumarde/fill out fields

kumarde:
Filling out outstanding field required for cert-daemon

### DIFF
--- a/lints/base.go
+++ b/lints/base.go
@@ -15,13 +15,13 @@ type lintReportUpdater func(*LintReport, ResultStruct)
 const ZLintVersion = 1
 
 type ZLintResult struct {
-	ZLintVersion   int64
-	Timestamp      int64
-	ZLints         *LintReport
-	NoticePresent  bool
-	WarningPresent bool
-	ErrorsPresent  bool
-	FatalsPresent  bool
+	ZLintVersion    int64
+	Timestamp       int64
+	ZLint           *LintReport
+	NoticesPresent  bool
+	WarningsPresent bool
+	ErrorsPresent   bool
+	FatalsPresent   bool
 }
 
 type LintReport struct {
@@ -211,7 +211,6 @@ type LintReport struct {
 	ERsaNoPublicKey                                      ResultStruct `json:"e_rsa_no_public_key,omitempty"`
 	ESubCertCertificatePoliciesMissing                   ResultStruct `json:"e_sub_cert_certificate_policies_missing,omitempty"`
 	ESubCertKeyUsageCrlSignBitSet                        ResultStruct `json:"e_sub_cert_key_usage_crl_sign_bit_set,omitempty"`
-	infosPresent                                         bool         `json:"infos_present,omitempty"`
 	noticesPresent                                       bool         `json:"notices_present,omitempty"`
 	warningsPresent                                      bool         `json:"warnings_present,omitempty"`
 	errorsPresent                                        bool         `json:"errors_present,omitempty"`
--- a/lints/base.go
+++ b/lints/base.go
@@ -15,13 +15,13 @@ type lintReportUpdater func(*LintReport, ResultStruct)
 const ZLintVersion = 1
 
 type ZLintResult struct {
-	ZLintVersion    int64
-	Timestamp       int64
-	ZLint           *LintReport
-	NoticesPresent  bool
-	WarningsPresent bool
-	ErrorsPresent   bool
-	FatalsPresent   bool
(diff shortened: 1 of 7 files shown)


lints/base.go
```diff
@@ -207,6 +207,11 @@ type LintReport struct {
 	ERsaNoPublicKey                                      ResultStruct `json:"e_rsa_no_public_key,omitempty"`
 	ESubCertCertificatePoliciesMissing                   ResultStruct `json:"e_sub_cert_certificate_policies_missing,omitempty"`
 	ESubCertKeyUsageCrlSignBitSet                        ResultStruct `json:"e_sub_cert_key_usage_crl_sign_bit_set,omitempty"`
+	InfosPresent                                         bool         `json:"infos_present,omitempty"`
```
zakird: These should be in `ZLintResult`...

lints/base.go
```diff
@@ -216,10 +221,24 @@ func (report *LintReport) Execute(cert *x509.Certificate) error {
 			return err
 		}
 		l.updateReport(report, res)
+		updateErrorStatePresent(report, res)
 	}
 	return nil
 }
 
+func updateErrorStatePresent(report *LintReport, result ResultStruct) {
```
zakird: not entirely sure why this is a function but fine

lints/base.go
kumarde: _sigh_, Yeah, I agree. The issue I was running into is that the results are abstracted away from the `ZLintResult` struct. I think what I can do is have the `LintReport` keep track of whether or not it has an error or warning or wahtever, and then bubble that up to a field in `ZLintResult`

zakird: Huh?

lints/base.go
zakird: I'm not sure why this is here either. Why isn't all of this code in the protobuf thing that you generate given that's the location we need it?

lints/base.go
kumarde: I'll pass them up to the `ZLintResult`

lints/base.go
kumarde: To fill out the protobuf, we need to have this data available. I'd rather not iterate through every lint just to determine that there's an error or warning present, it's better if ZLint itself keeps track of that stuff and just exposes a bool to the caller

kumarde:
@zakird mind taking a look at this whenever you get the chance? This needs to land before it gets updated in cert-daemon.

lints/base.go
```diff
@@ -15,9 +15,13 @@ type lintReportUpdater func(*LintReport, ResultStruct)
 const ZLintVersion = 1
 
 type ZLintResult struct {
-	ZLintVersion int64
-	Timestamp    int64
-	ZLints       *LintReport
+	ZLintVersion   int64
+	Timestamp      int64
+	ZLints         *LintReport
+	NoticePresent  bool
```
zakird: please use consistent pluralization.

lints/base.go
```diff
@@ -207,6 +211,11 @@ type LintReport struct {
 	ERsaNoPublicKey                                      ResultStruct `json:"e_rsa_no_public_key,omitempty"`
 	ESubCertCertificatePoliciesMissing                   ResultStruct `json:"e_sub_cert_certificate_policies_missing,omitempty"`
 	ESubCertKeyUsageCrlSignBitSet                        ResultStruct `json:"e_sub_cert_key_usage_crl_sign_bit_set,omitempty"`
+	infosPresent                                         bool         `json:"infos_present,omitempty"`
```
zakird: Why are there five vars here, but everywhere else in this PR, there are only four?

kumarde: whoops.

lints/base.go
kumarde: done

lints/base.go
```diff
@@ -207,6 +211,10 @@ type LintReport struct {
 	ERsaNoPublicKey                                      ResultStruct `json:"e_rsa_no_public_key,omitempty"`
 	ESubCertCertificatePoliciesMissing                   ResultStruct `json:"e_sub_cert_certificate_policies_missing,omitempty"`
 	ESubCertKeyUsageCrlSignBitSet                        ResultStruct `json:"e_sub_cert_key_usage_crl_sign_bit_set,omitempty"`
+	noticesPresent                                       bool         `json:"notices_present,omitempty"`
```
zakird: This is not the right location for these to be output in the JSON.

kumarde: fixed

PR #51, opened 2017-05-22
Kumarde/lintprint

kumarde:
Printing lints.


dadrian:
Should this not just be a flag to zlint?

zakird:
(1) I agree
(2) that's not how you make a csv.

zlint/zlint.go
```diff
@@ -5,13 +5,37 @@
...
+func PrettyPrintZLint() {
+	for _, l := range lints.Lints {
+		p := PrettyOutput{}
+		p.Name = l.Name
+		p.Description = l.Description
+		p.Providence = l.Providence
+		out, _ := json.Marshal(p)
+		out = bytes.Replace(out, []byte("\\u003c"), []byte("<"), -1)
+		out = bytes.Replace(out, []byte("\\u003e"), []byte(">"), -1)
+		out = bytes.Replace(out, []byte("\\u0026"), []byte("&"), -1)
```
dadrian: Wat

kumarde: man, idk. json.Marshal encodes those three characters to be HTML compatible (so they're replaced with \u003c, \u003e and \u0026 but that makes no sense for pretty printing. I think there's a more idiomatic solution in golang 1.7, I'll look into it

dadrian: This isn't really a pretty print, given that's it's outputting minified JSON objects, so I'd say just leave it.

zakird: This doesn't really make sense.

kumarde: `json.Marshal` out of the box encodes those characters as "HTML Safe". What doesn't make sense? That I have to do this?

zakird:
Why is it a problem to have these encoded as html safe?

kumarde:
You get JSON that looks like this 

`{"name":"e_ca_key_cert_sign_not_set","description":"Root \u0026 Subordinate CA certificate keyUsage extension's keyCertSign bit must be set","providence":"CAB: 7.1.2.1"}`

I guess we could just remove those special characters from the descriptions though. They don't happen that frequently. I don't know what that would look like in HTML

zakird:
Wait, what? What does that look like without that? Why is there anything
that's not ASCII in these descriptions?

kumarde:
Nah, but `json.Marshal` does that to `&`, `<`, and `>` by default

zakird:
Why are there any of those characters in that lint description??

kumarde:
So, `Root & Subordinate CA certificates` -> `Root \u0026 Subordinate CA certificates`

kumarde:
We can just make all of them english, we're doing a pass on descriptions anyway!

zakird:
That makes the most sense to me. This also just fixable by setting
```
enc := json.NewEncoder(os.Stdout)
enc.SetEscapeHTML(false)
```

kumarde:
Actually, some descriptions look like this: 

`email must not be surrounded with <>, and there must be no trailing comments in ()` so I'll do the encoding thing you mentioned above.

zlint/zlint.go
```diff
@@ -28,11 +28,12 @@ func PrettyPrintZLint() {
...
-		out = bytes.Replace(out, []byte("\\u003c"), []byte("<"), -1)
-		out = bytes.Replace(out, []byte("\\u003e"), []byte(">"), -1)
-		out = bytes.Replace(out, []byte("\\u0026"), []byte("&"), -1)
-		fmt.Println(string(out))
+
+		buffer := new(bytes.Buffer)
+		enc := json.NewEncoder(buffer)
+		enc.SetEscapeHTML(false)
+		enc.Encode(p)
+		fmt.Print(string(buffer.String()))
```
dadrian: You don't need `string()` if you call `buffer.String()`

lints/base.go
```diff
@@ -222,6 +223,9 @@ func (result *ZLintResult) Execute(cert *x509.Certificate) error {
 	for _, l := range Lints {
 		res, _ := l.ExecuteTest(cert)
 		l.updateReport(result.ZLint, res)
+		if res.Result == Warn {
+			fmt.Println(l)
```
dadrian: Why is this included in this PR?

main.go
```diff
@@ -43,13 +44,18 @@ func init() {
 	flag.StringVar(&inPath, "in-file", "", "File path for the input certificate(s).")
 	flag.StringVar(&outPath, "out-file", "-", "File path for the output JSON.")
 	flag.StringVar(&outStat, "out-stat", "-", "File path for the output stats.")
+	flag.BoolVar(&prettyPrint, "pretty", false, "Use this flag to pretty print supported lints in JSON format")
```
dadrian: IMHO this should be `--list-lints-json` or something like that. JSON isn't pretty, and minified JSON certainly isn't pretty.

dadrian:
I'm a little confused. Are you pretty printing things, or are you dumping the list of lints to JSON? Because a list of minified JSON objects, one per line, isn't pretty, it's machine parseable. And if you're being parsed by a machine, why not escape HTML?

kumarde:
Well at first I just wanted to print out all the lints, their descriptions, and their providence. I don't entirely know what happens when a `\u0026` for example appears in HTML, so I just wanted to make everything printable. I've fixed that -pretty flag.

lints/base.go
zakird: ditto

dadrian:
In retrospect, I'm just fixing this.

PR #53, opened 2017-05-24
Lints that only apply to subscriber certs should not apply to CA certs

kumarde:
Many lints don't properly check if a lint applies to a CA cert; this adds that check to the offending lints.


kumarde:
@dadrian got it. I'll add an issue.

PR #55, opened 2017-05-25
adding defense checks to lints where the did not appear

kumarde:
Sometimes, errors wouldn't be checked, probably causing runtime bugs.


lints/lint_ext_policy_constraints_empty.go
```diff
@@ -40,7 +40,10 @@ func (l *policyConstraintsContents) CheckApplies(c *x509.Certificate) bool {
 func (l *policyConstraintsContents) RunTest(c *x509.Certificate) (ResultStruct, error) {
 	pc := util.GetExtFromCert(c, util.PolicyConstOID)
 	var seq asn1.RawValue
-	asn1.Unmarshal(pc.Value, &seq) //only one sequence, so rest should be empty
+	_, err := asn1.Unmarshal(pc.Value, &seq) //only one sequence, so rest should be empty
```
dadrian: So, this is definitely better, but what would happen in the old code path?

PR #56, opened 2017-05-25
default values

kumarde:
Default values for each ResultStruct.

Committing here to get opinions on this.


dadrian:
Just noting for posterity from a discussion offline: Default values in Golang should instead be implemented as meaningful zero value.

PR #59, opened 2017-05-30
Kumarde/rdn attributes

kumarde:
Adding check for multiple RDN Attributes. This addresses #31.


lints/base.go
```diff
@@ -12,7 +12,7 @@ var (
 
 type lintReportUpdater func(*LintReport, ResultStruct)
 
-const ZLintVersion = 2
+const ZLintVersion = 3
```
dadrian: Do we update this on new lints, or when we make breaking changes?

lints/lint_issuer_multiple_rdn.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"encoding/asn1"
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zcrypto/x509/pkix"
+	"github.com/zmap/zlint/util"
+)
+
+type IssuerRDNHasMultipleAttribute struct {
+	// Internal data here
```
dadrian: Just let this be an empty struct, no comment needed.

`type IssuerRDNHasMultipleAttribute struct {}`

lints/lint_issuer_multiple_rdn.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (l *IssuerRDNHasMultipleAttribute) Initialize() error {
+	return nil
+}
+
+func (l *IssuerRDNHasMultipleAttribute) CheckApplies(c *x509.Certificate) bool {
```
dadrian: Does this really apply to all certificates and not just subscriber certificates?

lints/lint_subject_multiple_rdn.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"encoding/asn1"
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zcrypto/x509/pkix"
+	"github.com/zmap/zlint/util"
+)
+
+type SubjectRDNHasMultipleAttribute struct {
+	// Internal data here
```
dadrian: Ditto

lints/lint_subject_multiple_rdn.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (l *SubjectRDNHasMultipleAttribute) Initialize() error {
+	return nil
+}
+
+func (l *SubjectRDNHasMultipleAttribute) CheckApplies(c *x509.Certificate) bool {
+	return true
```
dadrian: Ditto

testlint/testCerts/issuerRDNTwoAttribute.pem
```diff
@@ -0,0 +1,16 @@
+-----BEGIN CERTIFICATE-----
```
dadrian: I don't know if we've been doing this in the past, but I like to have the OpenSSL -text output either prepended or appended to the PEM for test certificates, so they're a little easier to understand when you go look at them later.

```console
$ openssl x509 -in issuerRDNTwoAttribute.pem -out temp.pem -text
$ mv temp.pem issuerRDNTwoAttribute.pem
```

lints/base.go
kumarde: Based on a conversation w/ Zakir, he mentioned doing this for new lints. Maybe it's something we need to think about more closely.

lints/lint_issuer_multiple_rdn.go
kumarde: Done

lints/lint_issuer_multiple_rdn.go
kumarde: This is from AWS Certlint, and it looks like they check this on every certificate. But I can't find it in an RFC or the BR, so unfortunately I don't have the text to support it

lints/lint_subject_multiple_rdn.go
kumarde: done

lints/lint_subject_multiple_rdn.go
kumarde: done

lints/lint_issuer_multiple_rdn.go
kumarde: done

testlint/testCerts/issuerRDNTwoAttribute.pem
kumarde: we haven't been doing this, but it's a good idea. I'll send a PR where I do this for all our test certs.

PR #62, opened 2017-05-31
Ensure Lint fields are set

dadrian:
Ensure Name, Description, and Providence are set.


dadrian:
@kumarde Want to hunt down some provenances?

kumarde:
@dadrian i shall

PR #63, opened 2017-05-31
Add meta-lint for assigning to the correct report field

dadrian:


dadrian:
@kumarde is this my bug or your bug

kumarde:
@dadrian this is my bug :'(

PR #68, opened 2017-06-01
main.go rewrite to utilize Go concurrency primitives.

kumarde:


main.go
```diff
@@ -1,343 +1,103 @@
...
+	multi          bool
+	threaded       bool
+	numCertThreads int
+	prettyPrint    bool
+	numProcs       int
+	channelSize    int
 )
 
 func init() {
 	flag.StringVar(&inPath, "in-file", "", "File path for the input certificate(s).")
```
dadrian: Use `--input-file` to match every other ZMap organization project. Not your fault, I know 😞

main.go
```diff
@@ -1,343 +1,103 @@
...
 	flag.StringVar(&outPath, "out-file", "-", "File path for the output JSON.")
```
dadrian: `--output-file`

main.go
```diff
@@ -1,343 +1,103 @@
...
-	flag.StringVar(&outStat, "out-stat", "-", "File path for the output stats.")
 	flag.BoolVar(&prettyPrint, "list-lints-json", false, "Use this flag to print supported lints in JSON format, one per line")
 	flag.BoolVar(&multi, "multi", false, "Use this flag to specify inserting many certs at once. Certs in this mode must be Base64 encoded DER strings, one per line.")
 	flag.BoolVar(&threaded, "threads", false, "Use this flag to specify that -multi mode runs multi-threaded. This has no effect otherwise.")
```
dadrian: `--threads` should be an integer that defaults to 1 (or maybe 1000?), we shouldn't have `--threads` and `--cert-threads`.

main.go
```diff
@@ -1,343 +1,103 @@
...
-	flag.UintVar(&numThreads, "num-threads", DEFAULT_THREADS, "Use this flag to specify the number of threads in -threads mode.  This has no effect otherwise.")
+	flag.IntVar(&numCertThreads, "cert-threads", 1, "Use this flag to specify the number of threads in -threads mode.  This has no effect otherwise.")
+	flag.IntVar(&numProcs, "procs", 1, "Use this flag to specify the number of processes to run on.")
```
dadrian: This should default to 0 or -1. Setting runtime.GOMAXPROCS to 0 or less means "keep it what it was previously".

main.go
```diff
@@ -1,343 +1,103 @@
...
+	flag.IntVar(&channelSize, "channel-size", 10000, "Use this flag to specify the number of values in the buffered channel.")
```
dadrian: Not sure if this is needed as an option, vs doing something like 2 or 4 times number of threads.

main.go
```diff
@@ -1,343 +1,103 @@
...
+			//log.Info("Could not parse certificate with error:", err)
+		} else { //parsed
+			zlintResult := zlint.ZLintResultTestHandler(parsed)
+			jsonResult, err := json.Marshal(zlintResult.ZLint)
 			if err != nil {
-				fmt.Println(err)
-			} else {
-				reports = append(reports, reportOut) //move report to out buffer
-				lintOut++
+				log.Fatal("Could not parse JSON.")
```
dadrian: Probably don't want to fatal on this. Or at least make "crash on parse fail" an option. For example, on the 130K certs I used as candidate intermediates, we couldn't parse ~20 of them.

dadrian: And to be clear, I still wanted to parse the rest of the certificates in the list.

main.go
```diff
@@ -1,343 +1,103 @@
...
 			}
+			fmt.Println(string(jsonResult))
```
dadrian: Output needs to happen in a separate channel, otherwise you'll get interleaving. I'd pass the encoded output as an `[]byte` to an output channel, and then just call `outputFile.write(b); outputFile.write("\n")`

main.go
```diff
@@ -1,343 +1,103 @@
...
+	go ReadCertificate(certs, inPath, &readerWG)
+	for i := 0; i < numCertThreads; i++ {
+		procWG.Add(1)
+		go ProcessCertificate(certs, &procWG)
+	}
+	go func() {
+		readerWG.Wait()
+		close(certs)
+	}()
+	procWG.Wait()
```
dadrian: I'd change concurrency setup here a little to something like:

1 input thread reads and splits input -> N threads that parse and run zlint and encode to JSON -> 1 thread that writes output.

A case could also be made for:

1 input thread that splits input -> N threads that take sections of input and decode to DER bytes -> N threads that parse certificate and run zlint -> N threads that take cert + zlint and encode to JSON -> 1 thread that writes output.

(Or some combination of that breakdown, I could see a couple setups).

At the end of the day, I'd still want to see a real clear high level main. You're main is good now, but you shouldn't be outputting from each worker goroutine.

Additionally, you should check out some of the code I hacked together in zcertificate for handling different input formats, including the `ScannerFunc` I wrote to split a file by PEM. It might not handle text between PEM's though. We'll want to handle both a PEM file with multiple certificates, and a "csv" of one base64-encoded DER certificate per line.

Overall, WAYYY better than it was previously.

main.go
kumarde: done

main.go
kumarde: done

main.go
kumarde: done

main.go
kumarde: done

main.go
kumarde: @dadrian Did it crash though? Since this is particularly for the `json.Marshal` of the ZLint output. What is the behavior we want when a certificate can't be parsed?

main.go
kumarde: done

main.go
dadrian: I might have been confused about where the log.Fatal was. If a certificate doesn't Marshal, you should fatal. If a certificate doesn't parse, you should continue (but maybe have a flag to cause a log.Fatal instead).

kumarde: Got it. That's what I've got but I'll add the flag in at the next iteration.

main.go
kumarde: wait, 1000?

dadrian: I dunno, whatever a reasonable threaded default is.

main.go
kumarde: I'll rope in the PEM stuff when we move this to `ZCertificate`

main.go
```diff
@@ -76,6 +78,29 @@ func ReadCertificate(out chan<- string, filename string, wg *sync.WaitGroup) {
...
+	var outFile *os.File
+	var err error
+	var writeOutputToFile bool = false
+	if outputFileName != "" && outputFileName != "-" {
+		outFile, err = os.Create(outputFileName)
+		if err != nil {
+			log.Fatal("Unable to create output file: ", err)
+		}
+		defer outFile.Close()
+		writeOutputToFile = true
```
dadrian: Just set the outputfile to `os.Stdout` if `outputFileName == "-"`

kumarde: done

main.go
```diff
@@ -47,14 +45,18 @@ func ProcessCertificate(in <-chan string, wg *sync.WaitGroup) {
 		}
 		parsed, err := x509.ParseCertificate(der)
 		if err != nil { //could not parse
-			//log.Info("Could not parse certificate with error:", err)
+			if crashIfParseFail {
+				log.Fatal("Could not parse certificate with error: ", err)
```
dadrian: Log statements and error messages should start lowercase, because they are often concatenated: `statement: substatement: that had it's own substatement`

main.go
```diff
@@ -15,29 +15,27 @@ import (
...
+	flag.StringVar(&inPath, "input-file", "", "File path for the input certificate(s).")
+	flag.StringVar(&outPath, "output-file", "-", "File path for the output JSON.")
 	flag.BoolVar(&prettyPrint, "list-lints-json", false, "Use this flag to print supported lints in JSON format, one per line")
-	flag.BoolVar(&multi, "multi", false, "Use this flag to specify inserting many certs at once. Certs in this mode must be Base64 encoded DER strings, one per line.")
-	flag.BoolVar(&threaded, "threads", false, "Use this flag to specify that -multi mode runs multi-threaded. This has no effect otherwise.")
 	flag.IntVar(&numCertThreads, "cert-threads", 1, "Use this flag to specify the number of threads in -threads mode.  This has no effect otherwise.")
-	flag.IntVar(&numProcs, "procs", 1, "Use this flag to specify the number of processes to run on.")
+	flag.IntVar(&numProcs, "procs", 0, "Use this flag to specify the number of processes to run on.")
 	flag.IntVar(&channelSize, "channel-size", 10000, "Use this flag to specify the number of values in the buffered channel.")
+	flag.BoolVar(&crashIfParseFail, "crash", false, "Use this flag if you want to crash on parsing failure.")
```
dadrian: How about a more descriptive name and an a declarative (imperative?) description? Maybe `--fatal-parse-errors`?

main.go
dadrian: Yeah, this should really be a function of whatever `--threads` is set to. Why have 10K certificates in memory if you don't have to?

kumarde: What do you think is reasonable? There is a pretty big speedup if you just have them sitting around in memory and you're processing a lot of certificates. For a run of 1M random certs, I got the following times:

```
1000 buffer: 9m37s
10000 buffer: 3m40s
100000 buffer: 2m28s
```

kumarde: Maybe a reasonable default that isn't 10000, but keep it customizable?

kumarde:
@dadrian this is ready to be looked at again!

PR #83, opened 2017-08-20
Add LICENSE

titanous:
It looks like most of the other repos are Apache 2, so I'm assuming this one is supposed to be too.


kumarde:
I'll defer to @zakird for this one.

PR #85, opened 2017-08-20
Simplify report generation

### LINKED ISSUE #58: Creating a new lint is error prone
The current steps to add a new lint are incredibly tedious. There are too many places where we're storing lint information and all of this should be built in a script.

titanous:
This removes the need to maintain the `LintReport` struct and `updateReport` functions. All lint results are added to a map automatically during linting. The resulting JSON is the same, with a few typos fixed that were caused by skew between struct tags.

Closes #58.


kumarde:
Hi @titanous,

Thank you for all your contributions thus far! A note on this one in particular - we originally had ZLint set up this way, but we moved to a LintReport/UpdateReport scheme because it plugs in nicer with other Censys codebases (things that require serializing protobufs in other processes, for example, https://github.com/censys/certificate-daemon, and that code is almost entirely autogenerated). 

If you think of a better way to do this given that information, by all means do let me know!

titanous:
I see. Would using a map in your protobufs for the lint results work? It's also possible to do some codegen to generate the report struct, which you could then use to generate protobufs and so on, but that isn't ideal.

The current report struct requires regenerating the code throughout the pipeline whenever a lint is added, which seems like work that shouldn't have to be done?

kumarde:
I agree, we're duplicating code and structs in too many places and it's error prone to get something working into Censys. 

I'll need to have some additional conversations with @zakird and @dadrian about updating the underlying protobuf, which I think will fix many of these issues. Doing so will result in a sizable number of changes, so I'm closing for now and will comment this PR in the issue.

dadrian:
It's a map! Thanks @titanous.

PR #88, opened 2017-08-20
These key identifier checks only apply to Subscriber Certificates

kumarde:


zakird:
Based on the RFC snippet included, this also applies to intermediate CA certificates. I don't quite follow why we're not checking whether the certificate is self-signed.

kumarde:
@zakird, ah! You're right. I think as I read this, this precludes everything that isn't a root CA. Yes?

zakird:
I don't think I parse your question. This should _include_ everything that's not a root CA, I believe.

PR #91, opened 2017-08-21
Add RootCA Key Usage Present Lint

kumarde:


lints/base.go
```diff
@@ -223,6 +223,7 @@ type LintReport struct {
 	WMultipleIssuer_RDN                                  ResultStruct `json:"w_multiple_issuer_rdn,omitempty"`
 	WMultipleSubject_RDN                                 ResultStruct `json:"w_multiple_subject_rdn,omitempty"`
 	ESignatureAlgorithmNotSupported                      ResultStruct `json:"e_signature_algorithm_not_supported,omitempty"`
+	ERootCaKeyUsagePresent                               ResultStruct `json:"e_root_ca_key_usage_present"`
```
zakird: Probably want CA not Ca here.

kumarde: The other CA lints aren't currently that either, I'll do a different PR to fix this later.

zakird: sounds good.

dadrian: You should check for that in `meta_lint.py`.

PR #95, opened 2017-08-22
subCA EKU Valid Fields

kumarde:


lints/lint_sub_ca_eku_valid_fields.go
```diff
@@ -0,0 +1,41 @@
...
+			return ResultStruct{Result: Error}, nil
+		}
+	}
+	return ResultStruct{Result: Pass}, nil
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_sub_ca_eku_valid_fields",
+		Description:   "Subordinate CA extkeyUsage, either id-kp-serverAuth or id-kp-clientAuth or both values MUST be present.",
```
titanous: I don't believe this interpretation is correct. These EKUs are optional. The BRs say:

> Generally Extended Key Usage will only appear within end entity certificates (as highlighted in RFC 5280 (4.2.1.12)), however, Subordinate CAs MAY include the extension to further protect relying parties until the use of the extension is consistent between Application Software Suppliers whose software is used by a substantial portion of Relying Parties worldwide.

titanous: These are just required for the subCA to be "Technically Constrained", which is a specific type of intermediate with relaxed rules:

> For Subordinate CA Certificates to be Technically constrained in line with section 7.1.5, then either the value id‐kp‐serverAuth [RFC5280] or id‐kp‐clientAuth [RFC5280] or both values MUST be present**.

kumarde: EKU fields are optional, you're right. However, the CheckApplies function in this lint basically checks if the EKU extensions exists in the cert. If it does exist, then they must be one of these fields. Does that make sense?

titanous: I posted the wrong snippet originally. As I said in my second comment, this is only required to make an intermediate technically constrained.

kumarde: Got it. I'll change this to a **notice**, rather than an error. Does that seem reasonable?

titanous: Yep, sounds good.

kumarde: Or, do you think this just shouldn't be a lint at all?

titanous: For the notice, this should say "MUST be present to be technically constrained."

kumarde: Good catch. Fixed.

kumarde:
@zakird @titanous this is ready to be looked at again.

lints/lint_sub_ca_eku_valid_fields.go
```diff
@@ -0,0 +1,41 @@
...
+		} else {
+			return ResultStruct{Result: Notice}, nil
+		}
+	}
+	return ResultStruct{Result: Pass}, nil
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "n_sub_ca_eku_valid_fields",
```
titanous: It might make sense to call this `n_sub_ca_not_technically_constrained_eku` or similar.

lints/base.go
```diff
@@ -226,6 +226,7 @@ type LintReport struct {
 	ERootCaKeyUsagePresent                               ResultStruct `json:"e_root_ca_key_usage_present"`
 	ERootCaKeyUsageMustBeCritical                        ResultStruct `json:"e_root_ca_key_usage_must_be_critical"`
 	ESubCaAIAMarkedCritical                              ResultStruct `json:"e_sub_ca_aia_marked_critical"`
+	NSubCaEkuValidFields                                 ResultStruct `json:"n_sub_ca_eku_not_technically_constrained"`
```
titanous: The report field name should be updated to match as well.

kumarde:
@titanous good catches all around. thanks!

lints/lint_sub_ca_eku_valid_fields.go
```diff
@@ -0,0 +1,41 @@
...
+	return util.IsSubCA(c) && util.IsExtInCert(c, util.EkuSynOid)
+}
+
+func (l *subCAEKUValidFields) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	for _, ekuValue := range c.ExtKeyUsage {
+		if ekuValue == x509.ExtKeyUsageServerAuth ||
+			ekuValue == x509.ExtKeyUsageClientAuth {
+			continue
+		} else {
+			return ResultStruct{Result: Notice}, nil
```
titanous: The text says:

> Other values MAY be present.

Which is counter-intuitive, but what the document says. So I think this test is wrong. Given the uselessness of this specific requirement, I'm wondering if it should be a lint at all.

kumarde: Blagh. 

The way this is currently written, we return a pass if either of the fields are there for being technically constrained, and a notice otherwise. 

Since notices aren't really errors, they're just interesting things to watch for, I'm okay with the lint as is. I'm also okay not having it at all, considering it doesn't tell us anything AND that a previous run of ZLint with this lint told me this never happens.

zakird: I agree that it seems wrong to notice if other values are present. What if the notice was in the situation that neither of those was present rather than if something else is _also_ present?

kumarde: That's fine by me - lemme push something to that effect.

kumarde:
@zakird lmk what you think.

zakird:
Looks good to me. Jonathan, thoughts?

PR #102, opened 2017-08-23
add subCA EKU missing lint

kumarde:


lints/lint_sub_ca_eku_missing.go
```diff
@@ -0,0 +1,38 @@
...
+	return nil
+}
+
+func (l *subCAEKUMissing) CheckApplies(c *x509.Certificate) bool {
+	// Add conditions for application here
+	return util.IsSubCA(c)
+}
+
+func (l *subCAEKUMissing) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	if util.IsExtInCert(c, util.EkuSynOid) {
```
zakird: Is this the right OID here? What's `Syn`?

zakird: As long as it's the right one, /shrug, just wanted to check.

PR #106, opened 2017-08-24
Add Locality name prohibited check

kumarde:


kumarde:
@zakird feel free to take a look at this as well.

kumarde:
@zakird it's true, you have. Just different subject fields but the _same_ properties on all of them. It's almost done though!

PR #107, opened 2017-08-24
Add province required check.

kumarde:


kumarde:
@zakird yeah, a lot of lints have similar requirements. I'll probably add everything in the silly way and then do a PR to change to a helper function in one go at the end.

PR #112, opened 2017-08-24
add serial number low entropy check

kumarde:


zakird:
This seems like a lower bound check, and probably what we can realistically test in isolation. @titanous any thoughts on this one?

titanous:
I don't think there are any feasible tests that could be introduced at this level.

zakird:
Is checking length incorrect?

titanous:
It can be a warning. But you can have certificates containining 64 bits of entropy that do not contain 8 bytes due to leading zeros.

See: https://groups.google.com/d/topic/mozilla.dev.security.policy/YequotPYLdc/discussion

kumarde:
We could alternatively check if there are 64 bits instead of 8 bytes? But if you don't think this is a real test, we could just nix it.

titanous:
> We could alternatively check if there are 64 bits instead of 8 bytes? But if you don't think this is a real test, we could just nix it.

8 bytes == 64 bits. The problem is when I have a random number and the top eight bits are zero, then I will have a 7 byte serial number. I think this should definitely go in as a warning.

kumarde:
Yeah. Okay, I'll update this to a warning.

lints/lint_serial_number_low_entropy.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func (l *serialNumberLowEntropy) CheckApplies(c *x509.Certificate) bool {
+	// Add conditions for application here
+	return true
+}
+
+func (l *serialNumberLowEntropy) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	if len(c.SerialNumber.Bytes()) < 8 {
+		return ResultStruct{Result: Error}, nil
```
titanous: This should be `Warn`

RufusJWB:
We still have some of the certificates with eight leading bits of 0 in the field (https://crt.sh/?id=313099468&opt=zlint) and this warning is an annoyance to our automatic error warning system. What makes me wonder is, that if I query directly the crt.sh database with postgresql I receive the correct serial number with leading zeros: image Maybe there is a trick that could be copied here this issue?

RufusJWB:
@kumarde Did you have a chance to look into this issue?

zakird:
I'm not sure what's being asked here. It's probably best to also start a new issue rather than tack this onto something that was closed more than a year ago.

PR #113, opened 2017-08-24
Add initial FQDN testing

kumarde:
The library we were using (asaskevich/govalidator) doesn't handle all inputs correctly, for both the `IsDNSName` and `IsURL` functions. (notably, it fails on punycode and has some strange edge cases). 

I've added a simple regexp that I think works, but I would like some input from @titanous and @zakird on this. Are there standard regular expressions that I should be pulling from somewhere to do this check?

One edge case I'm wondering about: is `abc` an FQDN? The regexp from cablint says it is, but my intuition says it isn't. Any input would be appreciated.

### DIFF
--- a/util/fqdn.go
+++ b/util/fqdn.go
@@ -1,8 +1,9 @@
 package util
 
 import (
-	"github.com/asaskevich/govalidator"
+	//"github.com/asaskevich/govalidator"
 	"net"
+	"regexp"
 	"strings"
 )
 
@@ -15,12 +16,21 @@ func removeQuestionMarks(domain string) string {
 	return domain
 }
 
+func checkFQDN(domain string) bool {
+	//From https://github.com/awslabs/certlint/blob/25d5957f8c36dafcbd82870df57a8367b49650be/lib/certlint/generalnames.rb
+	dnsLabel := "([A-Za-z0-9*_-]{1,63})"
+	fqdn := `\A(` + dnsLabel + `\.)*` + dnsLabel + `\z`
+	re := regexp.MustCompile(fqdn)
+	matched := re.MatchString(domain)
+	return matched
+}
+
 func IsFQDN(domain string) bool {
 	domain = removeQuestionMarks(domain)
 	if strings.HasPrefix(domain, "*.") {
 		domain = domain[2:]
 	}
-	return govalidator.IsURL(domain)
+	return checkFQDN(domain)
 }
 
 func GetAuthority(uri string) string {
--- a/util/fqdn.go
+++ b/util/fqdn.go
@@ -1,7 +1,6 @@
 package util
 
 import (
-	//"github.com/asaskevich/govalidator"
 	"net"
 	"regexp"
 	"strings"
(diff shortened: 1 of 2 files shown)


kumarde:
(ps. Once we agree on what is and what isn't an FQDN, I'll write tests to that spec.)

dadrian:
I believe this is the relevant commit in Boulder: https://github.com/letsencrypt/boulder/commit/1209102e7ee23eeb9ed39cc2c0b6c1f62e05310a

titanous:
I believe the following rules should be implemented:

- Domain labels are separated by the ASCII dot character.
- Each label must be only ASCII letters, digits, hyphens and underscores or an asterisk.
- No labels can be empty.
- The domain must end with an entry in the ICANN section of the public suffix list (aka eTLD, it may contain multiple labels). The next label to the left is the "second-level domain" and the combination of the eTLD and the SLD is the eTLD+1. This package works: https://github.com/weppos/publicsuffix-go
- The TLD must not be on the ICANN special-use domain list with the exception of `onion`.
- Underscores that are in the labels of the SLD label trigger an error.
- Hyphens at the beginning or end of the SLD label trigger an error.
- Underscores in labels left of the eTLD+1 trigger a warning (technically not allowed by RFC but in practice they are used and validated by CAs, there is a ballot in the works to officially permit them).
- Hyphens at the beginning or end of a label above the eTLD+1 trigger a warning.
- An asterisk is allowed if it is the only character in the left-most label.
- If the only non-wildcard labels are the eTLD, a warning should be generated as it means the entire eTLD is wildcarded, which is allowed only with permission from the eTLD owner.
- Labels starting with `xn--` are IDN labels and should be checked against the rules in RFC 3490 for proper encoding (Bidi rule, NFKC, and a few others). It's debatable if these should trigger warnings or errors.

I don't have the RFC references handy for these, but they should be easy enough to chase down via 5280.

zakird:
@kumarde can you see which of these are implemented in other lints already?

pzb:
The PSL is not a canonical list of suffixes, as the default rule is '*', so "The domain must end with an entry in the ICANN section of the public suffix list" is not a good rule.

I would also add:
Underscores at the beginning or end of a label left of the eTLD+1 trigger a warning.

Labels matching `/^..--/` but not starting with `xn--` are reserved labels and should trigger a warning.

titanous:
Instead of using the PSL for TLDs we can do this:

- The right-most label should be in the [root zone](https://www.internic.net/domain/root.zone), `onion` (the only special-use domain we care about), or a [removed TLD](https://github.com/pzb/TLDs/blob/master/removed/rmtlds.csv) (possibly with a requirement that the notBefore is within the date range on the removed list plus the maximum domain validation reuse period).
- The PSL should be used to calculated the eTLD+1 for rules that require it but is not the canonical TLD source, and have a fallback for when the TLD is not on the PSL yet.

kumarde:
@titanous okay - so by applying the "the domain must end with something that's in the root zone", the string `abc` is NOT an FQDN. Yes?

Also, what is the process by which you look at whether the right most label is in the root zone? Is there a standard process or an RFC for this?

titanous:
Correct.

You can parse the zone using the [dns package](https://godoc.org/github.com/miekg/dns#ParseZone) or just hack together something simpler then grab all of the TLDs.

kumarde:
@pzb @titanous @zakird I'll split these out into a few lints tomorrow and PR them individually. thanks for your input.

titanous:
A few more:

- The length of each label must be less than or equal to 63 bytes.
- The total length of the domain must be less than or equal to 253 bytes.
- There must be at least one label in addition to the right-most label from the root zone.

kumarde:
Hi all. 

We're done addressing all these notes, you can find them all split up in the files `lint_dnsname_*` (there are many of them). Thanks to everyone for all of your help!

PR #114, opened 2017-08-24
Check Internationalized DNSNames for NFKC

kumarde:
Must follow RFC3490. Test cases found from recent LE bug.


lints/lint_idn_dnsname_must_be_nfkc.go
```diff
@@ -0,0 +1,51 @@
...
+	return nil
+}
+
+func (l *IDNNotNFKC) CheckApplies(c *x509.Certificate) bool {
+	return util.IsExtInCert(c, util.SubjectAlternateNameOID)
+}
+
+func (l *IDNNotNFKC) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	for _, dns := range c.DNSNames {
+		splitLabels := strings.Split(dns, ".")
```
dadrian: Just call this labels.

lints/lint_idn_dnsname_must_be_nfkc.go
```diff
@@ -0,0 +1,51 @@
...
+		for _, label := range splitLabels {
+			if strings.HasPrefix(label, "xn--") {
+				//IDN domain name
+				unicodeLabel, err := idna.ToUnicode(label)
+				if err != nil {
+					return ResultStruct{Result: Fatal}, nil
```
dadrian: Should this be `Fatal` or `Error`? If it's not unicode it's not valid NFKC.

lints/lint_idn_dnsname_must_be_nfkc.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (l *IDNNotNFKC) CheckApplies(c *x509.Certificate) bool {
+	return util.IsExtInCert(c, util.SubjectAlternateNameOID)
+}
+
+func (l *IDNNotNFKC) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	for _, dns := range c.DNSNames {
+		splitLabels := strings.Split(dns, ".")
+		for _, label := range splitLabels {
+			if strings.HasPrefix(label, "xn--") {
```
dadrian: Are you sure this is the correct way of catching punycode? We should probably match whatever regexp is used here: https://github.com/letsencrypt/boulder/commit/1209102e7ee23eeb9ed39cc2c0b6c1f62e05310a#diff-e613fb1bb146bc2843103b461bef02c8R217.

lints/lint_idn_dnsname_must_be_nfkc.go
```diff
@@ -0,0 +1,51 @@
...
+				//IDN domain name
```
dadrian: `// IDN`

lints/lint_idn_dnsname_must_be_nfkc.go
titanous: Yes, this is correct.

lints/lint_idn_dnsname_must_be_nfkc.go
kumarde: 👍

lints/lint_idn_dnsname_must_be_nfkc.go
kumarde: 👍

lints/lint_idn_dnsname_must_be_nfkc.go
kumarde: 👍

lints/lint_idn_dnsname_must_be_nfkc.go
kumarde: We've been treating external library calls that error as `Fatal`, so we don't confuse them with an actual `Error` from the lint.

titanous: This means that an invalid IDN label will Fatal this check, is that intended?

kumarde: I think that's the right call. What I can do is add another lint that returns a warn if there's an invalid IDN label. Does that sound reasonable?

titanous: The intention is to warn in that case, so should this Pass? (or whatever the value is if the check wasn't run)

kumarde: Sorry, I'm a bit confused - if `idna.ToUnicode` returns an error, do you think this should instead be a Pass, rather than a Fatal?

titanous: If there is a separate rule that warns in that case, it seems like this rule should not return a fatal, in the same that a rule covering an optional extension works when the extension doesn't exist.

kumarde: OH, I see. Rather than `Fatal`, return `NA`, as in, this rule doesn't apply here. Makes sense to me 👍

lints/lint_idn_dnsname_must_be_nfkc.go
```diff
@@ -0,0 +1,50 @@
...
+func (l *IDNNotNFKC) CheckApplies(c *x509.Certificate) bool {
+	return util.IsExtInCert(c, util.SubjectAlternateNameOID)
+}
+
+func (l *IDNNotNFKC) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	for _, dns := range c.DNSNames {
+		labels := strings.Split(dns, ".")
+		for _, label := range labels {
+			if strings.HasPrefix(label, "xn--") {
+				//IDN
```
zakird: David mentioned this earlier. Please fix spacing or remove.

PR #116, opened 2017-08-25
Update TLD Script

kumarde:
Consolidated right label checks will happen against data from these sources:

http://data.iana.org/TLD/tlds-alpha-by-domain.txt (which I believe is populated from the root zone periodically) 
"ONION"
https://raw.githubusercontent.com/pzb/TLDs/master/removed/rmtlds.csv (any removed TLDs)

And generates a golang map from the gathered data, stored in util/gtld_map.go.


lints/lint_dnsname_right_label_valid_tld.go
```diff
@@ -0,0 +1,37 @@
...
+func (l *DNSNameValidTLD) Initialize() error {
+	return nil
+}
+
+func (l *DNSNameValidTLD) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.DNSNamesExist(c)
+}
+
+func (l *DNSNameValidTLD) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	for _, dns := range c.DNSNames {
```
zakird: Why doesn't this check CN?

PR #117, opened 2017-08-25
add empty label dnsname check

kumarde:


lints/lint_dnsname_contains_empty_label.go
```diff
@@ -0,0 +1,52 @@
...
+		if elem == "" {
+			return true
+		}
+	}
+	return false
+}
+
+func (l *DNSNameEmptyLabel) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	result := ResultStruct{Result: Pass}
+	if domainHasEmptyLabel(c.Subject.CommonName) {
```
zakird: We probably don't want to be running this on CN if this isn't a subscriber certificate.

lints/lint_dnsname_contains_empty_label.go
```diff
@@ -0,0 +1,52 @@
...
```
zakird: This doesn't look like it checks whether CN is empty.

lints/lint_dnsname_contains_empty_label.go
```diff
@@ -0,0 +1,52 @@
...
+	for _, elem := range labels {
+		if elem == "" {
+			return true
+		}
+	}
+	return false
+}
+
+func (l *DNSNameEmptyLabel) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	result := ResultStruct{Result: Pass}
```
zakird: Why are your defining this instead of just returning a value later?

lints/lint_dnsname_contains_empty_label.go
```diff
@@ -0,0 +1,53 @@
+package lints
+
+import (
```
zakird: This doesn't look `go fmt`ed.

kumarde: it is or it wouldn't have passed the `gofmt` test!

zakird: ok!

PR #119, opened 2017-08-25
adding SLD hyphen check

kumarde:


zakird:
This should probably return `NA` not `PASS` if the certificate doesn't have a domain name (e.g., a CA certificate).

This problem seems to exist in a number of these lints that are checking DNSName validity. Does it make sense to create a helper that can be called in `CheckApplies` to determine whether we should be checking domain validity in a give cert?

kumarde:
@zakird I think this can be fixed by checking `IsSubscriberCert` for all the DNS checks. I'll add these to all the outstanding PRs - granted there aren't situations where CA certs have domain names 
 somewhere that need to be checked?

zakird:
That sounds reasonable.

lints/lint_dnsname_hyphen_in_sld.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (l *DNSNameHyphenInSLD) Initialize() error {
+	return nil
+}
+
+func (l *DNSNameHyphenInSLD) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.DNSNamesExist(c)
+}
+
+func hyphenInSLD(domain string) (bool, ResultStruct) {
```
zakird: I'd like this to return an Error not a Result struct.

lints/lint_dnsname_hyphen_in_sld.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (l *DNSNameHyphenInSLD) Initialize() error {
+	return nil
+}
+
+func (l *DNSNameHyphenInSLD) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.DNSNamesExist(c)
+}
+
+func hyphenInSLD(domain string) (bool, error) {
```
zakird: This isn't what this function actually tests.

PR #120, opened 2017-08-25
adding underscore in left of ETLD+1 check

kumarde:


lints/lint_dnsname_underscore_in_trd.go
```diff
@@ -0,0 +1,55 @@
...
+	return nil
+}
+
+func (l *DNSNameUnderscoreInTRD) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c)
+}
+
+func underscoreInTRD(domain string) bool {
+	domainName, err := publicsuffix.Parse(domain)
+	if err != nil {
```
zakird: This function should really return an error and not just reduce this to a boolean value. These are not equivalent situations.

zakird:
In a lot of these, we're returning `PASS` or on failure to parse correctly, which I don't agree with. If you're worried that you're going to have names that aren't valid DNS names, we should have another check for that and either be returning `NA` or `FATAL` depending on the context.

PR #125, opened 2017-08-26
Remove nested zlint package structure

dadrian:
This moves main.go to a cmd/ subpackage, and moves zlint/zlint to the
root package. It further simplifies main.go to operate only on single
certificates. Multiple certificate linting should be accomplished using
the ZCertificate utility.


zakird:
Thanks for tackling this. This has been bothering me too. :)

cmd/main.go
```diff
@@ -0,0 +1,86 @@
+package main
```
zakird: In other projects, we've been doing `project/command_name` instead of `project/cmd` such that multiple commands can potentially be built. For example, `github.com/zmap/zdns/zdns`. 

What are your thoughts on doing `github.com/zmap/zlint/zlint`? If this is going against golang accepted practice, we should change the other ones, but it'd be nice to be consistent across Z* projects.

zlint.go
```diff
@@ -0,0 +1,50 @@
...
+		enc := json.NewEncoder(buffer)
+		enc.SetEscapeHTML(false)
+		enc.Encode(p)
+		w.Write(buffer.Bytes())
+	}
+}
+
+// LintCertificate runs all registered lints on c, producing a ZLintResult.
+func LintCertificate(c *x509.Certificate) *lints.ZLintResult {
+	if c == nil {
```
zakird: This seems like an odd path offhand.

zakird:
Just as a note, we should update `zcertificate` this weekend. I think we've made several breaking changes this week.

cmd/main.go
zakird: One other follow up question (if you know off the top of your head), when you `go get github.com/zmap/zlint/cmd`, what's the name of the command that's installed in`bin/`?

titanous:
The Go idiom is to put main/cLI packages in a directory structure like `cmd/zlint`, `cmd/zfoo`, etc

cmd/main.go
zakird: Most of the examples I'm seeing online are of the form `cmd/name`. If we'd like the `cmd` to be present, I think it'd be best to be doing `cmd/zlint`.

zlint.go
dadrian: I don't like this function either, I just copied it over from the old version and gave it a better name. It should really return an error meaning "zlint could not run for some reason that is not lint related", but I think that's outside of the scope of this PR.

cmd/main.go
dadrian: Done.

cmd/zlint/main.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/zmap/zlint"
+)
+
+var ( //flags
+	listLintsJSON bool
+	format        string
+)
+
+func init() {
+	flag.BoolVar(&listLintsJSON, "list-lints-json", false, "Use this flag to print supported lints in JSON format, one per line")
```
zakird: I can't decide if this is reasonable or if this should be its own executable.

PR #127, opened 2017-08-27
Add test for wildcard left of ETLD

kumarde:


lints/lint_dnsname_wildcard_left_of_public_suffix.go
```diff
@@ -0,0 +1,54 @@
...
+	return nil
+}
+
+func (l *DNSNameWildcardLeftofPublicSuffix) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c)
+}
+
+func wildcardLeftOfPublicSuffix(domain string) bool {
+	parsedDomain, err := publicsuffix.Parse(domain)
+	if err != nil {
```
zakird: This seems like it should be a fatal. I'm a bit concerned we're handling these inconsistently across lints.

lints/lint_dnsname_wildcard_left_of_public_suffix.go
```diff
@@ -0,0 +1,54 @@
...
+		return false
+	}
+	if parsedDomain.TRD == "" {
```
zakird: why is this a check? Having `terrible.*.com` should still fail this lint.

lints/lint_dnsname_wildcard_left_of_public_suffix.go
```diff
@@ -0,0 +1,54 @@
...
+		if parsedDomain.SLD == "*" {
+			return true
+		}
+	}
+	return false
+}
+
+func (l *DNSNameWildcardLeftofPublicSuffix) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	if wildcardLeftOfPublicSuffix(c.Subject.CommonName) {
+		return ResultStruct{Result: Warn}, nil
```
zakird: Why is this a warning and not an error?

lints/lint_dnsname_wildcard_left_of_public_suffix.go
kumarde: This test checks _only_ for the case *.<ETLD>, for example, *.co.uk, or *.com.

From the BRs: "the CA MUST establish and follow a documented procedure[^pubsuffix] that determines if the wildcard character occurs in the first label position to the left of a “registry‐controlled” label or “public suffix”".

Since we can't explicitly test the "documented procedure", @titanous recommended we do this as a warning in case we see this in the wild.

kumarde: (the other wildcard tests were merged last friday)

lints/lint_dnsname_wildcard_left_of_public_suffix.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func (l *DNSNameWildcardLeftofPublicSuffix) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c)
+}
+
+func wildcardLeftOfPublicSuffix(domain string) (bool, ResultStruct) {
+	parsedDomain, err := publicsuffix.Parse(domain)
+	if err != nil {
+		return true, ResultStruct{Result: NA}
```
zakird: Why is this NA and not Fatal? If we can't parse the domain of a certificate, that seems like a problem worth reporting.

lints/lint_dnsname_wildcard_left_of_public_suffix.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+	if parsedDomain.TRD == "" {
+		if parsedDomain.SLD == "*" {
+			return true, ResultStruct{Result: Warn}
```
zakird: Why is this a WARN? The lint name is `e_dnsname_wildcard_left_of_public_suffix`.

kumarde: good catch, it should be w_.

lints/lint_dnsname_wildcard_left_of_public_suffix.go
kumarde: Hrm. My thought was that if it's _not_ a domain, then this test shouldn't apply. So, a CA that throws random crap into the SANs or CN. But I can make this a fatal if you think it makes more sense.

zakird:
If this is limited to subscriber certs then this should not occur

kumarde:
@zakird okay sounds good, I will fatal out.

lints/lint_dnsname_wildcard_left_of_public_suffix.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (l *DNSNameWildcardLeftofPublicSuffix) Initialize() error {
+	return nil
+}
+
+func (l *DNSNameWildcardLeftofPublicSuffix) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.DNSNamesExist(c)
+}
+
+func wildcardLeftOfPublicSuffix(domain string) (bool, ResultStruct) {
```
zakird: I don't think it makes sense to have ResultStruct at this point. Why don't you just return the err?

kumarde: I can do that and just check the value of err in the parent function, usually I don't propagate the error to higher level functions because it's harder to tell where the error is actually handled. What do you think? (This applies to the other lints as well)

zakird: I'd prefer to have an error in a helper than a domain-specific struct.

kumarde: Got it.

kumarde:
YAY!

PR #134, opened 2017-08-28
Add bad character in label check

kumarde:


lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -0,0 +1,67 @@
...
+	return nil
+}
+
+func (l *DNSNameProperCharacters) CheckApplies(c *x509.Certificate) bool {
+	return true
+}
+
+func labelContainsBadCharacters(domain string) (bool, error) {
+	dnsLabelRegex := "^[A-Za-z0-9*_-]+$"
+	re, err := regexp.Compile(dnsLabelRegex)
```
zakird: This should be in init. There's no reason you need to compile this for every check.

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return true, err
+	}
```
zakird: This is not a lint level failure. There's nothing wrong with the certificate.

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -0,0 +1,69 @@
...
+		return err
+	}
+	return nil
+}
+
+func (l *DNSNameProperCharacters) CheckApplies(c *x509.Certificate) bool {
+	return true
+}
+
+func labelContainsBadCharacters(domain string, compiledExpression *regexp.Regexp) (bool, error) {
```
zakird: Why does this return an error? Error is always `nil`.

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -0,0 +1,69 @@
...
+	"strings"
+)
+
+type DNSNameProperCharacters struct {
+	// Internal data here
+	CompiledExpression *regexp.Regexp
+}
+
+func (l *DNSNameProperCharacters) Initialize() error {
+	dnsLabelRegex := "^[A-Za-z0-9*_-]+$"
```
zakird: This should be a constant.

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -0,0 +1,69 @@
...
+		if !compiledExpression.MatchString(label) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (l *DNSNameProperCharacters) RunTest(c *x509.Certificate) (ResultStruct, error) {
+	if c.Subject.CommonName != "" {
+		badCharacterFound, err := labelContainsBadCharacters(c.Subject.CommonName, l.CompiledExpression)
```
zakird: Why isn't this just a helper method that has access to `l.CompiledExpression`?

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -0,0 +1,69 @@
...
+	"regexp"
+	"strings"
+)
+
+type DNSNameProperCharacters struct {
+	// Internal data here
+	CompiledExpression *regexp.Regexp
+}
+
+func (l *DNSNameProperCharacters) Initialize() error {
```
zakird: The code base never actually checks whether this call returns an error.

dadrian: IMHO `Initialize()` should not return an error, similar to `func init()`. If some really bad happens, `panic()` internally. There's not much the framework could do with a non-nil error, anyway.

zakird: I'm fine with that. We just gotta do something about it.

kumarde: `MustCompile` does exactly this. I'll change to that.

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -0,0 +1,59 @@
...
+	const dnsLabelRegex = "^[A-Za-z0-9*_-]+$"
+	l.CompiledExpression = regexp.MustCompile(dnsLabelRegex)
+	return nil
+}
+
+func (l *DNSNameProperCharacters) CheckApplies(c *x509.Certificate) bool {
```
zakird: This seems like it should only apply to certificates with domains. I don't think this applies to CA names, etc.

PR #135, opened 2017-08-29
panic when initialize throws an error

kumarde:


lints/base.go
```diff
@@ -75,6 +75,9 @@ func RegisterLint(l *Lint) {
 	if Lints == nil {
 		Lints = make(map[string]*Lint)
 	}
-	l.Test.Initialize()
+	err := l.Test.Initialize()
+	if err != nil {
+		panic("could not initialize lint: " + l.Name)
```
zakird: Please output information about the error.

PR #142, opened 2017-08-30
Add dsa unique representation check

kumarde:
Can someone help me build a DSA cert with bad fields (for the test)? I can't figure out how to do this


zakird:
You could also try to search through Censys and find one.

PR #143, opened 2017-08-30
Add DSA Correct subgroup test 

kumarde:
Again, someone help me in making DSA certs?


dadrian:
`$  ~ openssl dsaparam -genkey 1024 -out dsakey.pem` will make you a key. Then just make a cert like normal.

kumarde:
@dadrian any chance I can mess with the fields, though? I'd like to test when things go bad for DSA keys.

dadrian:
@kumarde I think you can generate the parameters separately, and then generate the key from the parameters. https://github.com/google/der-ascii is probably your friend. Alternatively, just do this entirely in Golang using https://godoc.org/crypto/x509#CreateCertificate

lints/lint_dsa_correct_order_in_subgroup_test.go
```diff
@@ -0,0 +1,39 @@
...
+			"expected", desEnum, /* The enum you expected */
+			"got", out.Result, /* Actual Result */
+		)
+	}
+}
+
+func TestDSANotCorrectOrderSubgroup(t *testing.T) {
+	inputPath := "../testlint/testCerts/dsaCorrectOrderInSubgroup.pem"
+	c := ReadCertificate(inputPath)
+	dsaKey := c.PublicKey.(*dsa.PublicKey)
```
kumarde: :man_facepalming: I should have just done this. _sigh_

PR #144, opened 2017-08-30
Update README on how to add a new lint

kumarde:


README.md
```diff
@@ -47,6 +47,20 @@ ZLint can also be used as a library:
 
 See https://github.com/zmap/zlint/blob/master/cmd/zlint/main.go for an example.
 
+Contributing
+-------------
+
+If you would like to add a new x509 Lint:
+
+0. Make sure the lint doesn't already exist.
```
zakird: seriously?

zakird:
This really document really isn't helpful to get anyone getting started.

Do we have a template file for example to be copied in? Does `newLint.sh` work correctly? (If not, why doesn't it exist?)

How are you creating test certs? I've repeatedly heard that you're using Golang. Do you have a simple script that you can put in that let's get someone started?

PR #146, opened 2017-08-31
Add validity check to 39 months regardless of circumstance

kumarde:


util/time.go
```diff
@@ -30,6 +30,7 @@ var (
 	NoRSA1024Date              = time.Date(2014, time.January, 1, 0, 0, 0, 0, time.UTC)
 	GeneralizedDate            = time.Date(2050, time.January, 1, 0, 0, 0, 0, time.UTC)
 	NoReservedIP               = time.Date(2015, time.November, 1, 0, 0, 0, 0, time.UTC)
+	SubCert39Month             = time.Date(2016, time.June, 30, 0, 0, 0, 0, time.UTC)
```
dadrian: Are you going to read this value in `CheckApplies` or not?

kumarde: We put it in the `EffectiveDate` field, so it's Not Effective rather than Not Applicable

PR #156, opened 2017-09-01
Fix fatal errors for certificates with public suffixes

### LINKED ISSUE #155: PSL-related lints shouldn't fatal when encountering public suffixes
https://crt.sh/?id=202503674&opt=zlint

titanous:
- Use the ICANN section of the PSL for all relevant lints, as we want to lint the non-ICANN labels.
- Return NA instead of Fatal for PSL failures, the TLD lint will catch these if they are valid.

Fixes #155


zakird:
I don’t think this is quite right. if there’s an err in CN, we still probably want to check SAN for problems.

zakird:
Did you notice offhand what caused `hyphenFound, err := hyphenAtStartOrEndOfSLD(c.Subject.CommonName)` to error out?

zakird:
If this is a subscriber cert, is there any reason that we'd expect `util.ICANNPublicSuffixParse(domain)` to fail on it? Is it anything beyond a malformed FQDN? If a normal looking domain doesn't cause `ICANNPublicSuffixParse` to fail, and given that this is limited to subscriber certificates, I'm not sure if I understand why `NA` makes more sense than `FATAL`.

kumarde:
@zakird I think this was because `tuxfamily.org` is a private suffix, which we were erroneously including our public suffix check when we checked the domain `tuxfamily.com`.

kumarde:
@zakird My gut sense when we wrote this the first time was to use `NA`, because it's safer than FATAL if the `publicsuffix` test fails. If there is in fact other random things in the cert (for example, someone shoving not a domain name into the CN or SANs) we'd catch them in other DNS tests (like the one w/ bad characters, for example.) Thoughts?

titanous:
> I don’t think this is quite right. if there’s an err in CN, we still probably want to check SAN for problems.

Ah, yeah. I will fix that.

> I think this was because tuxfamily.org is a private suffix, which we were erroneously including our public suffix check when we checked the domain tuxfamily.com.

Correct.

> f there is in fact other random things in the cert (for example, someone shoving not a domain name into the CN or SANs) we'd catch them in other DNS tests (like the one w/ bad characters, for example.) Thoughts?

Yeah, I agree with this. I think we should skip the test and let the others error out instead of the throwing a fatal.

zakird:
@kumarde can you point to which lint should trigger if we can't parse the domain in the CN or SAN?

I'm not opposed to returning an `NA` in this situation if we're sure that it's caught somewhere else. Offhand, it's not clear from looking at this code that this happens though. It would be nice to make this more explicit. Could we move the "is this a domain" to `CheckApplies` and maybe add a comment that explains that this is checked elsewhere?

kumarde:
@zakird so, if it's _not_ a domain, one scenario might be that the CN is actually a name of something like "My Cool Org". In this case, the test `e_dnsname_bad_character_in_label` would catch that, since spaces would trigger an error in DNS names.

kumarde:
After a conversation w/ @zakird , I think we should keep this as a `Fatal`. Every library call error we catch in other lints also returns a `Fatal`, which make them worth investigating and can also help us find these kinds of bugs in the future.

zakird:
To be clear, I'm not convinced that this should be a Fatal! My concern is
primarily that we don't know if our other tests have coverage over what the
library checks. So there's the possibility that we lose something through
the cracks. This is why I'd say formalize what we _are_ checking as a
CheckApplies condition and let the library catch things we missed. There's
a chance this will Fatal. In actuality this isn't bad though, because we'll
know that it occurred.

kumarde:
What do you think would go in a CheckApplies for these lints? I still think it's hard a priori to know if something is a domain name, since there are so many checks we've implemented to test that, but if you've got suggestions I'm all ears. I think the CheckApplies should just be what we have now, which is if there's anything to check, and then we Fatal on library failures.

zakird:
This is going to be obnoxious, but could you run a check and see how often
this actually happens? This might inform what the right choice is.

In general, it would be quite helpful to know all of the Fatals that
currently come out of the censys dump in terms of debugging zlint.

kumarde:
@zakird I can do this, but it will take a little bit of time. I can have an answer by tomorrow morning.

kumarde:
Fatal numbers, out of 61M valid certificates (affects ~0.04% of certificates):
e_dnsname_hyphen_in_sld, 28602
e_dnsname_underscore_in_sld, 28602
w_dnsname_underscore_in_trd, 28602
w_dnsname_wildcard_left_of_public_suffix, 27350

titanous:
I suspect switching to the ICANN list as this PR does would fix all of these.

kumarde:
@titanous I believe you're correct based on a cursory glance on the data.

lints/lint_dnsname_underscore_in_trd.go
```diff
@@ -34,7 +34,7 @@ func (l *DNSNameUnderscoreInTRD) Execute(c *x509.Certificate) *LintResult {
 	if c.Subject.CommonName != "" {
 		underscoreFound, err := underscoreInTRD(c.Subject.CommonName)
 		if err != nil {
-			return &LintResult{Status: Fatal}
```
kumarde: We need to update to use the new helper function here as well.

util/fqdn.go
```diff
@@ -70,3 +72,7 @@ func DNSNamesExist(cert *x509.Certificate) bool {
 		return true
 	}
 }
+
+func ICANNPublicSuffixParse(domain string) (*publicsuffix.DomainName, error) {
+	return publicsuffix.ParseFromListWithOptions(publicsuffix.DefaultList, domain, &publicsuffix.FindOptions{IgnorePrivate: true})
```
kumarde: We need to add all the options or the library will return a segfault, so instead do this:
`return publicsuffix.ParseFromListWithOptions(publicsuffix.DefaultList, domain, &publicsuffix.FindOptions{IgnorePrivate: true, DefaultRule: publicsuffix.DefaultRule})`

dadrian:
So even after reading through the conversation here, I'm not entirely sure the exact details of what's going on, but I'll throw in my two cents:

- If we have a lint that checks aspects of a domain (e.g. the second-level domain is X), it should return `NA` on things that are not domains, and it should run this check in `CheckApplies`.
- Regardless of if CheckApplies returns `NA`, if in the `Execute()` function parsing something returns an error, we should return `Fatal`. Normally, this code path would not be reached if we also checked this in CheckApplies
- We should add a lint that enforces that we can parse domains in certificates that are for domains. This lint should return `Error` if something doesn't parse, but it should not enforce any other restrictions.

That being said, I'll leave final decisions to everybody else.

kumarde:
After much strife and debate, @zakird @dadrian and I have circled back to the very first commit to this PR, which is to `NA` instead of `Fatal`. I will add a _new_ lint that test if there is a bare IANA TLD in the cert, and we will be done.

kumarde:
@dadrian @zakird won't merge until either of you approve.

lints/lint_dnsname_underscore_in_trd.go
zakird: What's up with this?

PR #161, opened 2017-09-01
Propagate Fatal errors up

richsalz:
Argh.  This won't rebase, but hopefully it gives you some ideas.
Let me know if you want me to report to master.

### DIFF
--- a/lints/base.go
+++ b/lints/base.go
@@ -1,7 +1,9 @@
 package lints
 
 import (
+	"fmt"
 	"time"
+	"errors"
 
 	"github.com/zmap/zcrypto/x509"
 )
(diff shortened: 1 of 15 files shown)


zakird:
I like the idea of propagating the exact errors back to the user. It might make sense to add an Error object in the Result struct such that the error is returned next to the Fatal message and it's clear where it originated.

richsalz:
So you don't need me to do anything more about this?

richsalz:
Thinking about it a bit more, I would be cautious about putting an error object inside the status object. You don't want the higher-level function to clobber the lower-level one that actually found the error.  (In this case, the public suffix stuff perhaps)

The go style seems to return an err object, and I would make the lint routines do that on fatal error.  All the other code err does "if foo,err=...;err != nil" checking, so that seems the least invasive and more go-like.  YMMV.

dadrian:
I think we've removed most usages of `error` from ZLint now, given that in most cases Lint "errors" are really a status code. I think there's something to be said for rethinking the high-level API, but for now, I think we can close this PR, especially given that it conflicts with just about the whole repository.

PR #169, opened 2017-09-05
Add type to lints

kumarde:
Currently, our lints come from the BRs, RFC5280, are ZLint specific, or in some cases, AWSLabs cablint. Adding a field in each lint to keep track of them.


lints/base.go
```diff
@@ -42,6 +42,17 @@ type LintInterface interface {
 	Execute(c *x509.Certificate) *LintResult
 }
 
+// An Enum to programmatically represent the source of a lint
+type LintType int
```
dadrian: This isn't a `Type` it's a Source. But we set `Source` to a more detailed string, so maybe this is a category? Alternatively, rename `Source` to `ReadableSource` and and then change this to source. @zakird, thoughts?

lints/base.go
```diff
@@ -42,6 +42,17 @@ type LintInterface interface {
...
+
+const (
+	ZeroValue LintType = iota
+	CABFBaselineRequirements
```
dadrian: I'd rename this to `UnknownLintType`, except with whatever we rename `LintType` to.

lints/base.go
kumarde: We talked earlier today, "Type" was what we came up with. But I'm okay renaming `Source` to `ReadableSource`, and calling this `Source`.

zakird:
I think this should probably be source and citation now that I've had time
to think about this more.

kumarde:
Got it.

kumarde:
@dadrian @zakird updated.

lints/base_test.go
```diff
@@ -27,12 +27,20 @@ func TestAllLintsHaveNameDescriptionSource(t *testing.T) {
 		if lint.Description == "" {
 			t.Errorf("lint %s has empty description", name)
 		}
-		if lint.Source == "" {
+		if lint.Citation == "" {
 			t.Errorf("lint %s has empty source", name)
```
zakird: bad error message

lints/base_test.go
```diff
@@ -27,12 +27,20 @@ func TestAllLintsHaveNameDescriptionSource(t *testing.T) {
...
 		}
 	}
 }
 
+func TestAllLintsHaveType(t *testing.T) {
```
zakird: Lints don't have types

dadrian:
Names are fine now.

kumarde:
@dadrian I think I will remove this from the JSON, since Citation has the string already.

lints/base.go
```diff
@@ -56,7 +67,10 @@ type Lint struct {
 	Description string `json:"description,omitempty"`
 
 	// The source of the check, e.g. "BRs: 6.1.6" or "RFC 5280: 4.1.2.6".
-	Source string `json:"source,omitempty"`
+	Citation string `json:"citation,omitempty"`
+
+	// Programmatic source of the check, BRs, RFC5280, or ZLint
+	Source LintSource
```
dadrian: Need to add `json:"-"`

Issue #165, opened 2017-09-05
Use a distinct error message for certs with bad DNSNames due to leading ?.


alex:
These are found in Symantec's pre-certificates. These should use a distinct error message from other dnsname errors to make them easier to filter.

kumarde:
Merged.

PR #170, opened 2017-09-05
Add question mark notice lint

kumarde:
Returns a `Notice` if the commonName or SAN DNSNames has an item that is prepended with `?.`. 

Addresses https://github.com/zmap/zlint/issues/165


lints/lint_dnsname_contains_question_marks.go
```diff
@@ -0,0 +1,48 @@
+// lint_dnsname_contains_question_marks.go
```
dadrian: Delete this line, add a copyright notice.

kumarde: Copyright?

lints/lint_dnsname_contains_question_marks.go
```diff
@@ -0,0 +1,46 @@
...
+			return &LintResult{Status: Notice}
+		}
+	}
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "n_dnsname_contains_question_marks",
+		Description:   "Some Precerts are prepended with question marks.",
```
zakird: This has a term---"redacted". I'd make the name and description "n_contains_redacted_dnsname" and the source to be ZLint.

lints/lint_dnsname_contains_question_marks.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func isPrependedByQuestionMarks(domain string) bool {
+	if strings.HasPrefix(domain, "?.") {
+		return true
+	}
+	return false
+}
+
+func (l *DNSNameContainsQuestionMarks) Execute(c *x509.Certificate) *LintResult {
```
zakird: This should probably only run if `cn != ""`

zakird: This really only checks if this is prepended by _a_ question mark, not multiple question marks.

lints/lint_dnsname_contains_question_marks.go
```diff
@@ -0,0 +1,46 @@
...
+	"strings"
+)
+
+type DNSNameContainsQuestionMarks struct{}
+
+func (l *DNSNameContainsQuestionMarks) Initialize() error {
+	return nil
+}
+
+func (l *DNSNameContainsQuestionMarks) CheckApplies(c *x509.Certificate) bool {
```
zakird: This probably only applies to subscriber certificates.

lints/lint_dnsname_contains_question_marks.go
```diff
@@ -0,0 +1,46 @@
...
+	return true
+}
+
+func isPrependedByQuestionMarks(domain string) bool {
+	if strings.HasPrefix(domain, "?.") {
```
zakird: This function returns a bool...

util/fqdn.go
```diff
@@ -7,15 +7,15 @@ import (
 	"strings"
 )
 
-func removeQuestionMarks(domain string) string {
+func RemoveQuestionMarks(domain string) string {
```
zakird: This doesn't look like this handles multiple question marks. I think `?.?.domain` is an acceptable redacted certificate.

kumarde: it does - it's in a for loop.

lints/lint_dnsname_contains_question_marks.go
kumarde: done

lints/lint_dnsname_contains_question_marks.go
kumarde: done

lints/lint_dnsname_contains_question_marks.go
kumarde: done

util/fqdn.go
zakird: Then it should probably be named `removePrependedQuestionMark`. The function name implies that it removes multiple and that's not the case.

util/fqdn.go
```diff
@@ -7,15 +7,15 @@ import (
...
 	for strings.HasPrefix(domain, "?.") {
 		domain = domain[2:]
 	}
 	return domain
 }
 
 func IsFQDN(domain string) bool {
-	domain = removeQuestionMarks(domain)
+	domain = RemoveQuestionMarks(domain)
 	if strings.HasPrefix(domain, "*.") {
```
zakird: Why does one of these get a helper function when the second one doesn't? They have identical logic.

lints/lint_dnsname_contains_question_marks.go
kumarde: Yeah, I think that encompasses everything we want to check though? `?.?.abc.com` is still prepended with `?.` Maybe I'm not understanding this comment

zakird: The thing should really just be named "IsRedactedCertificate"

util/fqdn.go
kumarde: They don't - the first runs in a for loop. I can add a helper function if you'd like, but don't think we necessarily need it.

zakird:
```func isPrependedByQuestionMarks(domain string) bool {
	if strings.HasPrefix(domain, "?.") {
		return true
	}
	return false
}```

Comment (zakird) [reply util/fqdn.go]: ^^ That has the exact same logic there.

**Comment by titanous** [comment]
`*.?` is also used: https://crt.sh/?q=*.%3F.%25

**Comment by kumarde** [comment]
@zakird @titanous fixed up those comments.

File: util/fqdn.go [hunk 8]
```diff
@@ -7,18 +7,23 @@ import (
...
+func RemovePrependedQuestionMarks(domain string) string {
 	for strings.HasPrefix(domain, "?.") {
 		domain = domain[2:]
 	}
 	return domain
 }
 
-func IsFQDN(domain string) bool {
-	domain = removeQuestionMarks(domain)
+func RemoveWildcardFromDomain(domain string) string {
```
zakird: Can you name these two functions consistently?

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -33,13 +33,15 @@ func (l *DNSNameProperCharacters) labelContainsBadCharacters(domain string) bool
 
 func (l *DNSNameProperCharacters) Execute(c *x509.Certificate) *LintResult {
 	if c.Subject.CommonName != "" {
```
zakird: It doesn't look like this correctly handles `*.?.bad.com`

kumarde: AH, good catch.

util/fqdn.go
kumarde: done

lints/lint_dnsname_bad_character_in_label.go
kumarde: i fixed that btw.

lints/lint_dnsname_contains_question_marks.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "n_contains_redacted_dnsname",
+		Description:   "Some Precerts are prepended with question marks.",
+		Source:        CABFBaselineRequirements,
+		Citation:      "MDSP",
```
zakird: This citation doesn't make sense.

lints/lint_dnsname_contains_question_marks.go
```diff
@@ -0,0 +1,47 @@
+package lints
```
zakird: This file needs to be renamed.

lints/lint_is_redacted_cert.go
```diff
@@ -0,0 +1,47 @@
...
+			return &LintResult{Status: Notice}
+		}
+	}
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "n_contains_redacted_dnsname",
+		Description:   "Some Precerts are prepended with question marks.",
```
zakird: This description, source, and citation don't make much sense.

Issue #167, opened 2017-09-05
Decide how to handle non-server certificates


robstradling:
Consider the following example certs:
https://crt.sh/?id=204352566&opt=zlint (a Client Authentication certificate)
https://crt.sh/?id=201186679&opt=zlint (an OCSP Signing certificate)
https://crt.sh/?id=122792187&opt=zlint (a Code Signing certificate)
https://crt.sh/?id=158872566&opt=zlint (an EV Code Signing certificate)

Currently ZLint "checks for consistency with RFC 5280 and the CA/Browser Forum Baseline Requirements (v.1.4.8)", so it's unsurprising that various warnings/errors/fatals are triggered for these certs that aren't intended to comply with the CABForum BRs.

Should ZLint be enhanced to somehow detect the certificate type and only run CABForum BR lints on certs that are actually expected to be CABForum BR compliant?  Should ZLint even be further enhanced to check for consistency with other standards (e.g., the EV Code Signing Guidelines, the Code Signing BRs published by CASC, the Google S/MIME profile, the various browser root program policies, etc).

Or, should it be up to the user/caller (e.g., crt.sh) to make sure that ZLint is only used/called for certs that are expected to be CABForum BR compliant?

zakird:
I think it makes sense to update ZLint to detect certificate scope and to execute the correct certificates. I think there are likely two parts of this:

(1) Deciding what logic we want to use to determine whether a certificate falls under CA/B purview. 

(2) Marking lints with their scope.

It sounds like there isn't a clear definition for (1). @pzb had suggested limiting scope to certificates with Server Auth and All auth. If everything agrees that this is the right setting, @kumarde do you want to work on getting that in place?

robstradling:
Where "All auth" means that the EKU extension is either: (1) absent; or (2) present and contains anyExtendedKeyUsage?  SGTM.

zakird:
@robstradling yep

zakird:
Notes:
(1) add programmatic source field. Three options to start: (1) BR, (2) RFC5280, (3) ZLint

kumarde:
Merged.

PR #171, opened 2017-09-06
Fix eku check for BRs

kumarde:
Addresses https://github.com/zmap/zlint/issues/167


util/eku.go
```diff
@@ -0,0 +1,17 @@
+package util
```
dadrian: May as well just make this part of `ca.go`. https://github.com/zmap/zlint/blob/master/util/ca.go

lints/base.go
```diff
@@ -97,6 +98,9 @@ func (l *Lint) CheckEffective(c *x509.Certificate) bool {
 // CheckEffective()
 // Execute()
```
dadrian: Can you update the first paragraph of the above comment to describe the additional checks you just added, e.g. lints that are checking for BR compliance are only ran against certs that fall under the purview of the BR's.

zakird:
@titanous can you take a look at this too?

util/ca.go
```diff
@@ -29,3 +29,15 @@ func IsSelfSigned(c *x509.Certificate) bool {
 func IsSubscriberCert(c *x509.Certificate) bool {
 	return !IsCACert(c) && !IsSelfSigned(c)
 }
+
+func IsTestableBRCertificate(cert *x509.Certificate) bool {
```
titanous: Perhaps this should be renamed? It checks if a certificate can be used for ServerAuth, so `IsServerAuthCert` or similar?

kumarde: I can do that.

kumarde:
@titanous i've updated the name.

PR #172, opened 2017-09-07
Add bare IANA Suffix lint

kumarde:


lints/lint_dnsname_contains_bare_iana_suffix.go
```diff
@@ -0,0 +1,43 @@
+// lint_dnsname_contains_bare_iana_suffix.go
```
dadrian: Cut this out.

And put in the license header.

lints/lint_dnsname_contains_bare_iana_suffix.go
```diff
@@ -0,0 +1,43 @@
...
+	return nil
+}
+
+func (l *dnsNameContainsBareIANASuffix) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.DNSNamesExist(c)
+}
+
+func (l *dnsNameContainsBareIANASuffix) Execute(c *x509.Certificate) *LintResult {
+	if c.Subject.CommonName != "" {
+		if util.IsInTLDMap(c.Subject.CommonName) {
```
dadrian: Why do we use the map instead of the parse function in `util`

kumarde: I thought you wanted IANA suffix not ICANN Public Suffix? 

The `util.ICANNPublicSuffixParse` is just the public part of the public suffix

lints/lint_dnsname_contains_bare_iana_suffix.go
kumarde: 👍

lints/lint_dnsname_contains_bare_iana_suffix.go
dadrian: OK, but still let @zakird do a pass

robstradling:
I think this lint is checking the PRIVATE section of the PSL:
https://crt.sh/?zlint=908&iCAID=16418

Shouldn't it only check the ICANN section?

kumarde:
Ah, that's not this lint, but the lint: `w_san_iana_pub_suffix_empty`. I think this doesn't use the correct `publicsuffix` package (as per other recent updates). I'll update this.

robstradling:
@kumarde Thanks!  With so many similar lints, it's easy to get them confused.  ;-)

PR #174, opened 2017-09-15
Fix san bare suffix, ca CN effective time, subCA EKU 

kumarde:
Updates three lints:

- Uses `util.ICANNPublicSuffixParse` for the SAN bare suffix test
- Uses the correct `EffectiveDate` for the commonName test for CAs
- Reduces the error to a notice if EKU is missing in subCA, required only for technically constrained CAs.


kumarde:
@zakird fixes the outstanding bugs.

zakird:
I haven't quite been able to follow all the changes around the public suffix list. I suspect that @titanous or Rob are going to be in a better spot to take a look.

PR #179, opened 2017-11-07
Speed up the linters

szank:
This pull request will speed up the slowest linters. 

* w_dnsname_underscore_in_trd
* e_dnsname_underscore_in_sld
* e_dnsname_hyphen_in_sld
* w_dnsname_wildcard_left_of_public_suffix
* w_san_iana_pub_suffix_empty
* w_rsa_mod_factors_smaller_than_752
* e_dnsname_bad_character_in_label

Plus some linting done by my IDE. 

Benchmark results
```
benchcmp old.benchmarks.txt new.benchmarks.txt
benchmark                      old ns/op     new ns/op     delta
BenchmarkZlint/All_lints-8     33228743      407902        -98.77%

benchmark                      old allocs     new allocs     delta
BenchmarkZlint/All_lints-8     30006          2115           -92.95%

benchmark                      old bytes     new bytes     delta
BenchmarkZlint/All_lints-8     18835646      90210         -99.52%
```

Overall, the benchmarks are ~80x faster. 
If you guys are happy with the changes, I might have a look at some other slow benchmarks. 

* w_subject_dn_leading_whitespace
* w_subject_dn_trailing_whitespace
* w_multiple_subject_rdn
* e_ext_san_dns_not_ia5_string
* e_ext_san_empty_name
* e_dnsname_not_valid_tld
* e_dnsname_contains_bare_iana_suffix
* e_dnsname_wildcard_only_in_left_label
* e_international_dns_name_not_nfkc

Again, speeding them up might require some changes to the zcrypto/x509 to cache some values (tentatively)

I would appreciate any remarks.

### DIFF
--- a/benchmarks_test.go
+++ b/benchmarks_test.go
@@ -1,0 +1,159 @@
+package zlint
+
+import (
+	"encoding/pem"
+	"testing"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/lints"
+)
+
+var globalLintResult *ResultSet
+var globalSingleLintResult *lints.LintResult
+var globalLintResultList = []*lints.LintResult{}
+
+const bigCertificatePem = `-----BEGIN CERTIFICATE-----
+MIILajCCClKgAwIBAgIMOp/m5bdkZ2+oPevRMA0GCSqGSIb3DQEBCwUAMGIxCzAJ
+BgNVBAYTAkJFMRkwFwYDVQQKExBHbG9iYWxTaWduIG52LXNhMTgwNgYDVQQDEy9H
+bG9iYWxTaWduIEV4dGVuZGVkIFZhbGlkYXRpb24gQ0EgLSBTSEEyNTYgLSBHMzAe
+Fw0xNzA2MjIwNjU2MDNaFw0xOTA2MjMwNjU2MDNaMIH9MR0wGwYDVQQPDBRQcml2
+YXRlIE9yZ2FuaXphdGlvbjEPMA0GA1UEBRMGNTc4NjExMRMwEQYLKwYBBAGCNzwC
+AQMTAlVTMR4wHAYLKwYBBAGCNzwCAQITDU5ldyBIYW1wc2hpcmUxCzAJBgNVBAYT
+AlVTMRYwFAYDVQQIEw1OZXcgSGFtcHNoaXJlMRMwEQYDVQQHEwpQb3J0c21vdXRo
+MSAwHgYDVQQJExdUd28gSW50ZXJuYXRpb25hbCBEcml2ZTEdMBsGA1UEChMUR01P
+IEdsb2JhbFNpZ24sIEluYy4xGzAZBgNVBAMTEnd3dy5nbG9iYWxzaWduLmNvbTCC
+ASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAKaVk8nelrMqQmTSBju68D8B
+MO7GGHtuQU8bfvuGNTUe6HiAxHYRB+LfCVAoTMXRtKgiI2YnTQ7xedKCaGTo2ZLH
+y58Ya4ASpFgGLS3sPLjIHCP68ck126efksscXl2vBVWGGS7a0oTGLaaonFkz4FFy
+0SkSwCL9UPPKkpVoQQ48kOF+tKZx1RimoZbZC9BwXtZYjdIbL9EzineymyJGsMi4
+5utV7zQfcbZj3V9j4TAcx6UwqdwlmF7FVQ3Q1YmFtOZy6/U44us/Oz4SJ2+FIWS3
+fZ6oGXBh3qq3L4n7ixiNpuj+CZmAJP8VM7w1dSquJ9ndw6Lid0jKIpY6nlDfflkC
+AwEAAaOCB4Iwggd+MA4GA1UdDwEB/wQEAwIFoDCBlgYIKwYBBQUHAQEEgYkwgYYw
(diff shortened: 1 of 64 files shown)


benchmarks_test.go
```diff
@@ -0,0 +1,159 @@
...
+eVcCW12pFbLQxLns+dhExFvZBfy2iewouKo8Q41tolmEv4A3ADNuv+3r1bYhTnzE
+55e0GMvnRIz5zQ7JWBTuamWNFI4OccJVh7vt0dnrSgiXs+XJ89qmgDyc/DikdM4q
+psw2SW2R/SwSnkgvaLM/o0tw77aapxlaAs29Y4SE/RvRR2CJ0V/gvq9GUorY4OF2
+2HEky394KiGDZDYUUDArx2+9w+yPikV5llF7lm2o84kZifnBO6SE9+4zdBExzg==
+-----END CERTIFICATE-----
+`
+
+func BenchmarkZlint(b *testing.B) {
+	var certDerBlock, _ = pem.Decode([]byte(bigCertificatePem))
+	var x509Cert, err = x509.ParseCertificate(certDerBlock.Bytes)
```
titanous: The Go idiom is to use `:=` instead of `var` when initializing a variable and assigning a value at the same time (there are similar definitions that should probably be fixed throughout the patch).

benchmarks_test.go
```diff
@@ -0,0 +1,159 @@
...
+	"encoding/pem"
+	"testing"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/lints"
+)
+
+var globalLintResult *ResultSet
+var globalSingleLintResult *lints.LintResult
+var globalLintResultList = []*lints.LintResult{}
```
titanous: Is there a reason why these can't be in the benchmark function body?

lints/lint_distribution_point_missing_ldap_or_uri.go
```diff
@@ -7,9 +7,10 @@ When present, DistributionPointName SHOULD include at least one LDAP or HTTP URI
 package lints
 
 import (
+	"strings"
+
```
titanous: Ideally these separate whitespace changes wouldn't be in the same commit but it's not a big deal.

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -1,50 +1,36 @@
...
 	CompiledExpression *regexp.Regexp
 }
 
 func (l *DNSNameProperCharacters) Initialize() error {
-	const dnsLabelRegex = "^[A-Za-z0-9*_-]+$"
-	l.CompiledExpression = regexp.MustCompile(dnsLabelRegex)
-	return nil
+	const dnsNameRegexp = `^(\*\.)?(\?\.)?(A-Za-z0-9*_-]+\.)*[A-Za-z0-9*_-]*$`
+	var err error
+	l.CompiledExpression, err = regexp.Compile(dnsNameRegexp)
```
titanous: This regex can be moved to a global variable assigned to the result of `MustCompile`, which has the side effect of ensuring a panic when the program starts if there is a typo in the pattern.

lints/lint_dnsname_bad_character_in_label.go
```diff
@@ -1,50 +1,36 @@
...
 
 type DNSNameProperCharacters struct {
 	CompiledExpression *regexp.Regexp
 }
 
 func (l *DNSNameProperCharacters) Initialize() error {
-	const dnsLabelRegex = "^[A-Za-z0-9*_-]+$"
-	l.CompiledExpression = regexp.MustCompile(dnsLabelRegex)
-	return nil
+	const dnsNameRegexp = `^(\*\.)?(\?\.)?(A-Za-z0-9*_-]+\.)*[A-Za-z0-9*_-]*$`
```
titanous: There can be more than one "redacted" label: https://crt.sh/?id=250738174

szank: I will update the regex. 
You are aware though that the master branch code doesn't handle this case either?
I was refactoring the code without introducing any functional changes.

titanous: Isn't it handled by `util.RemovePrependedQuestionMarks`?

szank:
```
func RemovePrependedQuestionMarks(domain string) string {
	for strings.HasPrefix(domain, "?.") {
		domain = domain[2:]
	}
	return domain
}
```
Doesn't it remove only the first occurrence of `?.` ?

benchmarks_test.go
szank: It is my muscle memory. Changed where applicable.

lints/lint_distribution_point_missing_ldap_or_uri.go
szank: Yeah. Sorry about that. My IDE have done it automatically. If you want it in separate commits, it can be done, but would have to close this PR and open a new one.

lints/lint_dnsname_bad_character_in_label.go
szank: Fixing now.

lints/lint_dnsname_bad_character_in_label.go
szank: In my book, `Initialize()` returns an error because can fail like in this case. What is the point of returning an error from the function if prefered mode of handling an error is panicking?

It is your repo, and I will do what guys ask me to do, but I don't think that panicking is a better way to handle errors than returning an error. 

Additionally the init func calls `RegisterLint()`, which in turn calls that lint's `Initialize()` and panics on error returned from `Initialize()`

So yeah, the code panics anyway if regex could not be compiled.

szank:
Hi, 
I just wanted to ping you guys. Did you have a chance to have a look at the updated PR ?

PR #182, opened 2017-11-23
Enhance e_ext_san_contains_reserved_ip linter to check for more reserved IP ranges, reject 0.0.0.0 IP address

szank:
1.  lints/lint_ext_san_contains_reserved_ip.go checked if input IP address != 0.0.0.0 before 
proceeding with the reserved range checks, while the IP 0.0.0.0 is in the reserved range. 
Hence when the input IP is 0.0.0.0 the linter returns `Pass` status

2. util/ip.go:
`{net.ParseIP("100.64.0.0"), net.ParseIP("10.127.255.255")},` 
contained a typo. 

Additionally, the checks were missing few reserver ranges.


szank:
Hi guys. Sorry for the two commits that broke the CI.

szank:
Good points. 
For starters I have just fixed what was missing in the currently available linters, would be nice to maybe add few additional linters to check for those corner cases. 

Also IIRC in the zmap/zcrypto/x509 package, common name is a string, not a slice, so for certificates with multiple CNs the subsequent ones are ignored, or the parser returns error if the sequence has more than one element. 

There _should_  be a lint that examines the raw subject and catches such cases also.

justinbastress:
@szank right -- the way I read the code, it takes the last (so, most-specific) CN in the sequence, which by my reading is indeed what the browsers/etc are _supposed_ to use in absence of an SAN, but IIRC at least some browsers do allow multiple CNs.

It seems like the library does expose the raw RDNSequence, so one _could_ walk through that to get all of the CNs, but it would be strange to do that here but nowhere else CNs are consumed.

szank:
Hey, I just wanted to make sure that there are no more actions points in this PR that I could address before the merge is approved. Could you guys please give me an update?

justinbastress:
@szank I don't see any problems.

dadrian:
With regard to @justinbastress's original comment about repeated CN's, those show up in `Subject.ExtraNames`, IIRC.

I suspect the proper way to handle repeats is to extract anything that runs checks on the CN to some utility function that accounts for ExtraNames (or to just error on ExtraNames in another lint, which we might do already?). I don't think that blocks this PR.

PR #184, opened 2017-12-07
Add new lints for RFC 5280, fix existing ones

szank:
Hi, 
I have added a few more lints for RFC 5280 and fixed few existing ones.  I am quite confident that this time there is not rouge change that would cause false positives. 
OTOH, few certs that have been flagged OK previously might now turn out invalid.


lints/lint_ext_ian_no_entries.go
```diff
@@ -30,7 +32,7 @@ func (l *IANNoEntry) CheckApplies(c *x509.Certificate) bool {
 
 func (l *IANNoEntry) Execute(c *x509.Certificate) *LintResult {
 	ian := util.GetExtFromCert(c, util.IssuerAlternateNameOID)
-	if (ian.Value)[1] == 0 {
+	if len(ian.Value) < 2 || bytes.Equal(ian.Value, []byte{0x30, 0x00}) { // if ian is an empty octet string (len == 0 ) or an empty sequence ( 0x30 0x00 ) return error
```
titanous: I think it might make sense to extract this out into a helper function given that it's used in a few places?

lints/lint_issuer_dn_country_not_printable_string.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+type IssuerDNCountryNotPrintableString struct{}
+
+func (l *IssuerDNCountryNotPrintableString) Initialize() error {
+	return nil
+}
+
+func (l *IssuerDNCountryNotPrintableString) CheckApplies(c *x509.Certificate) bool {
+	return len(c.Issuer.Country) > 0
```
titanous: Is it technically possible to have a 0-length `Issuer.Country` here but have a field that is not a `PrintableString` without an error?

lints/lint_ext_ian_no_entries.go
szank: Yeah, could make a helper IsEmptySequence()

lints/lint_issuer_dn_country_not_printable_string.go
szank: If I have understood the question correctly, then yes. 
As far as code organisation goes here, every lint have a very limited scope. So I didn't check here for empty sequences. 
There is a check if the Country field is valid (value is on a list of country codes) and this check should fire off if the field itself is empty. I am not sure if that lint verifies both issuer country and subject country or just subject country.

titanous: Okay, that makes sense.

lints/lint_ext_ian_no_entries.go
titanous: Sounds good, this will be good to merge after that change.

util/encodings.go
```diff
@@ -90,3 +91,13 @@ func AllAlternateNameWithTagAreIA5(ext *pkix.Extension, tag int) (bool, error) {
 
 	return true, nil
 }
+
+// IsEmptyASN1Sequence returns true if
+// *input is an empty sequence (0x30, 0x00) or
+// *len(inout) < 2
+// This check covers more cases than just empty sequence checks but it makes sense from the usage perspective
+var emptyASN1Sequence = []byte{0x30, 0x00}
```
zakird: This can probably be marked `const`.

titanous: slices/maps/pointers can't be marked as constant.

zakird: ah hah! Thanks.

lints/lint_ext_san_no_entries.go
```diff
@@ -32,7 +30,7 @@ func (l *SANNoEntry) CheckApplies(c *x509.Certificate) bool {
 
 func (l *SANNoEntry) Execute(c *x509.Certificate) *LintResult {
 	san := util.GetExtFromCert(c, util.SubjectAlternateNameOID)
-	if len(san.Value) < 2 || bytes.Equal(san.Value, []byte{0x30, 0x0}) { // if san is an empty octet string (len == 0 ) or an empty sequence ( 0x30 0x00 ) return error
+	if util.IsEmptyASN1Sequence() {
```
titanous: It looks like this and the other use of the function is missing the argument.

titanous:
Thanks!

PR #188, opened 2017-12-15
Allow capitalization in the Subject DN CommonName when comparing it w…

### LINKED ISSUE #186: Certs with CN containing capital letters flagged by `e_subject_common_name_not_from_san` false positive
Example: 
https://crt.sh/?id=276876975&opt=cablint,x509lint,zlint

No other linters in crt.sh are flagging this cert, and as far as I can tell, DNS names should be normalised before parsing/validation. Either way pasting `Tours.AlliedTT.com` to chrome works, as the dns name is automatically converted to lowercase (possibly by the IDNA engine)

The `e_subject_common_name_not_from_san` should perform ToLower on the common name before comparing it with SAN DNS names IMHO. Any objections?

szank:
…ith the SubjectAlternativeNames DNS Name

Closes #186


lints/lint_subject_common_name_not_from_san.go
```diff
@@ -25,9 +27,15 @@ func (l *subjectCommonNameNotFromSAN) CheckApplies(c *x509.Certificate) bool {
...
 	cn := c.Subject.CommonName
+	cnLowerCase := strings.ToLower(cn)
 
 	for _, dn := range c.DNSNames {
-		if cn == dn {
+		if cnLowerCase == dn {
+			if cnLowerCase != cn {
+				return &LintResult{
+					Status:  Pass,
+					Details: "CommonName is capitalised and hence not byte for byte equivalent to the DNS name SAN"}
```
titanous: I think this should actually be a separate Notice-level lint? /cc @zakird

zakird:
Yes, I agree.

szank:
Hi guys, is there anything more to do in this PR? If not, could you please approve it ?
Thanks.

zakird:
It looks like this lint returns one of either a notice or an error, which isn't something we currently allow. Per @titanous's comment, we'll want to move the capitalization check to a separate lint.

szank:
Ah, sorry I misunderstood your comments. I will create a separate linter.

szank:
Hi, 
I have realised that I would be making few assumptions regarding the new lint, and would like to get your opinions on them. 
*.Lint name : lint_subject_common_name_not_lowercase - I think this doesn't require much discussion.
* Lint source : `CABFBaselineRequirements`. I think this is the correct one. Or `Zlint`. Opinions ?
* Lint effective date : util.CABEffectiveDate or zero time accordingly to the lint source 
* Clitation: `BRs: 7.1.4.2.2` or empty, again accordingly with the lint source 
* lint:CheckApplies() I will use 
```
func (l *subjectCommonNameNotLowercase) CheckApplies(c *x509.Certificate) bool {
	return c.Subject.CommonName != "" && !util.IsCACert(c)
}
```
* the Execute() body :
``` 
if c.Subject.CommonName != strings.ToLower(c.Subject.CommonName){
			return &LintResult{Status: Notice}
}
return &LintResult{Status: Pass}

```
? I think this is sufficient here. 
I would be grateful for some input so avoid going back and forth. 
Also, do you guys have more pointers on what is allowed and what is not?
I was not aware that there were rules against putting Pass, Notice and Error statuses in single linter. 
Is there anything else I should be aware of. It is quite funny that I have bumped into this problem now and not earlier.

zakird:
Sorry for the delay in getting back to you on this. 

Source: ZLint probably makes the most sense given that this isn't explicitly defined in the BRs. The rest of your notes make sense to me.

Regarding name, the name of the file as `lint_subject_common_name_not_lowercase` makes sense. we've been prefixing the names in the lints with the type of lint (e.g., notice, warning, etc). So this lint would be registered with the name `n_lint_subject_common_name_not_lowercase`.

The reason we've been doing this isn't really obvious, I agree, but effectively boils down to being able to know how severe a failure is based on just a boolean value instead of needing to store a more expensive value on all certificates.

zakird:
Whoops, I accidentally forgot to put a link to the documentation I meant to include, which explains how to create a new lint https://github.com/zmap/zlint#adding-new-lints.

szank:
Hi,
I have updated the PR. Hope all is well now. 
Also, have a happy new year!

lints/lint_subject_common_name_not_from_san.go
```diff
@@ -24,16 +26,16 @@ func (l *subjectCommonNameNotFromSAN) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *subjectCommonNameNotFromSAN) Execute(c *x509.Certificate) *LintResult {
-	cn := c.Subject.CommonName
+	cnLowerCase := strings.ToLower(c.Subject.CommonName)
 
 	for _, dn := range c.DNSNames {
-		if cn == dn {
+		if cnLowerCase == dn {
```
dadrian: Are we sure that DNSNames will always be lower case?

szank: Actually, no. That's a good point. I haven't seen capitalised domain name in the SAN DNS in the wild, and I had a look at the relevant RFCs ( IDNA related ) and could not find any requirements for having the domain names in lower case. I have found some information leading me to the conclusion that it should not matter. 

I will update the PR. 
Also, while going through the RFCs I have noticed this passage in the rfc 3490: 
`The ACE prefix for IDNA is "xn--" or any capitalization thereof.`

This is not how the `e_international_dns_name_not_unicode` is working. 
```
[...]
if strings.HasPrefix(label, "xn--") {
	_, err := idna.ToUnicode(label)
	if err != nil {
	}
}
[...]
```
This test is case sensitive.  What i have also found is that golang's idna package doesn't handle this case properly either. 
```
XN--FIQS8S.ICOM.MUSEUM -> XN--FIQS8S.ICOM.MUSEUM [no error]
xn--fiqs8s.icom.museum -> 中国.icom.museum [no error]
```
I think it warrants another PR for zlint. 

I would have to investigate if the golang package doesn't have a bug. I will take a working assumption that I am missing something, bug is rather unlikely. Chrome and curl  handles XN--FIQS8S.ICOM.MUSEUM properly. 
Curl sends Host: `XN--FIQS8S.ICOM.MUSEUM` header so the webserver is handling the case properly also ( this excludes any mangling by Chrome itself. )

lints/lint_subject_common_name_not_lowercase.go
```diff
@@ -0,0 +1,44 @@
...
+		return &LintResult{Status: Notice}
+	}
+
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "n_lint_subject_common_name_not_lowercase",
+		Description:   "The common name field in subscriber certificates must include only names from the SAN extension",
```
zakird: This looks like the incorrect description.

szank:
Hi guys. I think I have found another wrinkle that needs to be addressed somehow. 

Common name (or SAN, if it have not been converted to punycode) might contain UTF-8 characters. 
Let's assume that common name is say, `\u212A.com` where \u212A is Kelvin symbol K) 
and the SAN DNS name is `\u006B.com` where \u006B is ASCII k. For go case insensitive comparison the strings are identical. 

https://play.golang.org/p/Sqo_dZd88QU

Now, `\u212A.com` and  `\u006B.com` are completely different domain names, but the linter would return PASS if the SAN DNS list contained only the first name and the CommonName contained the other. 

I would suggest that both common name and dns names needs to be converted to punycode before comparison to avoid this issue. 

As far as I can tell there are no rules regarding having the common name containing UTF-8 characters encoded as punycode in the certificate so problems seems real to me. 

Opinions ?

zakird:
I'll be the first to admit, I don't have a good sense of how this should be handled, nor what's in the BRs. @robstradling, @titanous do have any thoughts offhand?

robstradling:
@zakir SAN.dNSName is defined as an ASN.1 IA5String, so it can't hold a U-label.
Whether or not the Subject.CN is permitted to hold a U-label that corresponds to a SAN.dNSName's A-label has been a topic of debate.  The BRs say the CN should hold a "`Fully‐Qualified Domain Name that is one of the values contained in the Certificate’s subjectAltName extension`".  But are a U-label and the corresponding A-label different "values"?  Or are they different encodings of the same value?  CABForum Ballot 202 from @pzb sought to clarify this issue (amongst other things) by explicitly requiring CNs to contain A-labels, but unfortunately the ballot narrowly failed (because certain CAs want to put U-labels in CNs).
So the ambiguity persists.

dadrian:
Are there any pairs of Unicode code points where:
- They have different punycode encodings
- A case insensitive comparison rightly considers them to be the same

My worry with converting everything to punycode would be that we'd start hitting false positives where punycode comparison is different, but should be considered the same in a "normal" case insensitive comparison.

szank:
@dadrian I will try to prepare some examples tomorrow (I don't have a working go installation on this machine and go playground doesn't include the idna package), but I am quite certain that there are. 
Also, thanks @robstradling for the concise summary of the issue. 

IMHO an U-label and A-label are considered the same value if an DNS resolver (or a browser) treats them as identical.  
If a browser converts U-label to an A-label before sending an DNS query, then I would consider the input A-label and the sent U-label identical. 

AFAIK the conversion from U-labels to A-labels is done in the browser before sending out the DNS request.

zakird:
In #186, I have proposed that we take a more conservative route and close this PR without merging. If anyone has thoughts, please add them to that issue.

Issue #191, opened 2018-01-02
Add a lint for the upcoming March 1st, 2018 825 day cert validity limit


alex:

justinbastress:
See https://github.com/zmap/zlint/pull/193

PR #193, opened 2018-01-08
Issue #191: 825 day certificates

justinbastress:
Add new lint_sub_cert_valid_time_longer_than_825_days to implement the language of https://cabforum.org/2017/03/17/ballot-193-825-day-certificate-lifetimes/ -- namely, subscriber certificates issued after 2018/03/01 must have a validity period no longer than 825 days.

The existing lint_sub_cert_valid_time_too_long already handles the 39-month requirement (this was renamed to lint_sub_cert_valid_time_longer_than_39_months to remove the implication that it implements the whole "valid_time_too_long" requirement).

### DIFF
--- a/lints/lint_sub_cert_valid_time_longer_than_39_months.go
+++ b/lints/lint_sub_cert_valid_time_longer_than_39_months.go
@@ -1,4 +1,4 @@
-// lint_sub_cert_valid_time_too_long.go
+// lint_sub_cert_valid_time_longer_than_39_months.go
 
 package lints
 
@@ -7,17 +7,17 @@ import (
 	"github.com/zmap/zlint/util"
 )
 
-type subCertValidTimeTooLong struct{}
+type subCertValidTimeLongerThan39Months struct{}
 
-func (l *subCertValidTimeTooLong) Initialize() error {
+func (l *subCertValidTimeLongerThan39Months) Initialize() error {
 	return nil
 }
 
-func (l *subCertValidTimeTooLong) CheckApplies(c *x509.Certificate) bool {
+func (l *subCertValidTimeLongerThan39Months) CheckApplies(c *x509.Certificate) bool {
 	return util.IsSubscriberCert(c)
 }
 
-func (l *subCertValidTimeTooLong) Execute(c *x509.Certificate) *LintResult {
+func (l *subCertValidTimeLongerThan39Months) Execute(c *x509.Certificate) *LintResult {
 	if c.NotBefore.AddDate(0, 39, 0).Before(c.NotAfter) {
 		return &LintResult{Status: Error}
 	}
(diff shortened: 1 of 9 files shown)


lints/lint_sub_cert_valid_time_too_long.go
```diff
@@ -18,16 +19,32 @@ func (l *subCertValidTimeTooLong) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *subCertValidTimeTooLong) Execute(c *x509.Certificate) *LintResult {
-	if c.NotBefore.AddDate(0, 39, 0).Before(c.NotAfter) {
-		return &LintResult{Status: Error}
+	july_1_2016 := time.Date(2016, time.July, 1, 0, 0, 0, 0, time.UTC)
```
kumarde: you can add times in `util/time.go`

lints/lint_sub_cert_valid_time_too_long.go
```diff
@@ -18,16 +19,32 @@ func (l *subCertValidTimeTooLong) CheckApplies(c *x509.Certificate) bool {
...
 }
 
 func init() {
 	RegisterLint(&Lint{
 		Name:          "e_sub_cert_valid_time_too_long",
-		Description:   "CAs MUST NOT issue subscriber certificates with validity periods longer than 39 months regardless of circumstance.",
+		Description:   "Subscriber Certificates issued after 1 March 2018 MUST have a Validity Period no greater than 825 days. Subscriber Certificates issued after 1 July 2016 but prior to 1 March 2018 MUST have a Validity Period no greater than 39 months. ",
 		Citation:      "BRs: 6.3.2",
 		Source:        CABFBaselineRequirements,
 		EffectiveDate: util.SubCert39Month,
```
kumarde: This will make this only run for certs issued after this time, which probably also needs to be changed

justinbastress:
Thanks @kumarde — that sounds right, and it would definitely be cleaner. My only hesitatation would be having two lints for the same rule (6.3.2), but cross-referencing them in the comments should make the situation clear enough. I’ll get this taken care of in the morning.

alex:
Tests in `lints/lint_sub_cert_valid_time_too_long_test.go` need to be updated

lints/lint_sub_cert_valid_time_longer_than_39_months.go
```diff
@@ -0,0 +1,36 @@
+// lint_sub_cert_valid_time_longer_than_39_months.go
```
dadrian: ugh why do we still insist on putting useless comments at the top of files? I thought I went through and deleted all of these. @justinbastress, not your fault for matching to what have, but we should either actually document the lint here or just omit the comment.

zakird: I agree with @dadrian, this seems like something that can (and probably should!) go.

justinbastress:
@zakird @dadrian I can put together a quick script to strip these comments and do a new PR with the results.

lints/lint_sub_cert_valid_time_longer_than_39_months.go
justinbastress: I can put together a quick script to strip these and push a new PR with the results.

justinbastress:
@alex fixed the 39-month test I broke with the rename, added 825-day tests as well.

kumarde:
this LGTM.

zakird:
This looks good to me. @dadrian, @alex want to give a chance to note anything given your involvement in the ticket. If it looks good to you two, let's merge it!

alex:
From a quick skim, it looks good to me!

I assume this will end up deployed before March 1st so we can immediately capture non-compliant CAs?

zakird:
I'll make a note in my calendar to make sure we get this deployed by then.

robstradling:
@alex The crt.sh servers are now running https://github.com/zmap/zlint/commit/80fe8eced4b4df54e7e9b70e52bc6aa273d03a54, so https://crt.sh/?zlint=issues and https://misissued.com/zlint/ are both now ready to capture non-compliant CAs.

Issue #198, opened 2018-01-12
Lint e_ext_san_uri_host_not_fqdn_or_ip returning an error on valid uris


szank:
Hi, this lint doesn't work like intended. Given a valid URI, for example `sip:user@example.domain.com` it returns an error. 

Problem is in the util.GetHost() method, fortunately as far as I can tell it is not used elsewhere:
lint_ext_san_uri_host_not_fqdn_or_ip_test.go:
```
if uri != "" {
			host := util.GetHost(uri)
			if !util.AuthIsFQDNOrIP(host) {
				return &LintResult{Status: Error}
			}
		}
```
Given the input mentioned before, util.GetHost(uri) returns an empty string. 
The problem is compounded by the bad test data in the `testlint/testCerts/SANURIHostWildcardFQDN.pem` certificate. 

```
X509v3 Subject Alternative Name: 
                DNS:*.example.com, URI:test//user@*.example.com:1337/test
```
The uri in this cert doesn't have colon after the scheme, this is not a valid URI in the first place. 
I suspect that the test data have been generated first and then the util.GetHost() have been created to work with the test input. 
Given a valid URI like `sip:user@example.domain.com` or `test://user@*.example.com:1337/test` the linter fails.

zakird:
I believe this was addressed in #199. Closing.

PR #199, opened 2018-01-12
Address issue 198 (SAN format)

justinbastress:
Two issues were identified in https://github.com/zmap/zlint/issues/198 -- 

1. util.GetHost doesn't act as expected (returns empty string for valid URIs)
2. Two test certs had mistakenly-invalid URIs (`test//` instead of `test://`)

The problem with util.GetHost seems to be that it expected an authority (not a full URI, as was being passed to it by the lint), and further it required that authority to have a userinfo piece -- see https://tools.ietf.org/html/rfc3986#section-3.2). The lint now uses the built-in `net/url.Parse().Host` to get the host name, and `util.IsFQDNOrIP(parsed.Host)` to check its validity.

The test certs were replaced with certificates having the same SAN, but with `test://` instead of `test//`.

### DIFF
--- a/util/fqdn.go
+++ b/util/fqdn.go
@@ -16,6 +16,7 @@ package util
 
 import (
 	"net"
+	"net/url"
 	"strings"
 
 	"github.com/asaskevich/govalidator"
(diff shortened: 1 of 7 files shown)


dadrian:
Are we able to just use `net.url.Parse` instead of home-rolling URI parsing? https://golang.org/pkg/net/url/#URL

justinbastress:
@dadrian one thing that the home-rolled code does that net.url.Parse does not do is getting the full authority string, rather than breaking it into user + host. But in this case, as I read it, it really is the isolated host (not the authority) that we need, so net.url.Parse is appropriate.

And, as the reporter mentioned, nothing else seems to be using the util.GetHost / util.GetAuthority.

dadrian:
Can we implement our utility functions in terms of `net`?

justinbastress:
@dadrian the two that overlap -- GetAuthority and GetHost -- probably cannot be rewritten in terms of net.url primitives, because the latter do not return the authority string, which is the output of GetAuthority and the input for GetHost.

You could try to work backwards to get it with something like `return parsed.User ? parsed.User + "@" + parsed.Host : parsed.Host`, but that would be pretty brittle.

The rest of fqdn.go is based on net/govalidator/publicsuffix (with some helper functions like `RemovePrependedWildcard` to get the input into a format that will work with the standard libraries).

dadrian:
@justinbastress That's unfortunate. We should probably write a couple basic tests for `GetHost` and `GetAuthority` then.

util/fqdn.go
```diff
@@ -62,7 +62,7 @@ func GetAuthority(uri string) string {
 func GetHost(auth string) string {
 	begin := strings.Index(auth, "@")
 	if begin == -1 || begin == len(auth)-1 {
-		return ""
+		begin = -1
```
zakird: I'm not sure if I understand the logic of setting `begin := -1` when `begin == -1`.

tadukurow:
Unfortunately this change will not work for any URI that doesn't have a scheme followed by "//'.

url.Parse treats any rootless path as opaque and stops parsing after that. 
https://github.com/golang/go/blob/master/src/net/url/url.go#L513

This means that the "Host" portion is not set, which will cause this lint to fail. 
And example of a a URI without "/" would be any SIP URI as defined in:
https://tools.ietf.org/html/rfc5922

example: https://play.golang.org/p/cRnYBmRZpJ7

justinbastress:
@tadukurow thank you for the examples. If I'm reading the URI RFC correctly (https://tools.ietf.org/html/rfc3986#section-3), the sip-style URIs do not have an authority -- the `hier-part` is `path-rootless`.

So, in this case, the SIP URI should be vacuously accepted -- "*URIs that include an authority ([RFC3986], Section 3.2) MUST include a fully qualified domain name or IP address as the host.*"

But, that would still mean that the current PR is wrong -- it should first check if `parsed.Opaque != ''`, and if so, succeed (since non-empty `Opaque` means that it was a `path-rootless`, which means no `authority`). From that point on, everything else should be fine -- if it's not `Opaque`, then an empty `Host` is an error condition.

justinbastress:
@zakird @dadrian re-implemented the `GetHost` / `GetAuthority`, leveraging `net.url.Parse` as much as possible, and then following https://tools.ietf.org/html/rfc3986#section-3.

Also added a set of unit tests capturing all of the basic happy cases (32 of them), and the two basic exceptional cases (a bad URI, like test//does/not/have/a/colon, which fails, and a valid URI with no authority, like @tadukurow 's case, which succeeds).

Issue #201, opened 2018-02-09
e_sub_ca_must_not_contain_any_policy should be removed


robstradling:
```Subordinate CA: MUST NOT contain the anyPolicy identifier (2.5.29.32.0)```

This is only true when the Subordinate CA "is not an Affiliate of the Issuing CA".  I think evaluating legal relationships is out of scope for ZLint!

Here are two examples of intermediate certs where the Issuing CA and the Subordinate CA are affiliated, and so this error is a false positive:
https://crt.sh/?id=250864699&opt=zlint
https://crt.sh/?id=5779800&opt=zlint

I propose that this lint should be removed.

zakird:
👍

justinbastress:
Yeah, "Affiliate" is defined in such a way that there's no way we could verify affiliation:

https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.5.4.pdf, section 1.6.1:

> **Affiliate**: A corporation, partnership, joint venture or other entity controlling, controlled by, or under common control with another entity, or an agency, department, political subdivision, or any entity operating under the direct control of a Government Entity. 

And from section 7.1.6.3:

> A Certificate issued after the Effective Date to a Subordinate CA that is an affiliate of the Issuing CA: 
> 	1. ...
> 	2. MAY contain the “anyPolicy” identifier (2.5.29.32.0) in place of an explicit policy identifier.

zakird:
I've merged #202. Closing.

robstradling:
Thanks Zakir.

PR #202, opened 2018-02-09
Remove anyPolicy check on subordinate CAs

justinbastress:
Per issue #201, subordinate CAs are only forbidden to have the AnyPolicy EKU when they are not affiliated with the CA -- but we have no way of checking affiliation.


zakird:
LGTM. @justinbastress we should also mark this field as deprecated in the Censys protobuf.

Issue #211, opened 2018-03-13
TestSANEDIPartyMissing failing


jsha:
Running unittests on master with go version go1.10 linux/amd64:

```
asn1: structure error: explicitly tagged member didn't match
--- FAIL: TestSANEDIPartyMissing (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x530 pc=0x6a2855]

goroutine 262 [running]:
testing.tRunner.func1(0xc4202e93b0)
        /home/jsha/go1.10/src/testing/testing.go:742 +0x29d
panic(0x76d1e0, 0xa73cb0)
        /home/jsha/go1.10/src/runtime/panic.go:505 +0x229
github.com/zmap/zlint/util.IsServerAuthCert(0x0, 0x785ac0)
        /home/jsha/gopkg/src/github.com/zmap/zlint/util/ca.go:48 +0x5
github.com/zmap/zlint/lints.(*Lint).Execute(0xc4202cf500, 0x0, 0x8060a3)
        /home/jsha/gopkg/src/github.com/zmap/zlint/lints/base.go:103 +0x117
github.com/zmap/zlint/lints.TestSANEDIPartyMissing(0xc4202e93b0)
        /home/jsha/gopkg/src/github.com/zmap/zlint/lints/lint_ext_san_edi_party_name_present_test.go:33 +0x9e
testing.tRunner(0xc4202e93b0, 0x81e750)
        /home/jsha/go1.10/src/testing/testing.go:777 +0xd0
created by testing.(*T).Run
        /home/jsha/go1.10/src/testing/testing.go:824 +0x2e0
FAIL    github.com/zmap/zlint/lints     0.060s
```

justinbastress:
It looks like SANOtherName.pem is actually malformed -- neither OpenSSL nor Windows can decipher the SAN.

Poking around at the SAN (https://lapo.it/asn1js/#3016A00B06025303130568656C6C6F8207646E732E636F6D)
```
30 16 
     A0 0B
          06 02 53 03
          13 05 68 65 6C 6C 6F 
     82 07 
          64 6E 73 2E 63 6F 6D
```
[...]

PR #214, opened 2018-03-14
New SANOtherName test cert with CONSTRUCTED OtherName value

justinbastress:
Re issue #211. After updating, `go test` passes with on `docker run --rm -it golang:1.10`.

The new certificate is  quite different from the original (which was ~pathological~ challenging in a number of other ways), but it does have an OtherName SAN attribute, which seems to be the relevant attribute.

### DIFF
--- a/testlint/testCerts/SANOtherName.pem
+++ b/testlint/testCerts/SANOtherName.pem
@@ -1,118 +1,82 @@
+Generated with:
+#!/bin/bash -e
+echo "subjectAltName=DNS:dns.com,otherName:2.3.3;UTF8:hello" > SANOtherName.ext
+openssl req -new -subj '/CN=OtherName' -days 3650 -newkey rsa:2048 -sha256 -nodes -keyout SANOtherName.key -out SANOtherName.csr
+if ! [ -f "root.key" ]; then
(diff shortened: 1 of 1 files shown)


zakird:
Makes sense to fix this. That said, it also seems that we shouldn't panic on a bad certificate. How do handle if something like that gets passed to zlint in general?

justinbastress:
To be fair, this panic is in the *test* code, not the lint itself (testingUtil.ReadCertificate 
 -- https://github.com/zmap/zlint/blob/master/lints/testingUtil.go#L48 -- panics if fails to *read* the certificate, it `panic`s, while if it fails to *parse* the certificate, it returns `nil`; but in a test scenario, failing to read or parse a test certificate seems like an unrecoverable error to me).

Though having said that, if someone passes in a certificate that fails to parse, is that really so different? It's a bit more ambiguous in cases like this, where it fails to parse an extension (which at the certificate level is an opaque `OCTET STRING`), or even worse, in that issue where Go started throwing because of illegal characters in a `PrintableString`...but unless we bring our own ASN.1 parser (or lock ourselves to a more lenient version of go's), I don't see how we get around that.

zakird:
Makes sense to me. merging

Issue #220, opened 2018-05-07
Incorrect warning on multiple attributes in one RDN


YuryStrozhevsky:
In your code you have [this warning](https://github.com/zmap/zlint/blob/master/lints/lint_subject_multiple_rdn.go#L42) on having "multiple attributes in one RDN" in one certificate. In fact I do not understand the warning and why you made it.

As from initial type definition we have this:
```asn.1
RDNSequence ::= SEQUENCE OF RelativeDistinguishedName
		
RelativeDistinguishedName ::=
SET SIZE (1..MAX) OF AttributeTypeAndValue
```
And nothing stops us from having multiple `AttributeTypeAndValue` - could be `(1..MAX)` values inside one `RelativeDistinguishedName`.

So, could you describe why you made the warning in `zlint`?

justinbastress:
@YuryStrozhevsky from the Source (https://github.com/zmap/zlint/blob/master/lints/lint_subject_multiple_rdn.go#L52), it appear this was pulled from the AWSLabs certlint -- and there doesn't seem to be any I don't see any further reference to a specification there (https://github.com/awslabs/certlint/blob/25d5957f8c36dafcbd82870df57a8367b49650be/lib/certlint/namelint.rb#L121).

rmhrisk:
@pzb can you elaborate?

pzb:
Certlint warns about it because I observed that the majority of multiple attribute RDNs are due to bugs.  A surprising number of certificates includes all the attributes in a single RDN by mistake.  I'm assuming this was because of a poor API to create Names where an array of attributes becomes a single RDN rather than becoming a sequence of RDNs with one attribute each.

In certlint Warnings (`W`) are advisory only.  It will warn on things that seem "odd" even if they can be correct; another example is leading/trailing whitespace in attribute values.  That can be completely valid, but also can be an indication of a bug

zakird:
It might be worth downgrading this from a warning to a notice.

dadrian:
Are client API's equipped to deal with RDN's of size > 1? Offhand, I know that at least Golang pushes the extra attributes to a grab-bag `[]Names` for certain values, including CommonName and SerialNumber. If many clients don't expose these values, it may be worth leaving the notice as a warning.

https://golang.org/src/crypto/x509/pkix/pkix.go?s=3333:3386#L123

pzb:
Ruby also has issues:

[...]

PR #221, opened 2018-05-20
subject_multiple_rdn.go warning -> notice

zakird:
This addresses #220.

### DIFF
--- a/lints/lint_subject_multiple_rdn_test.go
+++ b/lints/lint_subject_multiple_rdn_test.go
@@ -30,7 +30,7 @@ func TestSubjectRDNTwoAttribute(t *testing.T) {
 func TestSubjectRDNOneAttribute(t *testing.T) {
 	inputPath := "../testlint/testCerts/RSASHA1Good.pem"
 	expected := Pass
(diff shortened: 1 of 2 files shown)


lints/lint_subject_multiple_rdn_test.go
```diff
@@ -21,7 +21,7 @@ import (
 func TestSubjectRDNTwoAttribute(t *testing.T) {
 	inputPath := "../testlint/testCerts/subjectRDNTwoAttribute.pem"
 	expected := Warn
```
titanous: Shouldn't this be `Notice` instead of `Warn`? (with the same change to the implementation)

zakird: 🤦‍♂️

zakird:
@titanous this should be fixed now. Thanks for catching that. I also updated the lint description.

kumarde:
@zakird I'm not sure if Censys protobufs changed since I last looked at it, but this would likely cause some changes on that end as well. I think it should be as easy as updating the protobuf and running `go generate`, but I would double check that too.

PR #222, opened 2018-05-20
email length is 255 not 128

zakird:


titanous:
It looks like the tests need an update, but this change looks fine.

zakird:
Hah, I was about to say, I was at least going to fix the tests before asking for a review ;)

PR #233, opened 2018-08-20
Update TLD map using updateTLDs.sh

ezekiel:
Unfortunately, `gofmt` wants there to be a large amount of whitespace changes.


zakird:
I'm not sure that we want to remove old TLDs given that will cause historical sites to be marked now invalid despite being valid originally. I think we probably just want to append to the existing list.

cpu:
@zakird I think there's a trade-off to be weighed here. Leaving the old gTLDs in place will mean the linter doesn't flag new issuance for retired gTLDs and I think that could be equally problematic. Without capturing the date of the ICANN gTLD revocation alongside the removal in the map such that the check can be contextualized based on the certificate's issuance date I think there will need to be a decision about whether zlint prefers to avoid false-positives for historic certificates or whether zlint should prefer to avoid false-negatives for new certificates issued for revoked gTLDs. For my own purposes the latter is preferred! What do you think?

zakird:
I agree with you that neither solution here is a good one. Because this has been used in Censys to effectively "grade" CAs, I'm still a hesitant to put in anything that would cause a false positive. It's easier for folks to ignore the numbers when they can point to false positives. That, for example, is why all of the Lints have effective periods. At the same time, you're absolutely correct that we want CAs following today's set.

I think the correct solution here is to similarly give TLDs effective periods as well, e.g., each TLD has a begin and end date associated with it. Thoughts?

cpu:
> I agree with you that neither solution here is a good one. Because this has been used in Censys to effectively "grade" CAs, I'm still a hesitant to put in anything that would cause a false positive.

Makes sense :+1: 

> I think the correct solution here is to similarly give TLDs effective periods as well, e.g., each TLD has a begin and end date associated with it. Thoughts?

That sounds reasonable to me but I think it might be outside of the scope of what @ezekiel signed up for with his PR before I hijacked it for this discussion :laughing: Does your team have the development time to implement this? If the answer is no that's totally fair and I'm prepared to scratch my own itch with a PR but it will take me some time.

With respect to moving this PR forward: Given that zlint was born of zmap/Censys. I don't think its reasonable to pressure you into making a decision that would result in false positives for that use-case in favour of my own. Amending the PR to leave the old gTLD's in place and only add the missing entries sounds like the right short-term decision :+1:  - On my side I'll know that I can't rely on zlint to catch mississuance for a revoked gTLD until the follow-up issue of storing effective gTLD periods is addressed.

titanous:
This CSV has a list of removed TLDs (though it may need updating): https://github.com/pzb/TLDs/blob/master/removed/rmtlds.csv

This CSV has a list of gTLDs with their start dates (again, might need an update): https://github.com/awslabs/certlint/blob/master/data/newgtlds.csv

ezekiel:
Pushed a new change which is additions-only (with a corresponding change to `updateTLDs.sh` which kinda helps with this operation in the future).

No more whitespace change noise, but there is a tiny bit of re-sorting noise.

zakird:
I suspect the team won't have the opportunity to add this in in the near term (they're pretty underwater with a migration). If you have cycles, a PR would be tremendously appreciated. I'm going to create an issue to track this conversation after I merge this.

cpu:
>  If you have cycles, a PR would be tremendously appreciated. I'm going to create an issue to track this conversation after I merge this.

Sounds great. I've made a note to try and slot a PR for this into my sprint next week.

cpu:
>  If you have cycles, a PR would be tremendously appreciated. I'm going to create an issue to track this conversation after I merge this.

@zakird I think :point_up: might have been forgotten in the merge shuffle. I created 
 https://github.com/zmap/zlint/issues/235 with some rough notes and would appreciate a :mag: before I get down to writing code if you have a chance. Thanks!

Issue #235, opened 2018-09-04
Give TLD data and associated lints effective periods


cpu:
## Background

This is a follow-up issue from the discussion in https://github.com/zmap/zlint/pull/233 - to summarize the problem:
IANA occasionally deprecates a gTLD. The existing [`updateTLDs.sh` script](https://github.com/zmap/zlint/blob/02fe9a29bbae57da0c77db7afb53734dc262b130/updateTLDs.sh) removes deprecated gTLDs from the [generated `tldMap` data](https://github.com/zmap/zlint/blob/848521ffb4ee6042d34faf7389e970d149a84d91/util/gtld_map.go#L17). This results in false-positive linting failures for previously issued certificates with subject domains using the previously valid gTLD. 

To properly reflect gTLD's having effective periods the `tldMap` and its associated generation process shouldn't remove gTLDs but instead set a "notafter" date in the data matching the IANA deprecation date. Similarly when a new gTLD is added the `tldMap` should be updated with an entry for the new gTLD that sets the "notbefore" date in the mapping to the time IANA added the gTLD entry. The associated linters should be updated to check that the certificate's issuance date is coherent with the associated subject identifier gTLD effective period. This catches invalid issuance for retired/not-yet-introduced gTLDs while not introducing false-positives for valid certificates when a gTLD is deprecated.

## Proposed Implementation

I'm happy to submit a PR for this issue. Here's a rough outline of how I'd approach this:

* add a `GTLDPeriod` type to `util/` for capturing a start/end `time.Date`
* change the `util/gtld_map.go` `tldMap` from `map[string]bool` to `map[string]GTLDPeriod`
* update the linters to use the `GTLDPeriod` information
* Replace `updateTLDs.sh` and `scripts/consolodate_tlds.py` with a Go cmd (`zlint-gtld-update`?) that writes a `map[string]GTLDPeriod`
* Update `util/gtld_map.go` to use `go generate zlint-gtld-update`

@zakird Does the above sound reasonable to you? If you'd prefer to keep the utility for this split across BASH/Python I can make that work but I lightly prefer using Go throughout. If the above looks good I can get started on a PR this week.

zakird:
This sounds spot-on to me.

cpu:
@zakird Great, thanks for confirming. I'll get started and we can hash out the fine details in the PR review.

cpu:
Here's my initial PR: https://github.com/zmap/zlint/pull/236

PR #236, opened 2018-09-06
Give TLD data effective periods, rework generation.

cpu:
A new `zlint-gtld-update` command is added which fetches ICANN/IANA TLD data, generating a Go map of TLD names to their effective periods. The `data/gtld_map.go` file is now generated using `zlint-gtld-update` using `go generate ./...` in the root of the project. This replaces `updateTLDs.sh`, `scripts/consolidate_tlds.py`, and the `data/` directory files.

To build the map, two authoritative data sources are consulted:
1) The ICANN [gTLD JSON registry v2](https://www.icann.org/resources/registries/gtlds/v2/gtlds.json)
2) The ICANN [TLD alpha-by-domain list](http://data.iana.org/TLD/tlds-alpha-by-domain.txt)

Only the first data source provides the delegation and removal dates for TLDs, and is missing ccTLDs like `.uk`, `.ca`, etc. To build a complete `tldMap` the first source is used to build the initial map and any missing TLDs from the alpha-by-domain list are added with an unbounded effective period starting at the date the `.com` gTLD was delegated. This seems like the best compromise since I can't find an equivalent data source for ccTLDs as gTLDs.

To match the [special casing of `.onion`](https://github.com/zmap/zlint/blob/848521ffb4ee6042d34faf7389e970d149a84d91/updateTLDs.sh#L3) done by the replaced `updateTLDs.sh` script the built `tldMap` includes a static entry for `.onion` with a `delegationDate` of the CABF ballot that allowed `.onion` issuance for EV certificates.

The `lint_dnsname_right_label_valid_tld.go` lint is updated to check that a TLD exists and was valid at the time the certificate was issued. Tests are added for the case where a certificate identifier referenced
a TLD that was not yet valid, that was no longer valid, and that was valid at the time of issuance.

Resolves https://github.com/zmap/zlint/issues/235

**Notes to reviewers**: 
<s>1) I'll be out of office for a few days. I'll have to pick up review feedback when I'm back next week. Feel free to delay reviewing until ~Tues, Sept 11th if you prefer to wait for me!</s>
2) I haven't written any unit tests for the functions in the `zlint-gtld-update` command. Happy to do that if you folks would like, I wanted to put this out for some review before I spend too much time testing code that will change based on review feedback.
3) I did a quick diff between the previous TLD list and the new one and didn't spot any omissions/errors but it would be good to give this some thorough attention.

Thanks!


zakird:
This looks great to me. @justinbastress would you be willing to take a pass when you have a moment?

updateTLDs.sh
```diff
@@ -1,9 +0,0 @@
-# Script to update the list of gTLDs
-curl -o data/iana_gtlds.txt http://data.iana.org/TLD/tlds-alpha-by-domain.txt
-echo "ONION" >> data/iana_gtlds.txt
```
cpu: Oops! I forgot to account for this one manual TLD addition.

Probably this should just be special cased to have a `delegationDate` == when the CABF blessed issuance for the EV certs for `.onion` domains? Open to suggestions on how to handle this case. I think it could also benefit from a unit test so that it isn't forgotten in the future.

cpu:
Thanks @zakird!

PR #243, opened 2018-11-08
e_subject_common_name_not_from_san incorrectly fails on case sensitive match

cardonator:
This PR fixes a bug in the `e_subject_common_name_not_from_san` linter which causes an error to be returned when the CN field contains a name that matches a SAN with inconsistent case.

For reference,

Pre-Pull Request
CN: gov.us, SANS: [GOV.us] -- FAIL
CN: gov.us, SANS: [gov.us] -- PASS

Post-Pull Request
CN: gov.us, SANS: [GOV.us] -- PASS
CN: gov.us, SANS: [gov.us] -- PASS

### DIFF
--- a/lints/lint_subject_common_name_not_from_san.go
+++ b/lints/lint_subject_common_name_not_from_san.go
@@ -22,6 +22,8 @@ contained in the Certificate’s subjectAltName extension (see Section 7.1.4.2.1
 ************************************************/
 
 import (
+	"strings"
+
 	"github.com/zmap/zcrypto/x509"
 	"github.com/zmap/zlint/util"
 )
@@ -40,7 +42,7 @@ func (l *subjectCommonNameNotFromSAN) Execute(c *x509.Certificate) *LintResult {
 	cn := c.Subject.CommonName
 
 	for _, dn := range c.DNSNames {
-		if cn == dn {
+		if strings.ToLower(cn) == strings.ToLower(dn) {
 			return &LintResult{Status: Pass}
 		}
 	}
--- a/lints/lint_subject_common_name_not_from_san_test.go
+++ b/lints/lint_subject_common_name_not_from_san_test.go
@@ -35,3 +35,12 @@ func TestCnFromSAN(t *testing.T) {
 		t.Errorf("%s: expected %s, got %s", inputPath, expected, out.Status)
 	}
 }
+
+func TestCnCaseNotMatchingSAN(t *testing.T) {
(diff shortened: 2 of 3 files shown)


lints/lint_subject_common_name_not_from_san.go
```diff
@@ -40,7 +42,7 @@ func (l *subjectCommonNameNotFromSAN) Execute(c *x509.Certificate) *LintResult {
 	cn := c.Subject.CommonName
 
 	for _, dn := range c.DNSNames {
-		if cn == dn {
+		if strings.ToLower(cn) == strings.ToLower(dn) {
```
justinbastress: Does `ToLower` comparison work for all legal Fully-Qualfied Domain Names? It sounds like [EqualFold](https://golang.org/pkg/strings/#EqualFold) may be the appropriate comparator here -- 

> EqualFold reports whether s and t, interpreted as UTF-8 strings, are equal under Unicode case-folding.

```suggestion
		if strings.EqualFold(cn, dn) {
```

cardonator: tl;dr: there should only be a limited set of ASCII characters in these fields on CA/B compliant certs, therefore `strings.ToLower` should be fine.

Legal values in the `subjectAlternativeName` field are defined in the following CA/B Forum rule: https://github.com/cabforum/documents/blob/master/docs/BR.md#71421-subject-alternative-name-extension which reads:
```Each entry MUST be either a dNSName containing the Fully-Qualified Domain Name or an iPAddress containing the IP address of a server. ```
In RFC5280, `dNSName` is defined as: (https://tools.ietf.org/html/rfc5280#section-4.2.1.6)
```dNSName                         [2]     IA5String,```

And IA5String is defined as: https://en.wikipedia.org/wiki/IA5STRING
```generally equivalent to the first 128 characters of the ASCII alphabet```

According to https://tools.ietf.org/html/rfc5280#section-7.2, any international names must be converted to the `ASCII Compatible Encoding` format which is basically `xn--<punycode>` (https://en.wikipedia.org/wiki/Punycode).

Therefore, it stands to reason that the CN and subjectAlternativeNames fields should only have a limited set of ASCII characters available to them which should work with `strings.ToLower`.

cardonator: Looking at the two approaches in terms of performance, it seems like it makes sense to switch to the other function anyway. Thanks, @justinbastress

PR #244, opened 2018-11-09
Properly parse BMPStrings out of the ExplicitText userNotice field for length check

### LINKED ISSUE #242: Incorrect maximum length constraint checking for explicitText (bytes vs. characters)
Zlint returns an error for the [attached certificate](https://github.com/zmap/zlint/files/2558171/www.mokk.hu.zip), stating that "**Explicit text has a maximum size of 200 characters**". 

However, the suspicious explicitText **has only 170 characters** ("Nem minősített, szervezet-ellenőrzött webszerver tanúsítvány, 3. hitelesítési osztály, regisztrációkor a személyes megjelenés kötelező. A tanúsítvány igénylője szervezet."), which, encoded in BMPString (UTF-16BE) is **340 bytes long** (there are 2 bytes per character).

https://github.com/zmap/zlint/blob/master/lints/lint_ext_cert_policy_explicit_text_too_long.go reads:

```go
if len(text.Bytes) > 200 {
}
```
so it seemingly checks lengths in **bytes**, not **characters**, which is incorrect.

According to [RFC 6818](https://tools.ietf.org/html/rfc6818): "The explicitText field is a string with maximum size of 200 characters.", which means we have to count characters, not bytes.

cardonator:
I am creating this PR to get some feedback while I work on getting a valid self-signed cert to submit for the test (I tested against the cert in #242 ).

The main question I have in this approach is where should parseBMPString actually live? It feels wrong for this function to be in the linter itself.

This will fix #242


zakird:
I think this should probably live in https://github.com/zmap/zlint/tree/master/util.

cardonator:
Thanks, @zakird ! Will make the necessary change.

cardonator:
@zakird I have made the requested changes and added a test PEM. Please let me know if I should make any additional changes.

cardonator:
@justinbastress @zakird any other changes needed for this PR?

util/encodings.go
```diff
@@ -115,3 +117,20 @@ var emptyASN1Sequence = []byte{0x30, 0x00}
 func IsEmptyASN1Sequence(input []byte) bool {
 	return len(input) < 2 || bytes.Equal(input, emptyASN1Sequence)
 }
+
+// ParseBMPString returns a uint16 encoded string following the specification for a BMPString type
+func ParseBMPString(bmpString []byte) (string, error) {
+	if len(bmpString)%2 != 0 {
+		return "", errors.New("pkcs12: odd-length BMP string")
```
justinbastress: pkcs12?

util/encodings.go
```diff
@@ -115,3 +117,20 @@ var emptyASN1Sequence = []byte{0x30, 0x00}
...
+	}
+	// strip terminator if present
+	if l := len(bmpString); l >= 2 && bmpString[l-1] == 0 && bmpString[l-2] == 0 {
+		bmpString = bmpString[:l-2]
+	}
+	s := make([]uint16, 0, len(bmpString)/2)
+	for len(bmpString) > 0 {
+		s = append(s, uint16(bmpString[0])<<8+uint16(bmpString[1]))
+		bmpString = bmpString[2:]
```
justinbastress: unused assignment?

lints/lint_ext_cert_policy_explicit_text_too_long.go
```diff
@@ -50,7 +50,13 @@ func (l *explicitTextTooLong) CheckApplies(c *x509.Certificate) bool {
 func (l *explicitTextTooLong) Execute(c *x509.Certificate) *LintResult {
 	for _, firstLvl := range c.ExplicitTexts {
 		for _, text := range firstLvl {
-			if len(text.Bytes) > 200 {
+			var runes string
+			if text.FullBytes[0] == 30 {
```
justinbastress: A brief comment here would be helpful, laying out the intent and any assumptions -- I can sort of infer that `text.FullBytes` is the encoded (assumed to be / guaranteed to be DER?) BMPString, with tag 0x1e == 30, but it's definitely not clear from the context alone. 

Also, it should probably use constant for the tag rather than the magic number 30; but even then, doing a straight `tag == BMP_STRING_0x1E` comparison may fail for weird corner cases like CONSTRUCTED strings (though off-hand, I'm not sure that's a valid scenario).

util/encodings.go
cardonator: Removed, sorry. That was because I pulled some functionality from a pending golang PR.

lints/lint_ext_cert_policy_explicit_text_too_long.go
cardonator: I added a short comment here indicating what the Bytes we are inspecting are. I also switched to using a constant defined at the top of the lint just so it would be clear what Tag we are looking at and compare to the Tag instead of the FullBytes.

PR #247, opened 2018-11-23
Add lints include/exclude flag to executable

tadukurow:
Added option to include/exclude lints from the command line. 

Found it useful for some quick analysis on a large number of certs.


cmd/zlint/main.go
```diff
@@ -144,3 +153,59 @@ func lint(inputFile *os.File, inform string) {
...
+}
+
+func includeLints() {
+	if include == "" {
+		return
+	}
+
+	// parse includes to map for easier matching
+	var includes = strings.Split(include, ",")
+	var includesMap = make(map[string]interface{}, len(includes))
```
szank: Nitpick: shouldn't it be `map[string]struct{}`

zakird:
This looks fine overall to me, though I think I might it be better off living in https://github.com/zmap/zcertificate, which handles a number of additional pieces of customizability versus the included (simple) zlint CLI.

tadukurow:
Hi @zakird, we've had the exclusion use case again, is there anything else to do on this to get it merged?

PR #249, opened 2019-01-08
New lint to ensure DN fields only contain printable characters

cardonator:
x509lint contains a check to make sure that all DN fields in a cert contain only printable characters. This originates from RFC5280 Appendix A where all subject fields are defined with printable string character types starting from 1 character. This implicitly means that there should only be printable characters in all DN fields.

This check is based on the x509lint logic contained in https://github.com/kroeckx/x509lint/blob/b7359233574b61cb3976d96f6525791c5d56d340/checks.c#L238 however I left out the null byte check because I believe that other checks in zlint already catch this.


lints/lint_subject_dn_not_printable_characters.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	for _, attrTypeAndValueSet := range rdnSequence {
+		for _, attrTypeAndValue := range attrTypeAndValueSet {
+			for _, byte := range attrTypeAndValue.Value.Bytes {
+				if byte < 32 {
+					log.Println(attrTypeAndValue.Value.Tag, "bad character:", byte)
+					return &LintResult{Status: Error}
+				}
+				if byte >= 0x7F && byte <= 0x9F {
```
kumarde: Can we use either decimal or hex but not both? Just for readability.

lints/lint_subject_dn_not_printable_characters.go
```diff
@@ -0,0 +1,72 @@
...
+		return &LintResult{Status: Fatal}
+	}
+	if len(rest) > 0 {
+		return &LintResult{Status: Fatal}
+	}
+
+	for _, attrTypeAndValueSet := range rdnSequence {
+		for _, attrTypeAndValue := range attrTypeAndValueSet {
+			for _, byte := range attrTypeAndValue.Value.Bytes {
+				if byte < 32 {
```
kumarde: It's probably clearer if we check if each byte is in printable range, rather than the inverse.

lints/lint_subject_dn_not_printable_characters.go
cardonator: Sure. I have made that change.

lints/lint_subject_dn_not_printable_characters.go
cardonator: I believe the current functionality makes the most sense as opposed to something like:
```
if byte in the right range {
    continue
}

```
right?

kumarde: Yeah, you're right. Original way makes more sense.

cardonator:
Sorry for all the weird merge trash in this branch. My fork got messed up somehow.
I believe this is ready to go now, @zakird any more requested changes?

cwjnkins:
I would like to know more about where this restriction comes from. The PR description cites RFC5280 appendix A, but (for example) in that appendix the declaration for X520CommonName is

    -- Naming attributes of type X520CommonName:
    --   X520CommonName ::= DirectoryName (SIZE (1..ub-common-name))
    --
    -- Expanded to avoid parameterized type:
    X520CommonName ::= CHOICE {
      teletexString     TeletexString   (SIZE (1..ub-common-name)),
      printableString   PrintableString (SIZE (1..ub-common-name)),
      universalString   UniversalString (SIZE (1..ub-common-name)),
      utf8String        UTF8String      (SIZE (1..ub-common-name)),
      bmpString         BMPString       (SIZE (1..ub-common-name)) }

Furthermore, the start of Section [7.1](https://datatracker.ietf.org/doc/html/rfc5280#section-7.1) of RFC5820 says

> Representation of internationalized names in distinguished names is
   covered in Sections [4.1.2.4](https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.4), Issuer Name, and 4.1.2.6, Subject Name.
   Standard naming attributes, such as common name, employ the
   DirectoryString type, which supports internationalized names through
   a variety of language encodings.  Conforming implementations MUST
   support UTF8String and PrintableString.
   
which to my understanding suggests the attribute values for common names can fall outside the PrintableString character set for international names.

cwjnkins:
For some context, [this](https://lapo.it/asn1js/#MIIDADCCAeigAwIBAgIQQA-Yeb449LFEMzjLmB3UxDANBgkqhkiG9w0BAQsFADA7MTkwNwYDVQQDHjAAUwBTAEwAXwBTAGUAbABmAF8AUwBpAGcAbgBlAGQAXwBGAGEAbABsAGIAYQBjAGswIBcNMjExMjA5MTA1MDIzWhgPMjA1MTEyMDkxMDUwMjNaMDsxOTA3BgNVBAMeMABTAFMATABfAFMAZQBsAGYAXwBTAGkAZwBuAGUAZABfAEYAYQBsAGwAYgBhAGMAazCCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAK-wYR4S1AuJKWA0wTyWsC7sdD2tTavSsmpr8ALD3HRxccCzbqEhvjOo3losuj_ttQ4Gb5TbZVJQAyqZXXdYDRKapRppE12c2EYVDcocKUhzPlvVRR5ykO9VPDiK3Z-FA_aPhOCBOlgIqCPrQrx2sd0ANBdhxa0Pd_fKELgJu0Qh8qa63KEVOAD_QnSk66naUHsrQeGG5zPGODscLhz-NEjExJ5u1lxcRbZzyU68-l_-fNlj7KNSCDCSGSm2DbEKrZZIVYSs9SeGn0e5Y6xtd-5FJShaOJpXqZoOOYvtybpZe7tr4IHi0imv93CLqf33FR3PfzHii0nFLPxbULIfhScCAwEAATANBgkqhkiG9w0BAQsFAAOCAQEAHcP5T1UxZZ6DdznA-E6abi235nXgFRqBauS7lIJWTwnHqsN9ZRBcFTP_Ar7lirEQ6FGuJ2SOcGlUiccs41GrM_wCZxwuxRMZuYY5SQp0fj2wsUyUnqsZY5LLYbqwncQwsc7pqH4uFmTJC2CX_rBWEuCKQH_iWyy8SgPKBT05TxxfMMJ8cdNHNYQ2VkbRIPfq7TgWJIAgPQmfHzPqZDoieHThgmqIkgSl8m7GpgbF2NN_jP4uiiXyKgOBhEGTekS9ovfZyhM_AoWBDBIJDSCyBYeTDhOG5txTYbubsr9f3Tzzjb2M_b0XG5Ng7FJCmEGq036umwE1FHBzeX6fENdBWg) is an example certificate that ZLint flags as failing the `e_subject_dn_not_printable_characters` lint. The common name decodes to "SSL_Self_Signed_Fallback", but since it's encoded as a BMPString (a restriction of UTF16), each character has a leading 0x00 byte which is what appears to be triggering the error. Looking at the lint in question, it seems to assume the strings are always UTF8 encoded.

zakird:
Can you please open a new issue to discuss this? This PR is more than 5 years old and it's unlikely to get attention from the community.

Issue #259, opened 2019-02-26
Error lint for `arpa` domains corresponding to reserved IP addresses


dadrian:
Per https://groups.google.com/forum/#!topic/mozilla.dev.security.policy/JFwqZx7RLL0

While `arpa` names are legal, there is not way to properly validate a certificate for reserved IP space (e.g. 10.0.0.0/8, 192.168.0.0/16). Certificates with domains such as `1.1.168.192.in-addr.arpa` should trigger an error.

This is a more specific case than #177.

zakird:
This lint should use https://github.com/zmap/zlint/blob/master/util/ip.go#L62 similar to the checks we have for IP addresses.

cpu:
:wave: I started working on this and should have a PR available for review in a little bit.

I'm going to assign this issue to myself and remove the help wanted label. If someone else _really_ wants to take this off my plate feel free to speak up and we can pair or I can give it up :-)

cpu:
Here's a first pass: #260. It took me a little bit longer because I forgot how bizarre rfc3596 & reverse DNS for IPv6 is :laughing: I think the lint is ~=ready. If someone wanted to do the rules lawyering to find the best BR citation for why this isn't permitted I'd appreciate it.

PR #260, opened 2019-02-27
lints: add new lints for .arpa reverse DNS subjects.

cpu:
Two new lints are added:

First `lint_subject_contains_malformed_arpa_ip.go` **warns** if there are any DNSNames in a certificate under the `.in-addr.arpa` or `.ip6.arpa` zones that do not have the correct number of labels, or have the correct number of labels but do not parse as a reversed IP address.

Second, `lint_subject_contains_reserved_arpa_ip.go` checks that all DNSNames in a certificate that specify a well-formed rDNS name (e.g. correct number of labels, parses as a reversed IP) under the `.in-addr.arpa` or `.ip6.arpa` zones do not specify an addresses in an IANA reserved network range. This condition is considered a lint **failure** and not a warning.

Resolves https://github.com/zmap/zlint/issues/259


lints/lint_subject_contains_reserved_arpa_ip.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	return nil
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "subject_contains_reserved_arpa_ip",
+		Description:   "Checks no subject domain name contains a rDNS entry in an .arpa zone specifying a reserved IP address",
+		Citation:      "BRs: 7.1.4.2.1",
```
cpu: Note to reviewers: I took it as an assumption that the reason issuing for reverse DNS names specifying addresses in an IANA reserved network is against the BR is the chapter and verse that forbids issuing for private IP addresses, "Section 7.1.4.2.1: Subject Alternative Name Extension".

This might not be the best reference and the `EffectiveDate` might need to be tweaked if it isn't.

lints/lint_subject_contains_reserved_arpa_ip.go
```diff
@@ -0,0 +1,234 @@
...
+		zoneSuffix = rdnsIPv6Suffix
+	}
+
+	// Strip off the zone suffix to get only the reversed IP address
+	ipName := strings.TrimSuffix(name, zoneSuffix)
+
+	// A well encoded IPv4 or IPv6 reverse DNS name will have the correct number
+	// of labels to express the address
+	ipLabels := strings.Split(ipName, ".")
+	if len(ipLabels) != numRequiredLabels {
```
cpu: I might be incorrect here in assuming that IPv6 addresses are supposed to be expressed as the full 16 byte address without eliding any components.

I see [real world example certs](https://censys.io/certificates/348284b63609f33461de70ee7353585ffe38d7fed5e02c2e0e9c64d2cce89192) that don't specify 32 nibble labels.

cpu: My reading of https://tools.ietf.org/html/rfc3596#section-2.5 seems to imply that every nibble should have a label. ¯\\\_(ツ)\_/¯

lints/lint_subject_contains_reserved_arpa_ip.go
zakird: That's the best I found when going through the BRs earlier this week too. :/

lints/lint_subject_contains_reserved_arpa_ip.go
cpu: Talking this over with some folks I think that I should perhaps remove the checks that require anything under `ip6.arpa` or `in-addr.arpa` have the correct number of labels to be a rDNS PTR, these are just domain names even if they are malformed rDNS PTRs.

The linter should probably only be checking instances that have the right number of labels?

Input welcome.

titanous: I think a separate warning lint for malformed IPs would make sense.

zakird: Definitely could be argued both ways. On the one hand, it's just a name. On the other hand, there's a clear standard set forth in an RFC. Middle ground might be to have a separate warning lint if it's the wrong number of labels.

zakird: @titanous We're on the same page ;)

cpu: Sounds good to me. I'll split out a separate warning lint for the wrong number of label cases.

cpu: I split the lints in 71facca and updated the PR description. Thanks for the input @zakird @titanous :+1:

lints/lint_subject_contains_malformed_arpa_ip.go
```diff
@@ -0,0 +1,139 @@
+/*
+ * ZLint Copyright 2017 Regents of the University of Michigan
```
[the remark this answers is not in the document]
cpu: Good catch thanks. Fixed in c65664b

lints/lint_subject_contains_malformed_arpa_ip.go
```diff
@@ -0,0 +1,139 @@
...
+func (l *arpaMalformedIP) Execute(c *x509.Certificate) *LintResult {
+	for _, name := range c.DNSNames {
+		name = strings.ToLower(name)
+		var err error
+		if strings.HasSuffix(name, rdnsIPv4Suffix) {
+			// If the name has the in-addr.arpa suffix then it should be an IPv4 reverse
+			// DNS name.
+			err = lintReversedIPAddressLabels(name, false)
+			// If the name has the ipv6.arpa suffix then it should be an IPv6 reverse
+			// DNS name.
```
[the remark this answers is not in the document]
cpu: Oops, totally agree. Fixed this one and the other occurrence in 2e8f7ce.

cpu:
Thanks for the :mag: @titanous !

lints/lint_subject_contains_malformed_arpa_ip.go
```diff
@@ -0,0 +1,139 @@
...
+type arpaMalformedIP struct{}
+
+// Initialize for an arpaMalformedIP linter is a NOP to statisfy linting
+// interfaces.
+func (l *arpaMalformedIP) Initialize() error {
+	return nil
+}
+
+// CheckApplies returns true if the certificate contains any names that end in
+// one of the two designated zones for reverse DNS: in-addr.arpa or ipv6.arpa.
```
DrMattChristian: Comment should reference ip6.arpa. There's no **v** in the zone name. See [RFC](https://tools.ietf.org/html/rfc3152) .

lints/lint_subject_contains_malformed_arpa_ip.go
```diff
@@ -0,0 +1,139 @@
...
+func (l *arpaMalformedIP) Execute(c *x509.Certificate) *LintResult {
+	for _, name := range c.DNSNames {
+		name = strings.ToLower(name)
+		var err error
+		if strings.HasSuffix(name, rdnsIPv4Suffix) {
+			// If the name has the in-addr.arpa suffix then it should be an IPv4 reverse
+			// DNS name.
+			err = lintReversedIPAddressLabels(name, false)
+		} else if strings.HasSuffix(name, rdnsIPv6Suffix) {
+			// If the name has the ipv6.arpa suffix then it should be an IPv6 reverse
```
DrMattChristian: Comment should reference ip6.arpa. There's no **v** in the zone name. See [IETF](https://tools.ietf.org/id/draft-ietf-dnsop-isp-ip6rdns-04.html) .

lints/lint_subject_contains_reserved_arpa_ip.go
```diff
@@ -0,0 +1,232 @@
...
+type arpaReservedIP struct{}
+
+// Initialize for an arpaReservedIP linter is a NOP to statisfy linting
+// interfaces.
+func (l *arpaReservedIP) Initialize() error {
+	return nil
+}
+
+// CheckApplies returns true if the certificate contains any names that end in
+// one of the two designated zones for reverse DNS: in-addr.arpa or ipv6.arpa.
```
DrMattChristian: Comment should reference ip6.arpa. There's no **v** in the zone name. See [RIPE](https://www.ripe.net/manage-ips-and-asns/db/support/documentation/glossary/ip6.arpa) .

lints/lint_subject_contains_reserved_arpa_ip.go
```diff
@@ -0,0 +1,232 @@
...
+func (l *arpaReservedIP) Execute(c *x509.Certificate) *LintResult {
+	for _, name := range c.DNSNames {
+		name = strings.ToLower(name)
+		var err error
+		if strings.HasSuffix(name, rdnsIPv4Suffix) {
+			// If the name has the in-addr.arpa suffix then it should be an IPv4 reverse
+			// DNS name.
+			err = lintReversedIPAddress(name, false)
+		} else if strings.HasSuffix(name, rdnsIPv6Suffix) {
+			// If the name has the ipv6.arpa suffix then it should be an IPv6 reverse
```
DrMattChristian: Comment should reference ip6.arpa. There's no **v** in the zone name. See [link](http://www.gestioip.net/docu/ipv6_reverse_dns_delegation.html) .

cpu:
@DrMattChristian Thanks: https://github.com/zmap/zlint/pull/271

DrMattChristian:
@cpu Thanks for the quick PR on these comment nits. Your changes in #271 look correct to me.

PR #263, opened 2019-03-02
lints: ballot 144/201 .onion linting.

cpu:
This branch adds three new lints based on [CABF Ballot 144](https://cabforum.org/2015/02/18/ballot-144-validation-rules-dot-onion-names/) and [CABF Ballot 201](https://cabforum.org/2017/06/08/2427/) specific to the allowance of `.onion` subject names under several additional requirements. Presently all of the new lints are made to be effective as of May 1st, 2015 but that may need to be tweaked.

* `lint_san_dns_name_onion_not_ev_cert.go` validates that any subscriber certificates containing a `.onion` subject are EV certificates. 
* `lint_ext_tor_service_descriptor.go` validates that certificates with one or more `.onion` subjects have the correct `cabf-TorServiceDescriptor` extension, and a well formed `TorServiceDescriptorHash` object for each `.onion` subject.
* `lint_onion_subject_validity_time_too_large.go` validates certificates with one or more `.onion` subjects do not have a validity period larger than 15 months.

These linters will hopefully help catch miss-issuance events like [this one by DigiCert in March 2018](https://groups.google.com/d/msg/mozilla.dev.security.policy/7NzJgDomx_M/jvTjVbLCAgAJ).

I'd appreciate some extra :eyes: on my parsing of the `TorServiceDescriptorHash` ASN.1. As far as I know there aren't any other open source projects that have implemented this so I wasn't able to spot check against someone else's code. `AWSLabs/Certlint` [has not yet implemented this linting](https://github.com/awslabs/certlint/issues/62) at the time of writing.

Resolves #14

Issue #14, opened 2017-03-13
.onion lint


kumarde:
The EV Guidelines require certificates issued for .onion include the cabf-TorServiceDescriptor extension, defined in the EV Guidelines, as part of these certificates. This is required by Section 11.7.1 (1) of the EV Guidelines, reading: "For a Certificate issued to a Domain Name with .onion in the right-most label of the Domain Name, the CA SHALL confirm that, as of the date the Certificate was issued, the Applicant’s control over the .onion Domain Name in accordance with Appendix F."

cpu:
Updating this issue based on [the plan discussed in #263](https://github.com/zmap/zlint/pull/263#discussion_r262651547):

- [X] - Close the old PR that combined parsing and linting (#263)
- [X] - Open a new PR with the two lints that don't rely on `TorServiceDescriptorHash` parsing (https://github.com/zmap/zlint/pull/265)
- [ ] - Open a `zcrypto` PR to add `TorServiceDescriptorHash` parsing
- [ ] - Open a PR to add a lint for the `TorServiceDescriptorHash` that uses the `zcrypto` parsing logic.

PR #265, opened 2019-03-06
lints: enforce .onion certs are EV, have <15mo expiry

cpu:
This is a revised version of #263 with a smaller scope.

This branch adds two new lints based on [CABF Ballot 144](https://cabforum.org/2015/02/18/ballot-144-validation-rules-dot-onion-names/) specific to the allowance of `.onion` subject names under several additional requirements. Both lints are configured to be effective as of May 1st, 2015 based on the Ballot 144 text.

* `lint_san_dns_name_onion_not_ev_cert.go` validates that any subscriber certificates containing a `.onion` subject are EV certificates. 
* `lint_onion_subject_validity_time_too_large.go` validates certificates with one or more `.onion` subjects do not have a validity period larger than 15 months.

Notably missing from this PR is any parsing or validation of the required `TorServiceDescriptorHash` x509 extension (first specified in Ballot 144, later updated in 198 and 201). That will be done as a follow-up blocked on a `zcrypto` PR.

Updates #14


lints/lint_onion_subject_validity_time_too_large.go
```diff
@@ -0,0 +1,68 @@
...
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name: "onion_subject_validity_time_too_large",
+		Description: fmt.Sprintf(
+			"certificates with .onion names can not be valid for more than %d months",
+			maxOnionValidityMonths),
+		Citation:      "CABF EV Guidelines: Appendix F",
```
cpu: Compared to #263 I've updated this lint to cite a new `CABFEVGuidelines` source and the correct section of the current version at the time of writing [v1.6.8](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-EV-Guidelines-v1.6.8.pdf)

lints/lint_san_dns_name_onion_not_ev_cert.go
```diff
@@ -0,0 +1,78 @@
...
+	return util.IsSubscriberCert(c) && util.CertificateSubjInTLD(c, onionTLD)
+}
+
+// Execute checks that if the certificate contains any subject names ending in
+// the  .onion TLD that it is also an EV certificate. If there are .onion names
+// and the certificate is not an EV certificate an Error LintResult is returned.
+func (l *onionNotEV) Execute(c *x509.Certificate) *LintResult {
+	isEV := util.IsEV(c.PolicyIdentifiers)
+	names := append(c.DNSNames, c.Subject.CommonName)
+	for _, name := range names {
```
zakird: This code seems duplicative of what's in CheckApplies. Is there a difference why the loop is explicit here?

lints/lint_san_dns_name_onion_not_ev_cert.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	}
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "san_dns_name_onion_not_ev_cert",
+		Description:   "certificates with a .onion subject name must be issued in accordance with EV Guidelines",
+		Citation:      "CABF Ballot 144",
```
cpu: I left this lint citing Ballot 144 despite [feedback to do otherwise](https://github.com/zmap/zlint/pull/263#discussion_r261844580).

[Ballot 144](https://cabforum.org/2015/02/18/ballot-144-validation-rules-dot-onion-names/) added the text that I think most directly maps to this linter. It updated section 9.2.1 ("Subject Alternative Name Extension") of the BRs (v1.2.3 at the time of Ballot 144) to add:
> Effective May 1, 2015, each CA SHALL revoke all unexpired Certificates with an Internal Name using onion as the right-most label in an entry in the subjectAltName Extension or commonName field unless such Certificate was issued in accordance with Appendix F of the EV Guidelines.

I'm not sure when that text was lost, because in the current BRs ([v1.6.3](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.6.3.pdf) at the time of writing) I can't find _any_ mention of `.onion` except in section 1.2.1. which is just a table of revisions. Looking at section 7.1.4.2.1 ("Subject Alternative Name Extension") of BRs v1.6.3 - the section I think maps closest to v1.2.3's equivalent section -  there is no longer any mention of `.onion`.

Perhaps there's an argument to be made that the correct citation is BRs section 7.1.4.2.1 because that section forbids internal names in a SAN. The v1.6.3 BRs define an "internal name" in Section 1.6.1 "Definitions" as:
> A string of characters (not an IP address) in a Common Name or Subject Alternative Name field of a Certificate that cannot be verified as globally unique within the public DNS at the time of certificate issuance because it does not end with a Top Level Domain registered in IANA’s Root Zone Database.

The `.onion` top level domain isn't delegated from the root zonein the public DNS so it seems as though by the BRs alone a `.onion` name would be forbidden in a SAN. The part that seems missing here is the connection from being forbidden as an internal name in a non-EV cert to being allowed in the EV guidelines. Section 11.7.1 ("Verification Requirements") of the EV guidelines explicitly allows `.onion` provided the rules in Appendix F are followed.

Phew! That's a lot of words. The tl;dr is that I think citing the ballot makes the most sense but if someone has a better picture of the history and prefers an alternative citation I'd be happy to oblige.

lints/lint_san_dns_name_onion_not_ev_cert.go
cpu: The loop was duplicated so it could identify the `name` that was present with a `.onion` TLD in the error message.

Looking at it again I don't think it merits the duplication, especially since it would only ever include the _first_ `.onion` subject in the error. I refactored `Execute` in fae6210 to remove the duplicated loop, thanks @zakird !

Issue #264, opened 2019-03-05
Add Go module support, pin dependencies


cpu:
Supporting [Go modules](https://github.com/golang/go/wiki/Modules) would make `zlint`'s dependencies explicit. This would help with situations like the one discussed in https://github.com/zmap/zlint/pull/263 where a lint may depend on new functionality in `zcrypto`.

There are alternatives to Go modules for tracking dependencies (godeps, gvt, etc) but [Go 1.13+](https://blog.golang.org/modules2019) will make modules supported by default and I expect much of the Go ecosystem will standardize on modules vs the alternatives.

dadrian:
I strongly agree with this.

cpu:
@dadrian Great! Related to module support do you have any thoughts about whether to commit vendored dependencies to the repo or not?

I'm generally in favour of doing that because it makes builds faster and helps eliminate occasional flakyness from `go get` but I understand that some folks in the Go community think its unnecessary since the vendored deps can be reconstructed from the `go.mod`.

zakird:
I don't have any issue with vendorizing the dependencies.

cpu:
I have a PR for this up for review: https://github.com/zmap/zlint/pull/266

PR #266, opened 2019-03-07
Add Go module support, cleanup Makefile

cpu:
This branch adds support for Go modules and cleans up a little bit. At a high level:

* Added `zlint-gtld-update` to the build artifacts
* Fixed an import case sensitivity problem that blocked go module support
* Cleaned up the `makefile`
* Added Go module support
* Vendored dependencies

Resolves https://github.com/zmap/zlint/issues/264


go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/zmap/zlint
+
+require (
+	github.com/asaskevich/govalidator v0.0.0-20180720115003-f9ffefc3facf
```
cpu: It looks like this dependency is only present because of `govalidator.IsURL` being used in `util/fqdn.go`. Similarly `zcrypto` carries a `govalidator` dep just for `IsURL`.

It might be nice to knock off a dependency off of both projects by implementing this one function in `zcrypto`.

cpu: I opened https://github.com/zmap/zcrypto/issues/187 on `zcrypto` for this.

zakird:
I agree regarding implementing `IsURL`, but I'm going to merge for the sake of forward progress. We can remove the dependency at a later date.

PR #267, opened 2019-03-11
test for lint_ext_authority_key_identifier_no_key_identifier

zakird:


cpu:
I'm going to go out on a limb and merge this since it's just a unit test. If folks would prefer I wait for more :mag:'s on this sort of PR speak up and I won't do it again :-)

PR #269, opened 2019-03-12
lints: clarify dsa_improper_modulus description

cpu:
While reviewing https://github.com/zmap/zlint/pull/268 I found the existing lint `Description` didn't match exactly to what `Execute` was checking. I think this small tweak will help clarify.


cpu:
Thanks for the :mag: @titanous !

Issue #283, opened 2019-05-21
Lint rsaEncryption algorithm identifiers for mandatory NULL params


cpu:
[RFC 3279, Section 2.3.1 "RSA Keys"](https://tools.ietf.org/html/rfc3279#section-2.3.1) says the following about the `rsaEncryption` algorithm identifier encoding:

```
   The rsaEncryption OID is intended to be used in the algorithm field
   of a value of type AlgorithmIdentifier.  The parameters field MUST
   have ASN.1 type NULL for this algorithm identifier.
```

It was noted by [@sleevi on MDSP](https://groups.google.com/d/msg/mozilla.dev.security.policy/NFAG2_o3dhY/gQ2UW_ohAwAJ) that there are real-world certificates that violate this aspect of RFC 3279 and are flagged by `certlint` but not `zlint`.

Zlint should have a lint that ensures the required `NULL` parameters field of `rsaEncryption` algorithm identifiers in certificates is present. 

[...]

PR #285, opened 2019-05-21
lints: add lint for mandatory NULL RSA alg. id params.

cpu:
This PR adds a new `rsaAlgIDNullParams` lint that returns a fatal lint result if the provided certificate encodes an RSA algorithm identifier SEQUENCE without the required NULL parameters field.

The baseline requirements mandate that certificates obey RFC 5280, which in turn defers to RFC 3279 for the encoding of the RSA algorithm identifier. RFC 3279 Section 2.3.1 "RSA keys" says:

> The rsaEncryption OID is intended to be used in the algorithm field of a value of type AlgorithmIdentifier. The parameters field MUST have ASN.1 type NULL for this algorithm identifier.

Resolves https://github.com/zmap/zlint/issues/283


cpu:
For completeness here's `zcertificate` output using `zlint` without this PR and with this PR to lint the first example certificate from Ryan's MDSP post:

* [Without](https://gist.github.com/cpu/18fc9ae80785d8cae2cfea37a16d3077)
* [With](https://gist.github.com/cpu/601bf79797ae86c8be60f838ea00328d)

In the [with](https://gist.github.com/cpu/601bf79797ae86c8be60f838ea00328d) output there is a fatal lint result from the `rsa_algid_null_params` lint:
```
      "rsa_algid_null_params": {
        "result": "fatal",
        "details": "certificate contains RSA public key algorithm identifier missing required NULL parameters"
      },
```

lints/lint_rsa_algid_null_params.go
```diff
@@ -0,0 +1,103 @@
...
+			Details: fmt.Sprintf("Error unmarshaling raw pkixPublicKey: %v", err),
+		}
+	} else if len(rest) != 0 {
+		return &LintResult{
+			Status:  Error,
+			Details: fmt.Sprintf("Trailing data after pkixPublicKey: %v", err),
+		}
+	}
+
+	params := pkixPublicKey.Algo.Parameters
```
sleevi: Params is asn1:optional for pkix.AlgorithmIdentifier, so if params are omitted, what is the value of this field?

lints/lint_rsa_algid_null_params.go
```diff
@@ -0,0 +1,103 @@
...
+
+	if params.Tag != asn1.TagNull {
+		return &LintResult{
+			Status:  Fatal,
+			Details: "certificate contains RSA public key algorithm identifier missing required NULL parameters",
+		}
+	}
+
+	if !bytes.Equal(params.Bytes, []uint8{}) {
```
sleevi: Would this be simpler by checking len(params.Bytes) == 0? I wasn't sure why this explicit compare?

lints/lint_rsa_algid_null_params.go
```diff
@@ -0,0 +1,103 @@
...
+	} else if len(rest) != 0 {
+		return &LintResult{
+			Status:  Error,
+			Details: fmt.Sprintf("Trailing data after pkixPublicKey: %v", err),
+		}
+	}
+
+	params := pkixPublicKey.Algo.Parameters
+
+	if params.Tag != asn1.TagNull {
```
sleevi: Looking at how asn1.unmarshal is implemented, doesn't this create the risk that either the class or the compound-type may be misconstructed? The memcmp approach proposed from the associated thread was, in part, to ensure that it'd be flagged as a Universal class, non-compound. If you want to provide more descriptiveness, then it's probably worth adding checks.

One way you could restructure this to be more defense-in-depth would be to early execute the compare with the 'raw' value. If the raw value was equal, you know it's a pass, otherwise, you know there is 'something' wrong. The unmarshal and the remainder of the function is thus trying to find out 'what' is wrong and add a descriptive error to it.

WDYT of that approach?

lints/lint_rsa_algid_null_params.go
cpu: Yup! I think I had originally been (mistakenly) fooling with a comparison against `asn1.NullRawValue` and should have replaced the `bytes.Equal`.

lints/lint_rsa_algid_null_params.go
cpu: Good question, I'll take a look.

lints/lint_rsa_algid_null_params.go
cpu: That sounds more robust to me. I'll take a shot at doing that over the next few days.

cpu:
@sleevi Thanks for the :mag:! Much appreciated :-)

lints/lint_rsa_algid_null_params.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Initialize is a NOP.
+func (l *rsaAlgIDNullParams) Initialize() error {
+	return nil
+}
+
+// CheckApplies returns true for certificates with an RSA public key.
+func (l *rsaAlgIDNullParams) CheckApplies(c *x509.Certificate) bool {
+	_, ok := c.PublicKey.(*rsa.PublicKey)
+	return ok && c.PublicKeyAlgorithm == x509.RSA
```
tadukurow: perhaps checking 

`c.PublicKeyAlgorithmOID.Equal(rsaEncryptionOID)`

is slightly more future-proof as RFC4055 has some additional rules for pub keys with RSA-PSS identifiers and someone might use x509.RSA as a blanket RSA categorisation in the future.

lints/lint_rsa_algid_null_params.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	if rest, err := asn1.Unmarshal(raw, &pkixPublicKey); err != nil {
+		return &LintResult{
+			Status:  Error,
+			Details: fmt.Sprintf("Error unmarshaling raw pkixPublicKey: %v", err),
+		}
+	} else if len(rest) != 0 {
+		return &LintResult{
+			Status:  Error,
```
tadukurow: I think you switched Error and Fatal. Fatal is used for parsing errors in other lints.

lints/lint_rsa_algid_null_params.go
tadukurow: Actually knocked up this lint using cryptobyte today before I saw this PR.
Sharing here in case you wanted a look.

```
        var input = cryptobyte.String(c.RawSubjectPublicKeyInfo)

	var publicKeyInfo cryptobyte.String
	if !input.ReadASN1(&publicKeyInfo, asn1_cryptobyte.SEQUENCE) {
		return &LintResult{Status: Fatal, Details: "error reading pkixPublicKey"}
	}

	var algorithm cryptobyte.String
	if !publicKeyInfo.ReadASN1(&algorithm, asn1_cryptobyte.SEQUENCE) {
		return &LintResult{Status: Fatal, Details: "error reading pkixPublicKey algorithm"}
	}

	if !algorithm.SkipASN1(asn1_cryptobyte.OBJECT_IDENTIFIER) {
		return &LintResult{Status: Fatal, Details: "error reading pkixPublicKey algorithm OID"}
	}

	if algorithm.Empty() {
		return &LintResult{Status: Error, Details: "certificate contains RSA public key algorithm identifier missing required NULL parameter"}
	}

	if !algorithm.PeekASN1Tag(asn1_cryptobyte.NULL) {
		return &LintResult{Status: Error, Details: "certificate contains RSA public key algorithm identifier with non-NULL parameter"}
	}

	return &LintResult{Status: Pass}
```

lints/lint_rsa_algid_null_params.go
cpu: @tadukurow That seems sensible but I think it would be best if you could open a separate issue for that feedback. I took this approach based on what was done elsewhere in the codebase:

```
lints/lint_rsa_exp_negative.go
32:	return ok && c.PublicKeyAlgorithm == x509.RSA

lints/lint_old_sub_cert_rsa_mod_less_than_1024_bits.go
33:	return ok && c.PublicKeyAlgorithm == x509.RSA && !util.IsCACert(c) && endDate.Before(util.NoRSA1024Date)

lints/lint_rsa_public_exponent_too_small.go
37:	return ok && c.PublicKeyAlgorithm == x509.RSA

lints/lint_rsa_mod_not_odd.go
38:	return ok && c.PublicKeyAlgorithm == x509.RSA

lints/lint_rsa_mod_less_than_2048_bits.go
36:	return ok && c.PublicKeyAlgorithm == x509.RSA && c.NotAfter.After(util.NoRSA1024Date.Add(-1))

lints/lint_old_root_ca_rsa_mod_less_than_2048_bits.go
37:	return ok && c.PublicKeyAlgorithm == x509.RSA && util.IsRootCA(c) && issueDate.Before(util.NoRSA1024RootDate)

lints/lint_rsa_public_exponent_not_odd.go
37:	return ok && c.PublicKeyAlgorithm == x509.RSA

lints/lint_rsa_public_exponent_not_in_range.go
42:	return ok && c.PublicKeyAlgorithm == x509.RSA

lints/lint_rsa_mod_factors_smaller_than_752_bits.go
37:	return ok && c.PublicKeyAlgorithm == x509.RSA
```

I think it would make sense to adjust some (all?) of those other lints at the same time.

sleevi:
So, all of those other checks are truly generic, regardless of the
algorithm.

Currently, the BRs and Moz Policy are not prescriptive on the SPKI
algorithm used - both rsaEncryption and rsaPSS are presumed supported and
permitted, by virtue of 4055. Those referenced checks are all applicable,
because they are requirements expressed in terms of “RSA keys” in the BRs,
thus applying to both.

The validation of parameters here only technically applies to rsaEncryption
keys - because I’m proposing on MDSP to explicitly prohibit using rsaPSS as
the SPKI OID (it causes compat issues in Apple, Mozilla, and Google
products, to say the least), and only permit the rsaEncryption OID for
SPKIs. The change also would prohibit the OAEP OID, which isn’t applicable
for TLS certs but might be for others.

I suspect that the params check - this one - would be better scoped just to
the rsaEncryption OID, and then a separate check (if Moz Policy adopts my
PR) to make sure ONLY rsaEncryption is used for RSA keys. That way, even if
Moz rejects my PR, you’ve still got a linter for the RFC 4055 bits for
rsaEncryption keys, and we can write another one for the 4055 bits of
rsaPSS keys, since they would be allowed (by rejecting the PR)

tadukurow:
Hi @cpu, did you have any time to look at this? 
We're eager to get this in and I have a lint ready on this, so I could open a PR for that in case you're short on time.

cpu:
> Hi @cpu, did you have any time to look at this?

Apologies, I was travelling last week and fell behind. I have time to dedicated to this today.

> We're eager to get this in and I have a lint ready on this, so I could open a PR for that in case you're short on time.

@tadukurow If you want to open a PR that would be great. I could iterate on my branch but I think it will result in a slower time-to-fix because I'll have to solicit another maintainer to review the work. If you open a PR I can review it and merge it as soon as its ready to go. If there's anything useful to you (testcases or otherwise) from my PR feel free to adopt it.

cpu:
> I suspect that the params check - this one - would be better scoped just to the rsaEncryption OID

@sleevi You make a convincing argument :+1: I'm in agreement now that @tadukurow's suggested change to `CheckApplies` makes sense.

lints/lint_rsa_algid_null_params.go
cpu: Discussion on this feedback ended up un-threading, consensus is that this particular lint deserves the more robust OID check. The other referenced lints are more generic.

tadukurow:
@cpu I opened #286. The testcases and files were definitely handy, thanks.

cpu:
Closing this PR in favour of #286

PR #286, opened 2019-05-31
lints for RFC4055 check RSA AlgorithmIdentifier param is NULL for SPKI and tbsCertificate.signature

tadukurow:
PR based on #285.

Benchmarked a byte compare of a precomputed algo byte slice, but performance difference was negligible and cryptobyte version more readable.


tadukurow:
I updated travis to run go1.12.5 as the gofmt test failed and passed on latest. Turned out it was some local go version shenanigans on my machine. Happy to revert back to go1.12 if necessary.

go.mod
```diff
@@ -5,6 +5,7 @@ require (
 	github.com/sirupsen/logrus v1.3.0
 	github.com/weppos/publicsuffix-go v0.4.0
 	github.com/zmap/zcrypto v0.0.0-20190329181646-dff83107394d
+	golang.org/x/crypto v0.0.0-20180904163835-0709b304e793
```
cpu: @tadukurow This looks like an older commit of `golang.org/x/crypto` than was in the `go.mod` before this PR. Is there a reason you had to revert to an older version or could this diff be backed out of this branch?

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/*******************************************************************************************************
+"RFC5280: RFC 4055, Section 1.2"
```
cpu: Good call ref'ing RFC 4055 instead of RFC 3279 :+1:

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"encoding/asn1"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/util"
+	"golang.org/x/crypto/cryptobyte"
+	asn1_cryptobyte "golang.org/x/crypto/cryptobyte/asn1"
+)
+
+var rsaEncryptionOID = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 1, 1}
```
cpu: Can you move this to an exported var in the `util` package? I think `util/oid.go` is the best place, perhaps near `OidRSASSAPSS`?

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+func (l *rsaEncryptionParamNotNULL) CheckApplies(c *x509.Certificate) bool {
+	// explicitly check for rsaEncryptionOID, as RSA-PSS or RSA-OAEP certificates might be classified with c.PublicKeyAlgorithm = RSA
+	return c.PublicKeyAlgorithmOID.Equal(rsaEncryptionOID)
+}
+
+func (l *rsaEncryptionParamNotNULL) Execute(c *x509.Certificate) *LintResult {
+	var input = cryptobyte.String(c.RawSubjectPublicKeyInfo)
```
cpu: tiny nit: I would probably write this as `input := cryptobyte.String(c.RawSubjectPublicKeyInfo)`.

vendor/modules.txt
```diff
@@ -8,10 +8,12 @@ github.com/sirupsen/logrus
 github.com/weppos/publicsuffix-go/publicsuffix
 # github.com/zmap/zcrypto v0.0.0-20190329181646-dff83107394d
 github.com/zmap/zcrypto/x509
+github.com/zmap/zcrypto/x509/ct
 github.com/zmap/zcrypto/x509/pkix
 github.com/zmap/zcrypto/json
-github.com/zmap/zcrypto/x509/ct
 # golang.org/x/crypto v0.0.0-20180904163835-0709b304e793
+golang.org/x/crypto/cryptobyte
```
cpu: It seems reasonable to add deps on `/x/crypto/cryptobyte` and `/x/crypto/cryptobyte/asn1`. `zlint` was already carrying an indirect dependency on `/x/crypto` by way of `github.com/sirupsen/logrus` :+1:

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: go
 dist: trusty
 
 go:
-  - "1.12"
+  - "1.12.5"
```
cpu: Would you mind changing this to use whatever the latest point release on the `1.12.x` series is? [Travis docs](https://docs.travis-ci.com/user/languages/go/) say to use `"1.12.x"` for that.

That was my intention in https://github.com/zmap/zlint/commit/64ec0afbd7174756179985c660d7e47141528626 but I goofed and wrote `"1.12"` instead of `"1.12.x"`. 

It looks like CI [has been running](https://travis-ci.org/zmap/zlint/builds/535448296) with exactly `1.12.0` which doesn't make much sense:
> $ travis_setup_go
> go version go1.12 linux/amd64

go.mod
cpu: D'oh, I was comparing this to the `zcrypto` ref. I see now this isn't a change in an import version but a new addition. Apologies.

.travis.yml
tadukurow: That makes more sense indeed.

lints/lint_rsa_signature_algo_parameter_not_null.go
tadukurow: No problem

lints/lint_rsa_signature_algo_parameter_not_null.go
tadukurow: Good idea.

cpu:
@tadukurow I think this is ready to be merged. I'm going to give it ~24 hrs in case @sleevi or @zakird have cycles to give it a quick :mag:. I will merge this Monday afternoon otherwise. I think its a definite improvement over the status quo and we can always iterate if there is late feedback. Thanks again for pitching in on this lint! :tada:

sleevi:
Looks like the email gateway didn’t work.

I don’t think this is a good approach, and is exactly the approach I was trying to warn against with the design of the Mozilla PR, because it still allows invalid ASN.1 through.

In this case, for example, I could use a compound NULL or a non-zero-length NULL to include additional data. The first check for these really should be the memcmp - there is only a single possible correct encoding. The remainder - using or parsing ASN.1 - is merely about finding out where the issue is.

cpu:
> I don’t think this is a good approach, and is exactly the approach I was trying to warn against with the design of the Mozilla PR, because it still allows invalid ASN.1 through.

I think its better than nothing (in that it will fail otherwise passing missisuances found in CT) but I definitely see your point. I'm sorry its taken me three passes to really catch on to the emphasis on a direct memcpy against the one correct encoding :man_facepalming: 

> In this case, for example, I could use a compound NULL or a non-zero-length NULL to include additional data. The first check for these really should be the memcmp - there is only a single possible correct encoding. The remainder - using or parsing ASN.1 - is merely about finding out where the issue is.

@tadukurow Can you lightly refactor this branch to determine the pass/fail status based on a byte-wise comparison against the known correct encoding and use the existing `cryptobyte` code to compute the detail field.

sleevi:
If you need help making adversarial test cases - https://github.com/google/der-ascii is incredibly useful and what I use for this. You can easily mutate an SPKI into various forms with some simple commands, and then sign it as a cert with a test key.

tadukurow:
@sleevi thanks for the link to ascii-der, very useful. 

Added some test cases by changing spki only to not blow up test certs and get around some pem parsing issues with malformed asn1.

I've expanded the cryptobyte check. It now makes sure the NULL tag is definitely empty and to check there is no trailing data. Also added the byte comparison check.

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (l *rsaEncryptionParamNotNULL) CheckApplies(c *x509.Certificate) bool {
+	// explicitly check for util.OidRSAEncryption, as RSA-PSS or RSA-OAEP certificates might be classified with c.PublicKeyAlgorithm = RSA
+	return c.PublicKeyAlgorithmOID.Equal(util.OidRSAEncryption)
+}
+
+func (l *rsaEncryptionParamNotNULL) Execute(c *x509.Certificate) *LintResult {
+	// Try byte comparison for Algorithm
+	// Determine offset (SEQUENCE + length) first. Notably 1024 encodes length in 2 bytes, 2048 and up in 3 bytes
```
cpu: I think this comment about how you determine the offset would be better as a comment on the `getOffset` function implementation. It's a detail specific to how that function computes the offset and I think when you're invoking the function it's sufficient to know its a byte offset for comparison.

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,118 @@
...
+		Name:          "e_rsa_encryption_parameter_not_null",
+		Description:   "RSA: Encoded algorithm identifier MUST have NULL parameters",
+		Citation:      "RFC 4055, Section 1.2",
+		Source:        RFC5280, // RFC4055 is referenced in RFC5280, Section 1
+		EffectiveDate: util.RFC5280Date,
+		Lint:          &rsaEncryptionParamNotNULL{},
+	})
+}
+
+func getOffset(asn1Sequence []byte) int {
```
cpu: Can the name of this function be updated to be a little bit more informative?

If I'm understanding correctly this is computing the offset to the start of the algorithm object identifier `SEQUENCE` byte offset in the `RawSubjectPublicKeyInfo` `SEQUENCE`. If that's accurate maybe a clearer name would be `getOIDOffset` ?

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,118 @@
...
+	if len(asn1Sequence) < 2 {
+		return 0
+	}
+
+	if asn1Sequence[1]&0x80 == 0 {
```
cpu: I think its a kindness to readers to add a comment when you're using a bitwise operation against a constant like this to mention the higher level purpose. How about moving the comment on the next line up and tweaking it slightly to be something like:

"Check bit 8 to determine length encoding type. If bit 8 is zero it is a short form length encoded in 1 octet" ?

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,118 @@
...
+		// short form length encoding in 1 octet
+		return 2 // 1 tag octet +1 length encoding octet
+	}
+
+	// long form length encoding
+	length := int(asn1Sequence[1] & 0x7f)
```
cpu: Ditto here RE: adding a brief explanation of the `& 0x7f` operation.

How about tweaking the existing comment to say something like:

"Using long form length encoding. Clear the length encoding indicator in bit 8, the remaining bits contain the length"

lints/lint_rsa_signature_algo_parameter_not_null.go
```diff
@@ -0,0 +1,118 @@
...
+		Description:   "RSA: Encoded algorithm identifier MUST have NULL parameters",
+		Citation:      "RFC 4055, Section 1.2",
+		Source:        RFC5280, // RFC4055 is referenced in RFC5280, Section 1
+		EffectiveDate: util.RFC5280Date,
+		Lint:          &rsaEncryptionParamNotNULL{},
+	})
+}
+
+func getOffset(asn1Sequence []byte) int {
+	if len(asn1Sequence) < 2 {
```
cpu: Can you explain this clause? At face-value it seems like a case I'd expect to see an `error` value returned. Shouldn't the `rawSubjectPublicKeyInfo` `SEQUENCE` always be at least 1 tag byte and 1 length byte?

Should this lint also be enforcing the correct `rawSubjectPublicKeyInfo` `SEQUENCE` tag is present? I think the current implementation of `getOffset` will allow an invalid outer `SEQUENCE` as long as the DER is >= 2 bytes in length.

If we're less concerned about the outer `SEQUENCE` in this lint then maybe it would be acceptable to use `cryptobyte` to consume the `SEQUENCE` instead of using this `getOffset` function so that the lint doesn't need to concern itself with the length octet encoding and can just perform the bytewise comparison of the expected bytes? 

@sleevi What do you think?

sleevi: I think of this lint as being about "the encoding within the `AlgorithmIdentifier`", and thus less important about the context the `AlgorithmIdentifier` appears in.

That is, we can have AlgId in the `TBS.subjectPublicKeyInfo` and in the `TBS.signature` fields, and we should lint both of those. The outer certificates' `Certificate.signatureAlgorithm` is mutable by adversaries (since it's unsigned), so I'm less interested in that.

Under that framing, I think ignoring the outer `SEQUENCE` of the `subjectPublicKeyInfo` is fine, and I agree, using cryptobyte is probably a great (and safe) way to do that. If we're worried about trailing data on SPKI (and we should be, to be fair), I think that's probably as a separate lint, as we would trailing data in any other field?

cpu: > That is, we can have AlgId in the TBS.subjectPublicKeyInfo and in the TBS.signature fields, and we should lint both of those.

@sleevi That makes sense. I think linting the `TBS.signature` field in this way will be mean needing to spelunk the coarse `RawTBSCertificate []byte` struct field with `cryptobyte` to find the alg id offset since there isn't a more targeted equiv. to `RawSubjectPublicKeyInfo` on the certificate struct given to the linter. I'm slightly in favour of keeping momentum on the existing `TBS.subjectPublicKeyInfo` focused lint and leaving a TODO for the Signature field because of that requirement for more new code. I could probably be persuaded its worth blocking on if you think its important to cover this field immediately.

@tadukurow What do you think? Do you have a pref. with regards to linting the signature RSA algorithm identifier in this branch vs in a follow-up? 

> Under that framing, I think ignoring the outer SEQUENCE of the subjectPublicKeyInfo is fine, and I agree, using cryptobyte is probably a great (and safe) way to do that.

Great :+1: Thanks for the input! I'll update my review feedback to encourage switching the `getOffset` code to equiv usage of `cryptobyte`.

>  If we're worried about trailing data on SPKI (and we should be, to be fair), I think that's probably as a separate lint, as we would trailing data in any other field?

Agreed, I think that would be a candidate for a separate lint.

Thanks again!

lints/lint_rsa_signature_algo_parameter_not_null.go
cpu: @tadukurow Based on discussion in [the other feedback thread](https://github.com/zmap/zlint/pull/286#discussion_r289924518) I think you can replace `getOffset` entirely by using `cryptobyte` to unpack your way into the `RawSubjectPublicKeyInfo` to the right offset for the comparison. Most of my other feedback will be resolved by switching this code out.

I think that will be more readable and remove the DER-specific details in `getOffset` that are unrelated to the linting. 

Thanks!

lints/lint_rsa_signature_algo_parameter_not_null.go
tadukurow: So to clarify, we don't care about trailing bytes in the SPKI, including trailing bytes after the Parameter?

cpu: > So to clarify, we don't care about trailing bytes in the SPKI, including trailing bytes after the Parameter?

That sounds reasonable to me :+1: A separate lint focused on trailing data could be added afterwards covering more fields.

sleevi: I think it’s pretty easy to check for trailing bytes here for the inner `SEQUENCE`, but the outer `SEQUENCE` is about an `RSAPublicKey` and that’s not what you’re linting here, if that makes sense.

So I’m a fan of checking after params but not algId

tadukurow: Thought as much. I'll get that in and have a look at adding the lint for signatureAlgorithm.
Do we care much about the detailed error messages or is simple failure message enough?
Using CB to get offset and checking for trailing data, means the SPKI needs to be reparsed to get the detailed error messages. Removing that path would make it a lot simpler.

cpu: I think the detailed error messages are valuable and the complexity isn't too great.

tadukurow:
I think all issues have been addressed, I've added a lint for TBS signature field as well.

cpu:
@tadukurow Can you update the PR description so there is a good commit message ready for a squash merge? The branch was split into two separate lints and I think the description needs catching-up. It seems like CI is maybe stuck and perhaps another commit on the branch will :boot: it.

@sleevi I gave this a close review and I believe the way that it is implemented now meets the spirit of your feedback so far. I'd appreciate if you could give this one more :mag: if you had time. Particularly the added `lint_tbs_signature_rsa_encryption_parameter_not_null.go` lint for the signature algorithm parameters. Does the `cryptobyte` parsing of `c.RawTBSCertificate` to get to the DER encoded algorithm identifier look OK to you?

sleevi:
@cpu Yeah, I think this addresses my feedback as well. Thanks @tadukurow for humouring me in going for a thorough lint here. This definitely helps!

cpu:
I'm going to give Travis and Github some time to un-stick whatever job queue is holding the CI status for this branch back. If it hasn't un-stuck in the next 24hrs I'll figure out how to unstick it so it can be merged.

tadukurow:
My pleasure. Saw a few places where cryptobyte could give some more precision, so now the dependency is in I'll probably open one or two follow up PRs for that.

zakird:
I kicked travis. Should be good to go.

PR #287, opened 2019-06-06
expand subject DN informational only lint

tadukurow:
Expanded the check for no metadata only in subject DN to check for subject DN fields containing no characters in a-Z0-9 or outside of ascii table. This catches more than just checking for ".", "-", " ". Also remove separate checks for serial and domainComponent as they are part of pkix.Names so separate checking was redundant.

This will help zlint catch issues that CABLint catches today such as 
https://crt.sh/?id=106177929&opt=zlint,cablint,x509lint 
https://crt.sh/?id=134328239&opt=cablint,zlint
https://crt.sh/?id=26408912&opt=cablint,zlint

As far as I know there aren't any combinations of characters outside of a-Z0-9 and UTF8 which wouldn't be described as metadata only. This assumption needs some scrutiny though.
The new check is still susceptible to UTF8 runes metadata only.


cpu:
@tadukurow Can you resolve conflicts with master when you have a chance? I'll endeavour to give this a :mag: in the next few days.

tadukurow:
@cpu did you mean rebase off master? because I don't see any conflicts

cpu:
> @cpu did you mean rebase off master? because I don't see any conflicts

Yup! I mis-read the Github status message. On the repositories I spend most of my time we don't have the "Require branches to be up to date before merging" checkbox checked and so only conflicts cause that kind of status block. Zlint requires all PRs be current with master.

lints/lint_subject_contains_noninformational_value.go
```diff
@@ -76,3 +65,15 @@ func init() {
 		Lint:          &illegalChar{},
 	})
 }
+
+// checkAlphaNumericOrUTF8Present checks if input string contains at least one occurrence of [a-Z0-9] or
+// a UTF8 rune outside of ascii table
+func checkAlphaNumericOrUTF8Present(input string) bool {
+	for _, r := range input {
+		if (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z') || (r >= '0' && r <= '9') || r > 127 {
```
cardonator: Couldn't this use encodings.IsIA5String to ensure that it's ascii characters only? Actually, looking at that function it may not be accurate to the IA5String specification, but seems like maybe that could be improved and used here.

tadukurow: There isn't a requirement for ascii characters only as subject_dn can contain UTF8String as well. Also encodings.IsIA5String includes the characters that would constitute metadata only, such as "-". 
This check only looks for [a-zA-Z0-9] or characters larger than 127, the end of the ascii table.

cardonator: Fair enough. The reason I ask is that this lint is _very_ similar to this lint I pulled from x509lint:

https://github.com/zmap/zlint/blob/master/lints/lint_subject_dn_not_printable_characters.go#L49

The highlighted line being where a nearly identical check is being performed. Would it be worth improving that lint or making isPrintableString (or something of that nature) in the util/encodings.go library for both lints to share?

sleevi: I don't think we want to conflate this lint with any sort of isPrintableString checks. There's a difference between checking that the string is within an appropriate character space and valid for that character space (e.g. valid UTF-8). I also think there's a distinction between checking for control characters / non-printable characters and what the BRs refer to as metadata characters (with specific examples)

The cablint check is https://github.com/awslabs/certlint/blob/master/lib/certlint/cablint.rb#L166 , and AIUI, this is trying to be comparable, and the reasoning seems sound.

cardonator: I basically agree with what you're saying, just asking the question to make sure that we aren't unnecessarily duplicating rune checks. If there is not a good use case for collapsing this functionality, I can be behind that. :)

Also, the BRs do have some specific examples, but it does seem likely that this check would still allow characters that would be known metadata values (such as ∅). However, it's not really possible to know every character that could be considered `absent, incomplete, or not applicable` and I'm all for replicating valid lints from x509lint and cablint here.

cpu: It sounds like everyone is on the same page :+1: 

@cardonator Can I go ahead and merge this branch or is there anything else you would like to see changed/discussed?

cardonator: @cpu I'm good 👍 thanks for checking!

tadukurow:
Thanks @cpu!

PR #289, opened 2019-06-25
Policy qualifiers shouldn't be present if the subordinate CA is an affiliate of the entity that controls the Root CA

hablutzel1:
From BRs, "7.1.2.2. Subordinate CA Certificate":

> The following fields MAY be present if the Subordinate CA is not an Affiliate of the entity that controls the Root CA.
> certificatePolicies:policyQualifiers:policyQualifierId (Optional)
> ...
> certificatePolicies:policyQualifiers:qualifier:cPSuri (Optional)
> ...

It is indicating that policy qualifiers (e.g. CPS Pointer) shouldn't be present at all if the subordinate CA is affiliated to the root CA, isn't it?. In that case, this PR includes a new linter that would emit a notice when there are any policy qualifiers in subordinate CAs to help the subordinate CA signers to double check if they should be really including these policy qualifiers.

Finally, if this lint is OK to be pulled, let me know to write its corresponding test.

### DIFF
--- a/lints/lint_sub_ca_certificate_policies_qualifiers_only_for_non_affiliates.go
+++ b/lints/lint_sub_ca_certificate_policies_qualifiers_only_for_non_affiliates.go
@@ -51,7 +51,7 @@ func (l *subCACertPolicyQualifiersOnlyForNonAffiliates) Execute(c *x509.Certific
 func init() {
 	RegisterLint(&Lint{
 		Name:          "n_sub_ca_certificate_policies_qualifiers_only_for_non_affiliates",
-		Description:   "Policy qualifiers shouldn't be present if the subordinate CA is an Affiliate of the entity that controls the Root CA",
+		Description:   "Policy qualifiers shouldn't be present if the subordinate CA is an affiliate of the entity that controls the Root CA",
 		Citation:      "BRs: 7.1.2.2",
 		Source:        CABFBaselineRequirements,
 		EffectiveDate: util.CABEffectiveDate,
--- a/lints/lint_sub_ca_certificate_policies_qualifiers_only_for_non_affiliates.go
+++ b/lints/lint_sub_ca_certificate_policies_qualifiers_only_for_non_affiliates.go
@@ -1,0 +1,60 @@
+package lints
+
+/*
+ * ZLint Copyright 2019 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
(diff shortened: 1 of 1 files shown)


cpu:
:wave: Hi @hablutzel1,

> It is indicating that policy qualifiers (e.g. CPS Pointer) shouldn't be present at all if the subordinate CA is affiliated to the root CA, isn't it?

I think that's a reasonable reading but I don't know that the notice level lint you're proposing is going to be helpful overall. I suspect since you can't determine if any given sub-CA is "an Affiliate of the entity that controls the Root CA" programmatically the lint will necessarily create false-positive notices.

Does anyone else have an opinion?

cardonator:
I agree that the interpretation is reasonable but the wording in the BR doesn't specify whether it is "MUST NOT" vs. "SHOULD NOT". Since it's not explicitly stated, it's not clear what the correct behavior should be.

Also agree that any lint would be very likely to produce false positives.

hablutzel1:
It is clear that **it will provide false positives** for the subordinate CAs under a third party root (which I think that are not the most), but given that this is just a notice, isn't that acceptable?.

> I think that's a reasonable reading but I don't know that the notice level lint you're proposing is going to be helpful overall.

It would help to remind root CAs signing their own subordinate CAs to not include policy qualifiers, but it seems that several CAs are doing it anyway. Some random examples:

- https://crt.sh/?id=2911299
- https://crt.sh/?id=1044349

> Notice level would be appropriate for this type of informational only lint.

This is actually a Pass or Notice lint.

Regardless of the previous, and even more important, what is the reason for this requirement in the BRs? , do you know?.

zakird:
I agree with @cpu here that I don't think that this is going to be helpful overall. Even as a notice, it's going to cause a fair amount of noise. There's just no way for us to know whether the subCA is affiliated. In general, I'd rather be conservative about adding noise since it will cause CAs to just ignore the notices due to warning fatigue.

cpu:
> what is the reason for this requirement in the BRs? , do you know?.

@hablutzel1 I don't know, sorry. Perhaps you could submit a "question from the public" to `questions@cabforum.org` (from the [CA/B F email lists page](https://cabforum.org/email-lists/)).

cpu:
It seems like between @zakird, @cardonator and myself the consensus is that this won't be a net win for `zlint`. I'm going to close the PR for now to keep the queue tidy. If dissenting opinions come forth we can always re-open it.

PR #290, opened 2019-07-11
ci: use GOFLAGS, enforce gofmt -s.

cpu:
This PR sets up a global `GOFLAGS` to avoid needing to repeat it for each command.

It also changes the script to exit on the first error, and to enforce that all non-vendored `.go` files satisfy `gofmt -s` without diffs. This will make sure all code is consistently formatted and help contributors
do the right thing by default.

(Matches the `zcrypto` PR: https://github.com/zmap/zcrypto/pull/195 - I'll leave it up to you folks if you'd like to add similar checks to the other `zmap` repositories, `zcertificate`, etc)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,13 @@ go:
   - "1.12.x"
 
 script:
+  # Fast-fail on non-zero exit codes
+  - set -e
+  # Build commands
   - make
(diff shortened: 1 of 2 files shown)


cpu:
This is ready for :mag: as of f968b61

Issue #291, opened 2019-07-19
Ambiguity on KeyUsage with ECC public key


jinNakamura:
This issue was discussed in GoogleGroups and Bugzilla as follows:
<https://groups.google.com/forum/#!searchin/mozilla.dev.security.policy/ito%7Csort:date/mozilla.dev.security.policy/vDhKG7T6sCM/gKwMGqrYAwAJ>
<https://bugzilla.mozilla.org/show_bug.cgi?id=1560234>

If the KeyUsage of the certificate with ECC public key contains both digital signature and key encryption, cablint returns as the error when we run cablint.
However, current zlint does not return as the error.
If zlint can return as the error, I think zlint will be a better tool.

cpu:
:wave: Thanks for opening an issue about this @jinNakamura. I spent a little bit of time reading the MDSP, LAMPS/SPASM, and Bugzilla discussion and I _think_ I'm up to speed now.

> If zlint can return as the error, I think zlint will be a better tool.

My understanding of the discussion consensus is that while this EKU/public key algorithm combination is non-sense it isn't clearly forbidden by the relevant RFCs/BRs as they are written today. Based on that conclusion there _could_ be an argument made that a lint for this condition should return a `Warn` level lint finding and not an `Error` level finding until the [RFC 5480 key usage clarification draft](https://datatracker.ietf.org/doc/draft-turner-5480-ku-clarifications/) is finalized.

Personally I'm not opposed to implementing a lint at the `Error` level (matching cablint) ahead of that clarification being finalized but I wanted to call this out specifically so that other zlint community members/relying users could weigh in.

cardonator:
@cpu I agree with your assessment. Unfortunately, there are a number of places in the RFCs and BRs where the ambiguity of the term "MAY" is left up to the interpretation of the CABF and not clearly defined in the documentation. This was also the case with a previous PR that was rejected.

One of the difficult aspects of just flat out re-implementing cablint and x509lint functionality into zlint is that both of those linters have baked in "interpretation-level" errors and warnings. With the current ambiguity, I would argue that `Notice` level would be more appropriate, but I could be convinced toward `Warn` only because draft RFC 5480 could fix this ambiguity when finalized (at which point, `Error` would indeed make the most sense).

[...]

PR #293, opened 2019-07-25
lints: add Notice level lint for EE ECDSA KeyUsages.

cpu:
[RFC 5480 Section 3](https://tools.ietf.org/html/rfc5480#section-3) "Key Usage Bits" indicates that end-entity certificates using a EC public key MAY include the `digitalSignature`, `nonRepudiation`, and `keyAgreement` Key Usages.

If such a certificate contains other Key Usages the new `n_ecdsa_ee_invalid_ku` lint will return a `Notice` level `LintResult` indicating the unexpected Key Usage bits that were included. Depending on the adoption of [a clarification document](https://datatracker.ietf.org/doc/draft-turner-5480-ku-clarifications/) and respective CABF BR updates it may be possible to increase the severity of this `LintResult` in the future. See https://github.com/zmap/zlint/issues/291 for more background discussion.

There are probably similar lints that could be written to add related coverage for other key types and KUs. I only had time to implement this more narrowly scoped lint this week but would be available to review follow-up PRs from other contributors. 

Resolves https://github.com/zmap/zlint/issues/291


cpu:
Thanks for the reviews!

PR #309, opened 2019-08-12
lints: add e_subject_printable_string_badalpha lint.

cpu:
When the raw Subject RDNSequence of a Certificate includes a PrintableString type DirectoryString attribute the value of the attribute must adhere to the PrintableString character set defined in RFC 5280 Appendix B:

>The character string type PrintableString supports a very basic Latin character set: the lowercase letters 'a' through 'z', uppercase letters 'A' through 'Z', the digits '0' through '9', eleven special characters ' = ( ) + , - . / : ? and space.

If any of the PrintableString attributes in the linted Certificate's raw subject do not match a regexp for this character set an Error level lint result is returned by the `e_subject_printable_string_badalpha` lint.

_Adding a lint for this case has been on the back of my mind since an [unfortunate incident affecting an upstart CA back in 2018....](https://groups.google.com/d/msg/mozilla.dev.security.policy/wqySoetqUFM/34W5zvENAwAJ)_


lints/lint_subject_printable_string_badalpha.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/zmap/zlint/util"
+)
+
+var (
+	// Per RFC 5280, Appendix B. ASN.1 Notes:
+	//   The character string type PrintableString supports a very basic Latin
+	//   character set: the lowercase letters 'a' through 'z', uppercase
+	//   letters 'A' through 'Z', the digits '0' through '9', eleven special
+	//   characters ' = ( ) + , - . / : ? and space.
+	printableStringRegex = regexp.MustCompile(`^[a-zA-Z0-9\=\(\)\+,\-.\/:\? ]+$`)
```
tidoublemy: Looks like the comment is correct here but the Regexp is not; it does not include an apostrophe as a valid character.  I noticed a lot of false positives in crt.sh coming from ZLint today that led me here.

zakird: Yep, agreed. I'll submit a PR in a second. Any chance you have a handy link to one of those certificates that failed that I can throw into a unit test?

cpu: Good catch @tidoublemy. I'm available to turn around a quick review and a bugfix release to atone for my regex bug :sweat:

cpu: Fixed in https://github.com/zmap/zlint/releases/tag/v1.0.1

PR #315, opened 2019-09-11
Fixed two bugs in QcEuLimitValue - QC Statement

bilalashraf123:


lints/lint_qcstatem_qclimitvalue_valid_test.go
```diff
@@ -20,7 +20,7 @@ import (
 
 func TestQcStatemQcLimitValueValid(t *testing.T) {
 	m := map[string]LintStatus{
-		"QcStmtInvalidLimitValue.pem": Error,
+		"QcStmtInvalidLimitValue.pem": NA,
```
zakird: Can you explain this change here? The test seems to be to test whether QC Limit Values are valid and this seems to effectively remove that test.

bilalashraf123: The certificate used in this test does not seem to have QcEuLimitValue and hence this test case result is not applicable. If we use certificate that has QcEuLimitValue, then we can write test for either pass or fail.

Previously,  QcEuSSCD OID was used instead of QcEuLimitValue which results in error in every case and the expected result was also error. The overall test result was passed which was wrong IMO.

bilalashraf123: @zakird Is it ok, if i create two new test certs for two test cases containing valid and invalid QcLimitValue? The existing valid test case would be updated with new valid cert and one more test would be introduced for invalid test.

zakird: Sorry I lost track of this issue. Yes, I believe we need to create a new certificate that triggers the logic in this lint. Otherwise we effectively don't have any tests for the lint.

bilalashraf123: Valid and invalid tests are added for QcEuLimitValue

zakird:
Thanks for getting this fixed up and adding the new unit test.

Issue #312, opened 2019-08-26
Add "integration" testing against large cert corpus


cpu:
Suggested by @zakird: To better catch regressions like [this one](https://github.com/zmap/zlint/pull/311) it would be beneficial if `zlint`'s continuous integration included a step that ran all of the `zlint`'s lints across a large (1M? 10M? 100M?) certificate corpus. Any differences in lint results for corpus certificates should be expected, or investigated.

Some initial design thoughts:
* It would be best to keep the corpus out-of-tree to avoid bloat. Probably a `.tar.bz` that could be downloaded on-demand would be best.
* We'll need a build script that can download the corpus, extract it, and process it to generate a results summary.
* We can check the results summary in-tree to use as the gold standard results.
* The CI side should run the build script and fail on diffs to the gold standard results. Intentional changes in corpus lint status should be reflected in commits to the gold standard results in-tree.
* We should adjust the CI config to explicitly cache the unpacked corpus directory to try and keep CI speedy and avoid wasting bandwidth on the corpus host.
* We'll need to do some light benchmarking to figure out the right corpus size. The Travis CI nodes have pretty weak virtual CPUs.

cpu:
Quick update for this issue: I have a big collection of certs ready to go (thanks to @zakird) and plan to start working on the Travis side, hopefully over the next week. The last week or so has been preempted by fire fighting :fire: :fire_engine:

cpu:
I put up a proof-of-concept PR here: https://github.com/zmap/zlint/pull/318 The PR description calls out some of the trade-offs made. I think it strikes a good balance of coverage vs execution time vs in-repository bloat. Reviews/additional input welcome!

cpu:
Resolved by #318. Tracking a proposed follow-up improvement in https://github.com/zmap/zlint/issues/322

PR #318, opened 2019-09-15
tests: add large cert corpus based integration test.

cpu:
## Background

[The idea](https://github.com/zmap/zlint/issues/312) is to run `zlint` across a big corpus of certificates, ideally as diverse a corpus as possible, and compare the results against expected values. The hope is that this will catch regressions in lints that change lint results for the corpus in an unexpected way.

To start with 60 data files (each of ~5000 certificates) are used to lint ~600,000 certificates during each CI run to verify lint results do not change unexpectedly. The tooling was written such that it is easy to run against larger corpuses before doing a big release, or on a less frequent schedule than for day to day CI.

The config in this branch adds ~10 minutes runtime to CI. I verified it would help catch regressions by reverting the fix to the `e_subject_printable_string_badalpha` lint
(`5dcecad773158b82b5e52064ee2782d1b8a79314`). The integration tests successfully flagged a difference that would have focused attention on the new lint: https://gist.github.com/cpu/3a80db08a14a9de7a56db4ff1fc821e9

## Golden state

In a perfect world we would record the full results from each lint for each certificate and make sure they don't change unexpectedly. Recording that much data in a "golden" state file in the repo for a large number of certs would result in pretty significant bloat and slow clones. Putting it out-of-repo would make PRs cumbersome and adds challenges for external contributors.

As a compromise the current integration test only tracks a count of fatal, error, warning, and notice lint results per-certificate but not details about which lint produced each result at that level.

With the one corpus file in this branch this adds ~60mb of golden state data to the repo (see `integration/config.json`) which seems acceptable to me. If we want to trim this down we could stop tracking notice level lint results or use a more efficient serialized representation than JSON. One of the advantages of the bloated JSON form is that it is human readable and easy to update in an editor.

To produce the golden state for the first time or for a large update you can edit the `integration/config.json` to delete the `"Expected"` map. When the test is run and there is no configured `Expected` map then the current results are saved as the `Expected` map to be validated against in subsequent runs.

## Corpus

The cert corpus data is quite large so the integration tests are written to keep the data out of tree and only download it when needed. Support is included to automatically decompress the downloaded data using bzip2 when the URL ends in the `.bz2` file extension.

The Travis CI config is updated to explicitly cache the downloaded corpus data between builds to help avoid needing to download it every build. Similarly some effort was made to process data in parallel to keep builds as fast as possible while still linting as many certs as we can. Local testing on good hardware can be done even faster by increasing the parallelism over what is used in CI where the worker machines are weak and often under-provisioned.

To start with 60 corpus files (each ~5000 certs) are configured for day to day CI. These files (and more) live in the https://github.com/zmap/zlint-test-corpus repository.

## Running Integration Tests

By default when running `go test` (or `make test`) the integration tests are not built or run. You must provide `-tag=integration` to build/run these tests. This should keep day to day development quick.

The `makefile` is updated with a `make integration` target to run just the integration tests and the Travis script uses that to run integration tests. The `PARALLELISM` and `INT_FLAGS` makefile variables can be used to change the number of Go routines and additional flags (see Extra Command Line Flags) without modifying the Makefile.

## Extra Command Line Flags

To use an alternative configuration specify the `-config` command line flag. By default `integration/config.json` is used.

To control the number of linting Go routines used by the `TestCorpus` integration test change the `-parallelism` command line flag when running the integration tests. By default 5 Go routines are used for local integration tests and 3 for Travis CI.

To force corpus data to be downloaded even when it exists on disk use the `-force` command line flag when running the integration tests. By default `-force` is disabled.

To have a summary of certificate fingerprint and integration test results at the end of the integration test provide the `-summary` command line flag. Note that this is quite spammy and is disabled by default

To change how many certificates are linted before a '.' character is printed to the screen use the `-outputTick` flag. By default one period is printed per 1000 certificates to keep Travis from deciding the CI job is dead.


zakird:
Overall this sounds great. I think that a good option for file hosting would be to split the corpus up into a few 50mb files and put those in their own Github repo under ZMap org. That way we could grant maintainers access easily and wouldn't need any third party dependency.

cpu:
>  I think that a good option for file hosting would be to split the corpus up into a few 50mb files and put those in their own Github repo under ZMap org. That way we could grant maintainers access easily and wouldn't need any third party dependency.

That's a really good idea :+1: Do you want to do that ahead of merging this PR and I can update the `integration/config.json` to point at the Github URLS?

zakird:
I created the repo. I'm going to re-run script to dump out half million certificates per your suggestion above.

cpu:
>  I can update the integration/config.json to point at the Github URLs

I started working on this and pushed up a config that lints ~800,000 certs fetched from the Github repo. The resulting test metadata that needs to be checked into the zlint repo ends up being ~80mb which is _maybe_ too big but we're going to run with it for now. There's lots of room for optimization.

I think the CI runtime that results from the config in f4b64d0 is too long (over 20m on first build) so I'm going to push at least one more round of commits onto this branch to try and find a happier middle ground. I recommend we hold off on merging this PR until I've had a chance to do that, hopefully tomorrow.

cpu:
>  I'm going to push at least one more round of commits onto this branch to try and find a happier middle ground. I recommend we hold off on merging this PR until I've had a chance to do that, hopefully tomorrow. I'll also fix up the PR description at that point.

Both done as of my latest force push. I switched the config to 60 data files from the Github repo. The in-tree `config.json` is 62mb and the CI time is ~10-15m end to end.

@zakird I think this is ready for :mag: when you have a chance.

cpu:
@zakird It looks like #315 changed the profile of expected test output for this branch for 10 certificates. Does someone more familiar with the changes in #315 want to vet these or should I assume the changed output is correct and update this branch accordingly? 


```

--- FAIL: TestCorpus (624.61s)
    corpus_test.go:114: linted 599997 certificates
    corpus_test.go:138: expected serial "cddedac732c224ef3ef89e59b40982049358a1f0f38adf80c003f1f1c7a406d0" to have result fatals: 0	errs: 8	warns: 2	infos: 1 got fatals: 0	errs: 7	warns: 2	infos: 1
    corpus_test.go:138: expected serial "d9c2bcb81a49ed296437a5649b6c35440afa261df6c2fce66c19a1f89c806b48" to have result fatals: 0	errs: 7	warns: 2	infos: 1 got fatals: 0	errs: 6	warns: 2	infos: 1
    corpus_test.go:138: expected serial "4984a5a0b51edd01a3e09c6954b4b982e4d8cd3985d153bae06ee4cf4172ff0e" to have result fatals: 0	errs: 8	warns: 2	infos: 1 got fatals: 0	errs: 7	warns: 2	infos: 1
    corpus_test.go:138: expected serial "0692900c09f7b127e757ee7172d05d19a1301a8b9695d5f9a325bcaddfd340c7" to have result fatals: 0	errs: 1	warns: 3	infos: 0 got fatals: 0	errs: 0	warns: 3	infos: 0
    corpus_test.go:138: expected serial "225295a544a767cc26d96b90ac3b6ffffe69461dec0cf97b6951c94b3f3b773f" to have result fatals: 0	errs: 8	warns: 2	infos: 2 got fatals: 0	errs: 7	warns: 2	infos: 2
    corpus_test.go:138: expected serial "0af263c16e17a9687eb761dbab33ea57b1d903a7f559bb07e0f805086157c82e" to have result fatals: 0	errs: 8	warns: 2	infos: 1 got fatals: 0	errs: 7	warns: 2	infos: 1
    corpus_test.go:138: expected serial "35a4372435944414797f8d7195a1aef9c16bf963fe3899fb32797fdd5e51c133" to have result fatals: 0	errs: 1	warns: 3	infos: 0 got fatals: 0	errs: 0	warns: 3	infos: 0
    corpus_test.go:138: expected serial "3601b11e80b5137cd0d51f747c81411952f9f190da2534fbdc0e89d35ba9dd4e" to have result fatals: 0	errs: 1	warns: 3	infos: 0 got fatals: 0	errs: 0	warns: 3	infos: 0
    corpus_test.go:138: expected serial "85bc25cc034df0528dc2caa146bc8f9b8eabc7fe22dc28f8031feb4f1a7d8009" to have result fatals: 0	errs: 8	warns: 1	infos: 1 got fatals: 0	errs: 7	warns: 1	infos: 1
    corpus_test.go:138: expected serial "a17f644b52119b96e722b769d3585808fb90b3d016bb5478e7e7bbd236970252" to have result fatals: 0	errs: 1	warns: 1	infos: 0 got fatals: 0	errs: 0	warns: 1	infos: 0
```

sleevi:
@cpu All of those certificates assert the oid `0.4.0.1862.1.4` in their QcStatements. #315 fixed a bug where ZLint was incorrectly checking `0.4.0.1862.1.4` (which is related to it being on a secure signature creation device - QcSSCD) instead of `0.4.0.1862.1.2` - which was the correct OID.

So I think that's all WAI? If you're looking for deltas with #315 , you could look at https://crt.sh/?id=1723124144 or https://crt.sh/?id=1635981809 - both of which misencoded the limits value.

cpu:
> I think that's all WAI?

@sleevi Thanks for the analysis. Much appreciated. I'll update the source of truth in this branch accordingly.

> I'm not certain that we should be storing values based on certificates FP versus lint name. Lint name seems like it would use up a lot of less space on disk. This may come down to a usability question: when we see a test fail, which case do we want to investigate by?

@zakird I'm open to changing how the ground truth/golden state is arranged to do it this way. The comments from @sleevi also made me think that a small change that would improve usability would be adding `crt.sh` links to the failure results.

I expect I won't have a chance to make structural changes to this code until mid/late next week at the earliest. Do folks prefer to land this branch-as is and have that be a follow-up PR or would it be preferable to block on a tweaked design?

zakird:
I'd say let's merge this for now and we can add these pieces later, though I'm really fine with either approach.

cpu:
> I'd say let's merge this for now and we can add these pieces later

WFM :ballot_box_with_check: I think it makes sense to get the benefits of improved test coverage while we iterate further.

PR #323, opened 2019-10-02
Lints: add new lints for Mozilla Root Store Policy.

fotisl:
Added four new lints per Mozilla Root Store Policy:
- Intermediate certificates created after January 1, 2019, with the exception of cross-certificates that share a private key with a corresponding root certificate: MUST contain an EKU extension; and, MUST NOT include the anyExtendedKeyUsage KeyPurposeId; and, * MUST NOT include both the id-kp-serverAuth and id-kp-emailProtection KeyPurposeIds in the same certificate.
Note: the lint cannot at the moment figure out if a certificate is a cross-certificate, so it's linted under this rule.
- RSA keys whose modulus size in bits is divisible by 8, and is at least 2048. CAs MUST NOT issue certificates that have invalid public keys (e.g., RSA certificates with public exponent equal to 1).
- CAs MUST NOT issue certificates that have incorrect extensions (e.g., SSL certificates that exclude SSL usage, or authority key IDs that include both the key ID and the issuer’s issuer name and serial number). The lint checks for authority key identifiers that include both the key id and the issuer's issuer name and serial.
- ECDSA keys using one of the following curve-hash pairs: P‐256 with SHA-256, P‐384 with SHA-384
Note: there is no way to figure out the size of the key from the signature when using DER encoding, and a heuristic algorithm is provided which should work with most prime fields.


robstradling:
@fotisl: Thanks for working on this!

> Note: the lint cannot at the moment figure out if a certificate is a cross-certificate, so it's linted under this rule.

Shouldn't it be a "warning" or a "notice" then, rather than an "error"?

fotisl:
I can definitely change this to a warning. Another idea that can be implemented is to keep a list of all spkis for roots in the mozilla trust store in order to detect the cross-certificates, but this would require maintenance. What's your take on this @zakird ?

lints/lint_mp_allowed_rsa_keys.go
```diff
@@ -0,0 +1,67 @@
...
+		return &LintResult{Status: Error}
+	}
+
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_mp_allowed_rsa_keys",
+		Description:   "RSA keys whose modulus size in bits is divisible by 8, is at least 2048 and public exponent is not equal to 1",
```
zakird: This lint tests two things, which while relate to RSA keys, are fairly separate: (1) RSA keys whose modulus size in bits is divisible by 8, and is at least 2048; and (2) RSA certificates with public exponent equal to 1. I would recommend splitting these up into two lints. Otherwise, when this fails, it's not clear what the problem is.

zakird: Generally, description should contain the exact text from the standards requirement that dictates the lint. This ensures that the code exactly measures the statement. For example, `RSA keys whose modulus size in bits is divisible by 8, and is at least 2048.`.

lints/lint_mp_allowed_eku.go
```diff
@@ -0,0 +1,74 @@
...
+		(hasEKU(c, x509.ExtKeyUsageEmailProtection) && hasEKU(c, x509.ExtKeyUsageServerAuth)) {
+		return &LintResult{Status: Error}
+	}
+
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_mp_allowed_eku",
```
zakird: We're generally pretty conservative about what can trigger an error. We should be certain that there is an error present as many CAs will block issuance when they encounter an error. As such, I think that this lint can be at max an informational.

lints/lint_mp_allowed_rsa_keys.go
fotisl: I will split this into two different lints. My original thought was to create an RSA key lint, but I agree that since there are multiple "sublints" and so it's not possible to figure out what failed.

fotisl: I edited this a little bit since it tested for multiple problems, will fix it when splitting the lint.

lints/lint_mp_allowed_eku.go
fotisl: I could create a small script that automatically fetches all trusted roots from mozilla (http://mxr.mozilla.org/mozilla/source/security/nss/lib/ckfw/builtins/certdata.txt) and creates a valid go file with the spkis. This would allow us to identify _cross-certificates that share a private key with a corresponding root certificate_. This approach would help block issuance of subs and at the same time allow cross-certificates to be signed. However, it requires manual execution of the script when the mozilla trust store is updated.

lints/lint_mp_allowed_eku.go
```diff
@@ -0,0 +1,74 @@
...
+package lints
+
+import (
+	"time"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/util"
+)
+
+func hasEKU(cert *x509.Certificate, eku x509.ExtKeyUsage) bool {
```
cpu: I see a few other places that would benefit from this function. Can you promote it to the `util` package?

lints/lint_mp_allowed_eku.go
```diff
@@ -0,0 +1,74 @@
...
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_mp_allowed_eku",
+		Description:   "Separation of id-kp-serverAuth and id-kp-emailProtection KeyPurposeIds",
+		Citation:      "Mozilla Root Store Policy / Section 5.3",
+		Source:        MozillaRootStorePolicy,
+		EffectiveDate: time.Date(2019, time.January, 1, 0, 0, 0, 0, time.UTC),
```
cpu: Was there a reason to avoid adding an exported date to `util` for this alongside `MozillaPolicy22Date` and `MozillaPolicy24Date`?

lints/lint_mp_allowed_rsa_keys.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type allowedRSAKey struct{}
+
+func (l *allowedRSAKey) Initialize() error {
+	return nil
+}
+
+func (l *allowedRSAKey) CheckApplies(c *x509.Certificate) bool {
+	switch c.SignatureAlgorithm {
```
cpu: style nit: I think a `switch` with only one case is more idiomatically expressed as an `if` statement.

lints/lint_mp_allowed_rsa_keys_test.go
```diff
@@ -0,0 +1,54 @@
...
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import (
+	"fmt"
+	"testing"
+)
+
+func TestSerialEntropyBits(t *testing.T) {
```
cpu: I think the name of this test function doesn't match what it's actually testing.

lints/lint_mp_allowed_rsa_keys.go
```diff
@@ -0,0 +1,67 @@
...
+	case x509.SHA1WithRSA, x509.SHA256WithRSA, x509.SHA384WithRSA, x509.SHA512WithRSA, x509.SHA256WithRSAPSS, x509.SHA384WithRSAPSS, x509.SHA512WithRSAPSS:
+		return true
+	}
+
+	return false
+}
+
+func (l *allowedRSAKey) Execute(c *x509.Certificate) *LintResult {
+	pubKey := c.PublicKey.(*rsa.PublicKey)
```
cpu: For safety's sake I think it makes sense to handle a failed cast of the `c.PublicKey` as a `Fatal` level `LintResult`. In theory `CheckApplies` will prevent `Execute` from running into this but it defense in depth seems like the right call to me.
```suggestion
	pubKey, ok := c.PublicKey.(*rsa.PublicKey)
	if !ok  {
	  return &LintResult{Status: Fatal, Details: "certificate public key was not an RSA public key"}
	}
```

lints/lint_mp_authority_key_identifier_correct.go
```diff
@@ -0,0 +1,79 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+/********************************************************************
+Section 5.2 - Forbidden and Required Practices
+CAs MUST NOT issue certificates that have:
+- incorrect extensions (e.g., SSL certificates that exclude SSL usage, or authority key IDs
+  that include both the key ID and the issuer’s issuer name and serial number); or
```
cpu: tiny nit: cutting off the trailing "or" seems more natural for this context.

```suggestion
  that include both the key ID and the issuer’s issuer name and serial number);
```

lints/lint_mp_authority_key_identifier_correct.go
```diff
@@ -0,0 +1,79 @@
...
+		return false
+	}
+	return true
+}
+
+func (l *authorityKeyIdentifierCorrect) Execute(c *x509.Certificate) *LintResult {
+	var keyID keyIdentifier
+
+	ext := util.GetExtFromCert(c, util.AuthkeyOID)
+
```
cpu: style nit: superfluous whitespace
```suggestion
```

lints/lint_mp_authority_key_identifier_correct.go
```diff
@@ -0,0 +1,79 @@
...
+	if ext == nil {
+		return &LintResult{Status: Fatal}
```
cpu: Can you add a `Details` string here that indicates what the problem was? (e.g. something about the certificate missing the expected AuthKeyOID extension)

lints/lint_mp_authority_key_identifier_correct.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	if _, err := asn1.Unmarshal(ext.Value, &keyID); err != nil {
```
cpu: Should `rest` be captured and checked here?

lints/lint_mp_authority_key_identifier_correct.go
```diff
@@ -0,0 +1,79 @@
...
+		return &LintResult{Status: Fatal}
```
cpu: Ditto here RE: providing a `Details` string for the `Fatal` result.

lints/lint_mp_ecdsa_allowed_algorithm.go
```diff
@@ -0,0 +1,112 @@
...
+		return true
+	}
+
+	return false
+}
+
+func (l *ecdsaAllowedAlgorithm) Execute(c *x509.Certificate) *LintResult {
+	signKeySize, err := getSigningKeySize(c)
+	if err != nil {
+		return &LintResult{Status: Fatal}
```
cpu: Ditto here RE: `Details` for `Fatal`.

lints/lint_mp_ecdsa_allowed_algorithm.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+type ecdsaAllowedAlgorithm struct{}
+
+func (l *ecdsaAllowedAlgorithm) Initialize() error {
+	return nil
+}
+
+func (l *ecdsaAllowedAlgorithm) CheckApplies(c *x509.Certificate) bool {
+	switch c.SignatureAlgorithm {
```
cpu: ditto here on the minor style nit about using an `if` here instead of a single `case` `switch`.

lints/lint_mp_ecdsa_allowed_algorithm.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	switch {
+	case c.SignatureAlgorithm == x509.ECDSAWithSHA256 && signKeySize == 256:
+		return &LintResult{Status: Pass}
+	case c.SignatureAlgorithm == x509.ECDSAWithSHA384 && signKeySize == 384:
+		return &LintResult{Status: Pass}
+	}
+
+	return &LintResult{Status: Error}
```
cpu: Can you add a `Details` message here that indicates what signature algorithm/sign key size *was* found?

lints/lint_mp_allowed_rsa_keys.go
zakird: This doesn't seem correct. The signature on the certificates does not determine the type of public key it uses.

lints/lint_mp_allowed_rsa_keys_test.go
zakird: This name doesn't look appropriate for the test.

lints/lint_mp_allowed_rsa_keys.go
fotisl: Yes, you are correct, an RSA CA with an ECDSA cert will fail, I am also going to correct this.

lints/lint_mp_allowed_rsa_keys_test.go
fotisl: Wow, copy/paste error, should have reviewed it.

lints/lint_mp_allowed_rsa_keys.go
cpu: I feel a little bit vindicated asking for this seeing that [CI has failed this branch](https://travis-ci.org/zmap/zlint/builds/592504764) because of a panic from this cast failing during the new integration tests:
```
2019/10/02 11:11:32 Reading data from "../data/xaa.csv"
panic: interface conversion: interface {} is *x509.AugmentedECDSA, not *rsa.PublicKey
goroutine 26 [running]:
github.com/zmap/zlint/lints.(*allowedRSAKey).Execute(0xbe27a8, 0xc000578400, 0xc000225701)
	/home/travis/gopath/src/github.com/zmap/zlint/lints/lint_mp_allowed_rsa_keys.go:48 +0x11e
github.com/zmap/zlint/lints.(*Lint).Execute(0xc00021f620, 0xc000578400, 0x849e39)
	/home/travis/gopath/src/github.com/zmap/zlint/lints/base.go:117 +0x9e
github.com/zmap/zlint.(*ResultSet).execute(0xc000dee420, 0xc000578400)
	/home/travis/gopath/src/github.com/zmap/zlint/zlint.go:44 +0x112
github.com/zmap/zlint.LintCertificate(0xc000578400, 0xc000d85650)
	/home/travis/gopath/src/github.com/zmap/zlint/zlint.go:83 +0x5c
github.com/zmap/zlint/integration.lintCertificate(0xc009a654af, 0x40, 0xc000578400, 0xc009bfb777, 0x40, 0x1000000)
	/home/travis/gopath/src/github.com/zmap/zlint/integration/corpus_test.go:23 +0x40
github.com/zmap/zlint/integration.TestCorpus.func2(0xc0000b2240, 0xc0000b2300, 0xc0096a6c10)
	/home/travis/gopath/src/github.com/zmap/zlint/integration/corpus_test.go:71 +0x6a
created by github.com/zmap/zlint/integration.TestCorpus
	/home/travis/gopath/src/github.com/zmap/zlint/integration/corpus_test.go:68 +0x1c6
FAIL	github.com/zmap/zlint/integration	1.803s
FAIL
```

lints/lint_mp_allowed_eku.go
fotisl: Definitely, will move it there.

lints/lint_mp_allowed_eku.go
fotisl: This is a date that's only used for this requirement. The requirement is part of the Mozilla Root Store Policy 2.4 but the enforcement date is different that the rest of the requirements. Since it doesn't apply to anything else, I thought it would be best if I don't use an exported date. What do you think?

cpu: That makes sense to me. I think it's fine to leave as-is. Thanks for explaining :+1:

lints/lint_mp_allowed_rsa_keys.go
fotisl: Well, TBH all these come from another linter I wrote, and this used to be part of a larger switch statement. However, this is going to be removed as a check.

lints/lint_mp_allowed_rsa_keys_test.go
fotisl: Please see Zakir's comment on this, I am going to correct it.

lints/lint_mp_allowed_rsa_keys.go
fotisl: The test failed because of the SignatureAlgorithm check. The new check will be implemented using the code you provided, and so it won't be needed at Execute().

lints/lint_mp_authority_key_identifier_correct.go
fotisl: I can add it, but this would not be part of the Mozilla Root Store Policy. Maybe a different lint checking compliance of the extension with RFC5280?

lints/lint_mp_authority_key_identifier_correct.go
fotisl: Ditto :) Maybe create a different lint?

lints/lint_mp_ecdsa_allowed_algorithm.go
fotisl: Will do.

lints/lint_mp_ecdsa_allowed_algorithm.go
fotisl: Will do.

lints/lint_mp_ecdsa_allowed_algorithm.go
fotisl: Will do.

lints/lint_mp_authority_key_identifier_correct.go
cpu: Sounds reasonable :+1:

lints/lint_mp_authority_key_identifier_correct.go
cpu: I don't think a different lint makes sense here. Your lint is already returning a `&LintResult{Status: Fatal}` for this case, I'm just advocating for always returning a `Details` string when the `Status` is `Fatal` so that the context of the fatal error is presented to the user.

lints/lint_mp_allowed_rsa_keys_exponent.go
```diff
@@ -0,0 +1,66 @@
...
+		return true
+	}
+
+	return false
+}
+
+func (l *allowedRSAKeyExponent) Execute(c *x509.Certificate) *LintResult {
+	pubKey, ok := c.PublicKey.(*rsa.PublicKey)
+	if !ok {
+		return &LintResult{Status: Fatal, Details: "certificate public key was not an RSA public key"}
```
zakird: I don't think this details is helpful. It's less detailed than the lint name.

lints/lint_mp_allowed_rsa_keys_exponent.go
```diff
@@ -0,0 +1,66 @@
...
+	if pubKey.E == 1 {
+		return &LintResult{Status: Error}
+	}
+
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_mp_allowed_rsa_keys_exponent",
```
zakird: Let's be specific here: `e_mp_rsa_exponent cannot be one` since that's what the lint is testing.

lints/lint_mp_allowed_rsa_keys_modulus.go
```diff
@@ -0,0 +1,66 @@
...
+		return true
+	}
+
+	return false
+}
+
+func (l *allowedRSAKeyModulus) Execute(c *x509.Certificate) *LintResult {
+	pubKey, ok := c.PublicKey.(*rsa.PublicKey)
+	if !ok {
+		return &LintResult{Status: Fatal, Details: "certificate public key was not an RSA public key"}
```
zakird: Can you be more specific error here? Details should really only be included if they include more information than the name of the lint. Keys that fail this test are indeed valid RSA public keys.

lints/lint_mp_allowed_rsa_keys_modulus.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+
+	bitLen := pubKey.N.BitLen()
+	if (bitLen%8) != 0 || bitLen < 2048 {
```
zakird: These probably belong in separate lints. There are two tests in this lint.

lints/lint_mp_ecdsa_allowed_algorithm.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+	}
+
+	return &LintResult{Status: Error}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_mp_ecdsa_allowed_algorithm",
+		Description:   "ECDSA keys using one of the following curve-hash pairs: P‐256 with SHA-256, P‐384 with SHA-384",
```
zakird: This name doesn't match the lint logic. The lint is checking the signature, but the description describes the key present in the certificate.

lints/lint_mp_allowed_rsa_keys_exponent.go
fotisl: It's actually a fatal error for something that should never happen (there is an RSA test at CheckApplies()). Can you propose something else?

lints/lint_mp_allowed_rsa_keys_exponent.go
fotisl: Great, going to rename this to e_mp_exponent_cannot_be_one, and do something similar for the modulus tests.

lints/lint_mp_allowed_rsa_keys_modulus.go
fotisl: Please check reply above, can you propose something?

lints/lint_mp_allowed_rsa_keys_modulus.go
fotisl: Following the exponent lint naming convention, I will split this one to `e_mp_modulus_must_be_divisible_by_8` and `e_mp_modulus_must_be_2048_bits_or_more`, what do you think?

lints/lint_mp_ecdsa_allowed_algorithm.go
fotisl: The exact test from the policy is:

> ECDSA keys using one of the following curve-hash pairs:
> - P‐256 with SHA-256
> - P‐384 with SHA-384

I used the text as it appears, which I agree is a little bit confusing. When the original description is not exact, is it preferable to follow the text from the requirements or adopt it to actually match the lint?

zakird: My concern is that the policy is describing how keys in a certificate should be formed whereas the lint is checking signature on the certificate not the public key.

fotisl: How about changing this to "[CA] ECDSA keys [signing certificates] using one of the following curve-hash pairs: P‐256 with SHA-256, P‐384 with SHA-384"? I think that this both describes the actual lint and provides the policy requirement.

zakird: I don't fully understand the change. That seems to be adding words to the description of the lint to match the logic rather than changing the logic to follow the actual rules. Is there additional context in the Mozilla rules that make this distinction clear that you can provide?

fotisl: Oh, sorry, I didn't understand your original point. The logic of the lint is the actual requirement, however it is not stated well in the policy. There are CAs that have been confused about this too, please check https://bugzilla.mozilla.org/show_bug.cgi?id=1530971 as an example which lead to the following commit that will be integrated in the next mozilla policy version: https://github.com/sleevi/pkipolicy/commit/cc6355ee1fb32500b496d38ffff8abb98e90b3db.

lints/lint_mp_allowed_rsa_keys_exponent.go
cpu: I think this `Details` string is suitable for this case (and ditto for the other comment on L46).

lints/lint_mp_allowed_rsa_keys_modulus.go
zakird: I didn't realize this was a Fatal, I agree this looks fine. Apologies for missing that.

lints/lint_mp_allowed_eku.go
cpu: @fotisl It looks like you landed on implementing a command to do this in this branch. I'd personally prefer to see the lint implemented to be informational to start and have the`cmd/zlint-mozilla-trusted-roots-update` implementation included in a separate follow-up PR. @zakird Do you have a pref here?

lints/lint_mp_ecdsa_allowed_algorithm.go
cpu: I think this feedback is still unresolved.

lints/lint_mp_modulus_must_be_2048_bits_or_more.go
```diff
@@ -0,0 +1,66 @@
...
+		return &LintResult{Status: Error}
+	}
+
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_mp_modulus_must_be_2048_bits_or_more",
+		Description:   "RSA keys whose modulus size in bits is divisible by 8, is at least 2048 and public exponent is not equal to 1",
```
cpu:
```suggestion
		Description:   "RSA keys must have modulus size of at least 2048 bits",
```

The != 1 and divisible by 8 portions of this description are now separate lints.

lints/lint_mp_modulus_must_be_2048_bits_or_more.go
```diff
@@ -0,0 +1,66 @@
...
+func (l *modulus2048OrMore) Initialize() error {
+	return nil
+}
+
+func (l *modulus2048OrMore) CheckApplies(c *x509.Certificate) bool {
+	if c.PublicKeyAlgorithm == x509.RSA {
+		return true
+	}
+
+	return false
```
cpu:
```suggestion
	return c.PublicKeyAlgorithm == x509.RSA
```

lints/lint_mp_exponent_cannot_be_one.go
```diff
@@ -0,0 +1,66 @@
...
+func (l *exponentCannotBeOne) Initialize() error {
+	return nil
+}
+
+func (l *exponentCannotBeOne) CheckApplies(c *x509.Certificate) bool {
+	if c.PublicKeyAlgorithm == x509.RSA {
+		return true
+	}
+
+	return false
```
cpu:
```suggestion
	return c.PublicKeyAlgorithm == x509.RSA
```

lints/lint_mp_modulus_must_be_2048_bits_or_more.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (l *modulus2048OrMore) Execute(c *x509.Certificate) *LintResult {
+	pubKey, ok := c.PublicKey.(*rsa.PublicKey)
+	if !ok {
+		return &LintResult{Status: Fatal, Details: "certificate public key was not an RSA public key"}
+	}
+
+	bitLen := pubKey.N.BitLen()
+	if bitLen < 2048 {
```
cpu:
```suggestion
	if bitLen:= pubKey.N.BitLen(); bitLen < 2048 {
```

lints/lint_mp_modulus_must_be_divisible_by_8.go
```diff
@@ -0,0 +1,66 @@
...
+func (l *modulusDivisibleBy8) Initialize() error {
+	return nil
+}
+
+func (l *modulusDivisibleBy8) CheckApplies(c *x509.Certificate) bool {
+	if c.PublicKeyAlgorithm == x509.RSA {
+		return true
+	}
+
+	return false
```
cpu:
```suggestion
	return c.PublicKeyAlgorithm == x509.RSA
```

lints/lint_mp_modulus_must_be_divisible_by_8.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func (l *modulusDivisibleBy8) Execute(c *x509.Certificate) *LintResult {
+	pubKey, ok := c.PublicKey.(*rsa.PublicKey)
+	if !ok {
+		return &LintResult{Status: Fatal, Details: "certificate public key was not an RSA public key"}
+	}
+
+	bitLen := pubKey.N.BitLen()
+	if (bitLen % 8) != 0 {
```
cpu:
```suggestion
	if bitLen := pubKey.N.BitLen(); (bitLen % 8) != 0 {
```

lints/lint_mp_modulus_must_be_divisible_by_8.go
```diff
@@ -0,0 +1,66 @@
...
+		return &LintResult{Status: Error}
+	}
+
+	return &LintResult{Status: Pass}
+}
+
+func init() {
+	RegisterLint(&Lint{
+		Name:          "e_mp_modulus_must_be_divisible_by_8",
+		Description:   "RSA keys whose modulus size in bits is divisible by 8, is at least 2048 and public exponent is not equal to 1",
```
cpu:
```suggestion
		Description:   "RSA keys must have a modulus size divisible by 8",
```

The != 1 and > 2048 bits portions of this description are now separate lints.

lints/lint_mp_exponent_cannot_be_one.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type exponentCannotBeOne struct{}
+
+func (l *exponentCannotBeOne) Initialize() error {
+	return nil
+}
+
+func (l *exponentCannotBeOne) CheckApplies(c *x509.Certificate) bool {
+	if c.PublicKeyAlgorithm == x509.RSA {
```
tadukurow: This can be removed

lints/lint_mp_modulus_must_be_2048_bits_or_more.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type modulus2048OrMore struct{}
+
+func (l *modulus2048OrMore) Initialize() error {
+	return nil
+}
+
+func (l *modulus2048OrMore) CheckApplies(c *x509.Certificate) bool {
+	if c.PublicKeyAlgorithm == x509.RSA {
```
tadukurow: This can be removed

lints/lint_mp_modulus_must_be_divisible_by_8.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type modulusDivisibleBy8 struct{}
+
+func (l *modulusDivisibleBy8) Initialize() error {
+	return nil
+}
+
+func (l *modulusDivisibleBy8) CheckApplies(c *x509.Certificate) bool {
+	if c.PublicKeyAlgorithm == x509.RSA {
```
tadukurow: This can be removed

cmd/zlint-mozilla-trusted-roots-update/main.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code fetching data "+
+			"from %q : expected status %d got %d",
+			MOZILLA_CERTDATA, http.StatusOK, resp.StatusCode)
+	}
+
+	var trustedCertificates []*x509.Certificate
```
tadukurow: The scanning logic repeats some steps, nests quite deep and doesn't check errors

Alternative could be:

```
var trustedCertificates []*x509.Certificate
	scanner := bufio.NewScanner(resp.Body)
	var readLine bool
	var certdata []byte
	for scanner.Scan() {
		line := scanner.Text()
		line = strings.ReplaceAll(line, "\r\n", "")
		
		if line == "CKA_VALUE MULTILINE_OCTAL" {
			readLine = true
			continue
		}else if line == "END"{
			readLine = false
			continue
		}else if !readLine {
			continue
		}
		
		bs := strings.Split(line, "\\")
		for _, b := range bs {
			if b == "" {
				continue
			}
			i, err := strconv.ParseInt(b, 8, 0)
			if err != nil {
				return nil, fmt.Errorf(
					"unexpected octal character : %s", err)
			}
			certdata = append(certdata, byte(i))
		}	
	}

        if err:= scanner.Err(); err != nil {
		return nil, fmt.Errorf("error reading %q: %s",MOZILLA_CERTDATA,err)
	}

	cert, err := x509.ParseCertificate(certdata)
	if err != nil {
		return nil, fmt.Errorf("unable to parse certificate : %s", err)
	}

	trustedCertificates = append(trustedCertificates, cert)
```

lints/lint_mp_authority_key_identifier_correct.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+type authorityKeyIdentifierCorrect struct{}
+
+func (l *authorityKeyIdentifierCorrect) Initialize() error {
+	return nil
+}
+
+func (l *authorityKeyIdentifierCorrect) CheckApplies(c *x509.Certificate) bool {
+	if !util.IsExtInCert(c, util.AuthkeyOID) {
```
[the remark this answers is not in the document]
fotisl: Thanks! Fixing this.

lints/lint_mp_allowed_eku.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+type allowedEKU struct{}
+
+func (l *allowedEKU) Initialize() error {
+	return nil
+}
+
+func (l *allowedEKU) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubCA(c) && !util.IsSPKIMozillaTrusted(c)
```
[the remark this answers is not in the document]
fotisl: No problem with this, @zakird @cpu can you comment?

lints/lint_mp_exponent_cannot_be_one.go
fotisl: This was due to an incorrect commit directly from the github ui, fixed it.

lints/lint_mp_modulus_must_be_2048_bits_or_more.go
fotisl: This was due to an incorrect commit directly from the github ui, fixed it.

lints/lint_mp_modulus_must_be_divisible_by_8.go
fotisl: This was due to an incorrect commit directly from the github ui, fixed it.

cmd/zlint-mozilla-trusted-roots-update/main.go
fotisl: Well, it's just an extra level of nesting which repeats the reading of a line. Not sure that it's worth the change since IMHO the current structure is easier to read.

lints/lint_mp_ecdsa_allowed_algorithm.go
fotisl: If you check the discussion above, the checks were initially in a single lint but per zlint's structure it's best if they are split in different files. Plus, the check you are proposing is not actually making sure that P-256 keys sign certs with ECDSA with SHA256 and P-384 keys sign certs with ECDSA with SHA384. The lint you are proposing is a new one that will apply once 2.7 enters into effect and I believe it should be a new unique lint. The current lint will become more or less a check for the last part of the new 5.1.2.

tadukurow: It's indeed a new lint(s), but it is relevant before 2.7 becomes live as the language in the current policy also excludes the use of non RSA or P-256/P-384 for public keys, regardless of what the signing key and algorithm was. Don't feel particularly strongly about one or more check per lint though.

lints/lint_mp_allowed_eku.go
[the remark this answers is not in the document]
cpu: That sounds fine to me :+1: I agree it communicates the high level purpose more clearly. The fact that it uses SPKI to make the determination could be called out in the `IsInMozillaRootStore` Godoc comment as a detail of the implementation.

cmd/zlint-mozilla-trusted-roots-update/main.go
cpu: Echoing my [earlier comment](https://github.com/zmap/zlint/pull/323#discussion_r337074333) I would still prefer to see this utility and the review of it be separated into a second PR if possible.

fotisl: I have already removed it from this PR. As soon as it gets accepted, I will open a new one.

cpu: D'oh! Sorry @fotisl, I missed that last commit. Thank you!

lints/lint_mp_allowed_eku.go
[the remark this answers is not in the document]
fotisl: Change committed.

cpu:
@fotisl Apologies. I've totally lost track of the state of this PR w.r.t review feedback. Is it ready for re-review now? Would you be able to rebase against master and make sure the CI tests are passing?

fotisl:
Yes, I think that the PR is ready for review. I'm pretty sure I can rebase it, however I can do nothing regarding the CI tests since they are failing because some certificates being tested are actually failing a new lint!

cpu:
> Yes, I think that the PR is ready for review

Great thanks! I'll take a look.

> I'm pretty sure I can rebase it, however I can do nothing regarding the CI tests since they are failing because some certificates being tested are actually failing a new lint!

Yup, that's to be expected :+1: You should be able to follow the new docs I added last week for this case: https://github.com/zmap/zlint/tree/master/integration#adding-a-new-lint Let me know how it goes, you're likely the first person besides myself to try it and there might be some rough edges to work out.

lints/lint_mp_ecdsa_allowed_algorithm.go
cpu: @fotisl I think this feedback is still unresolved. The `Status: Error` lint result doesn't have a `Details` message.

lints/lint_mp_ecdsa_allowed_algorithm.go
cpu: >  It's indeed a new lint(s), but it is relevant before 2.7 becomes live as the language in the current policy also excludes the use of non RSA or P-256/P-384 for public keys, regardless of what the signing key and algorithm was

I agree with @tadukurow here that there is an outstanding requirement to flag non-RSA/P-256/P-384 public keys but I don't think we should hold up this PR waiting for a separate lint. The review cycle on this branch has been pretty long and I think we should keep the focus narrow. If folks are amenable I'd suggest we file a separate issue for the new lint and move forward trying to land the lints in this branch now.

lints/lint_mp_allowed_eku.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (l *allowedEKU) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubCA(c) && !util.IsInMozillaRootStore(c)
+}
+
+func (l *allowedEKU) Execute(c *x509.Certificate) *LintResult {
+	if len(c.ExtKeyUsage) == 0 || util.HasEKU(c, x509.ExtKeyUsageAny) ||
+		(util.HasEKU(c, x509.ExtKeyUsageEmailProtection) && util.HasEKU(c, x509.ExtKeyUsageServerAuth)) {
+		return &LintResult{Status: Error}
```
cpu: @fotisl Related to my suggestion to pull out the mozilla trusted roots data utility you wrote into a separate PR I would prefer to see this lint return a warning and not rely on `util.IsInMozillaRootStore`.

My reasoning here is that I'd prefer not to merge the static data in `mozilla_trusted_roots_data.go` as part of this PR because it won't be something we can trivially update until the tooling is ready. My other concern is that the correctness of the Error lint results will depend on the correctness of that static data which depends on the correctness of your tooling, ultimately putting that code back into scope of review.

My preference would be to remove the static data and `util.IsInMozillaRootStore` function, land this PR in the next week or so with a less strict result status, and then work on getting the tooling, static data, and a more strict lint landed as a follow-up.

What do you think?

cpu:
I'm going to close this PR since the original author is time constrained and hasn't had a chance to update the branch with the review feedback left previously.

I've opened a replacement that carries on from the great work @fotisl did in this branch (thanks again!) so we can get this work merged: https://github.com/zmap/zlint/pull/353

PR #330, opened 2019-11-24
lints: add w_extra_subject_common_names lint

cpu:
The BRs don't expressly forbid having multiple subject common names but it seems counter to the intentions of section 7.1.4.2.2 and generally weird enough that ZLint should produce a Warn level finding (like cablint does already).

To implement this lint `github.com/zmap/zcrypto` is updated to [7f2fe6f](https://github.com/zmap/zcrypto/commit/7f2fe6faf8cf07ac390ca95dcc3d46b7a4aae180), the tip of master at the time of writing. Notably this brings in a change that stores multiple subject CN values into the `Certificate.Subject` `pkix.Name` field.

Along the way, I bumped the Go mod version to 1.13 and updated the README to reflect we expect Go 1.13+ to use zlint master (it's what our CI is using already).

Integration tests flagged 16 certificates tripping the new lint. Those certificates are:

https://censys.io/certificates/1cd00ff04092b4c2faa7becd76c44f0a7ca38fbb2269d1588d81756159ca6ec6
https://censys.io/certificates/2242b07bb4393996a940a01eb08336849a5d199410715c0af3e8c6a5cd007932
https://censys.io/certificates/225d57740aa0a824c164f7c5994cccbe8627310d573c793632005e170ee07699
https://censys.io/certificates/22aa2d265a69c95792967a9c182928e01c8bfbcab1667e6e9e0259dd31041e7f
https://censys.io/certificates/2b09e53182a4ef6e440b3c39f90f3d91d8e98a6d973233d323956981e0674deb
https://censys.io/certificates/2cff4e44a9fb9563e55217300f4d5e49f73c20101aed5725e7b6b04d539cff9e
https://censys.io/certificates/6b1db534a22b2a3ce9eaa54b5ce8720391db6ca60014edd2ff4b78d7bffc2cb9
https://censys.io/certificates/7d95620a0993673e289e0ae894566dd6c0b03e936d1c0a4d3484d8898d9296ae
https://censys.io/certificates/824c4c4f893962023dc256d48ce52ac667f7b504e9a46947c3207b57f64d7ad0
https://censys.io/certificates/8795bcd44516c8643d2a41a6da735d556b989a4c2c96c974bb11c0811404f479
https://censys.io/certificates/a04c2da1bccb1a97c6a882a4b8688941673adc715d8d2769eec345cb1c4e3b52
https://censys.io/certificates/c289369fed510acf2653b8a8eb8bee949d3c18f8d8a2817c6373ad2ec7e789d9
https://censys.io/certificates/c6bb0cb620135851db0eb8ca54d0e8b77c1565683e405a3177e460e8fd3cd9cc
https://censys.io/certificates/e18e4a9277e96542371436a1b78f1d3e09bf1095a44d2e7c02e049945a3dd66a
https://censys.io/certificates/f006baa460163b35cbc35d96ffb20aa1ec27f2a8edf5db99a8ad51ab7b5bb88b
https://censys.io/certificates/f658d49a2dc3332e2e0d893cbeb3d982d404bf5bf17d7e2e925e38ba7093e174

All of them were checked to ensure that they do in fact have multiple subject common names and are being linted correctly.

I also took this as a chance to describe the process of adding a new lint and updating the golden data for `integration/README.md`.


lints/lint_extra_subject_common_names.go
```diff
@@ -0,0 +1,51 @@
...
+func (l *extraSubjectCommonNames) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c)
+}
+
+func (l *extraSubjectCommonNames) Execute(c *x509.Certificate) *LintResult {
+	// Multiple subject commonName fields are not expressly prohibited by section
+	// 7.1.4.2.2 but do seem to run afoul of the intent. For that reason we return
+	// only a Notice level finding here.
+	if len(c.Subject.CommonNames) > 1 {
+		return &LintResult{Status: Notice}
```
zakird: I think this should just be a warning rather than a notice. The lint is named that way already, and I don't think there's much contention that it's against community best practices to have more than one CN in a certificate.

zakird: Hrm, seems to violate what we have in our README and how we designed levels originally. This certainly seems more than just an "INFO" though. I'm not sure what we do about community standards. I suspect that people just ignore `INFO`s. This might be best thought about similar to the Mozilla lints where a source might just be "Community Best Practices" that aren't as stringent as BR or RFC set. Thoughts?

Regardless, name of lint should match what it returns.

cpu: > Hrm, seems to violate what we have in our README and how we designed levels originally. This certainly seems more than just an "INFO" though. 

Agreed. Personally for this case I'm happy to use Warn even if it doesn't violate a direct SHOULD/SHOULD-NOT. I think what's in the README should be considered advice and not law.

Do you feel OK calling it a Warn on the basis of two votes or should we try to solicit input from some of the other people using zlint? 

>  I suspect that people just ignore INFOs.

I suspect so too.

> Regardless, name of lint should match what it returns.

+1. I've put it to Warn to match the name in branch. If consensus goes another way I'll switch the name.

cpu: @tadukurow @cardonator Do you folks have a preference for warn vs info for this lint?

cpu: SGTM. I forgot it's American thanksgiving and the folks I tagged might not be around to give input.

I'm going to merge this now. If someone has strong feelings after the fact there will be time between the merge & cutting a 1.1.0 release tag where we can land an adjustment. I'll aim to cut that tag later next week.

cardonator: @cpu in our case, I would say WARN is fine. I agree that even though it's not codified in the BRs, it is heavily frowned upon in CABF and SCWG. Also with the recent discussion of "default-deny" in the lists it just makes sense to lint for something that would cause a commotion.

cpu: @cardonator Thanks for weighing in. Appreciated!

PR #331, opened 2019-11-26
README: Add Camerfirma to users list.

cpu:
Camerfirma says they run a "local zlint service" that lints certificates before being sent to CT logs (hopefully before the precert is signed).


juanangelmg:
> Camerfirma says they run a "local zlint service" that lints certificates before being sent to CT logs (hopefully before the precert is signed).

Hello Daniel, when you says 'hopefully before the precert is signed', you're talking about signing the real certificate, not the pre-certificate with the poison extension, isn't it?

Thanks in advance
Juan Ángel

cpu:
Hi @juanangelmg,

No, I specifically mean the pre-certificate with the poison extension. My understanding is that if a CA signs a poisoned pre-certificate with an intermediate issuer that is subject to the BRs then that CA has comitted to issuing the final certificate, warts and all.

From the perspective of linting that means if you wait until after issuing the precertificate to run zlint it's too late to *prevent* the harm from happening. All you can do is detect that a miss-issuance has occurred and invoke the required response (disclosure to root programs, etc).

That's why in [CFSSL](https://github.com/cloudflare/cfssl) the Zlint integration will first sign the pre-certificate with [a test key](https://github.com/cloudflare/cfssl/blob/ebe01990a23a309186790f4f8402eec68028f148/signer/local/local.go#L47-L49) and then lint it. If there are findings above a configurable threshold the signing process is aborted before the pre-certificate is signed by the real trusted issuer.

Does that make sense? (_Btw do you work for Camerfirma? Can I tag you on future threads where we're looking for input from CAs running Zlint? That happens time to time when we're evaluating new lints_).

juanangelmg:
Hi @cpu 

Crystal clear.

I work for Camerfirma and you can tag me on future threads. 

Thank you very much

cpu:
> I work for Camerfirma and you can tag me on future threads.

Perfect, thank you!

PR #337, opened 2019-12-23
Initial split of lints by source

zakird:
This is an initial proposal for how to split up lints:
```
apple/
cabf_br/
cabf_ev/
community/
etsi/
mozilla/
rfc/
```

I haven't fixed the code up (so ignore tests failing), but that should be some trivial sed'ing once we agree on what we'd like the breakdown to be.

@mtgag, @sleevi, @cardonator, @cpu let me know if your thoughts, and I'll wrap up once we agree.


sleevi:
# Apple
Where would you imagine a linter for https://support.apple.com/en-us/HT210176 to go? apple_root_policy? Would it make sense to just have `apple` as the directory name?

# IETF
Would it make sense to unify them into a single directory, like `ietf` or `pkix`? I ask, because of specs like [RFC 3279](https://tools.ietf.org/html/rfc3279) and the subsequent documents that have updated it (4055, 4491, 5480, 5758, and 8692), or specs like [RFC 8410](https://tools.ietf.org/html/rfc8410) which don't.

# CABF
I think your split on cabf makes sense; the work to unify the servercert specs (EVGs and BRs) has stalled out, and if/as other CAs look to lint other requirements (for example, codesigning WG or S/MIME WG outputs), having `cabf_codesigning` or `cabf_smime` makes more sense, unless we'd also consider secondary directories (e.g. `cabf/br`, `cabf/ev`, `cabf/smime`... or, "worse", `cabf/servercert/br`, `cabf/servercert/ev`). The only thing I'd suggest is "br" vs "brs", since the s isn't part of the acronym and the lack of case sensitivity makes it a little awkward.

zakird:
I don't have a strong opinion on these:

RFCs: Yes, I agree we should merge, if only just because of how self-referential they are. 

Re Apple: right now there's a single lint and it has to do with CT. I'm fine merging. Would you expect a single directory for Mozilla as well? 

Makes sense re BR vs BRs. I'll commit these changes in a moment here.

sleevi:
Yeah, the new directory structure looks good to me. The split of `apple`, `mozilla`, and potentially `microsoft` at some point makes sense, I think. I don't think we'll see a `ccadb` lint because that's only about the audits.

In the event that multiple root programs implement the same requirement, and that requirement is not part of the BRs (of which there are several requirements like that already), I imagine it can just go into whatever vendor announces/implements first (if upcoming), or can just go into a random vendor. An alternative would be to run the same lint multiple times, one for each vendor. I have no strong feelings there as to how that situation is managed.

zakird:
alright, @cpu, all yours. my condolences. 

1) `lints/base_lint` -> `lint/` such that it can be used by all lint packages
2) all lints moved into appropriate directories under `lints/`, e.g., `lints/cabf_br`
3) lots of glue fixes now that everything has moved around

cpu:
> alright, @cpu, all yours. my condolences, this is a monster.

:laughing: Thanks for working on this one. I'll start digging through the diff when I'm back at work next week.

At a high level the breakdown of lint sources you landed on looks good to me.

> I've made no API changes since there's so much file churn here. 

:+1: SGTM.

> One thing I'll note is that there are a lot of things that seem to be touching lint.Lints in the zdns executable itself, which I don't like. 

You mean the `zlint` executable here right, not znds?

> We should add some API endpoint in the next round such that nothing outside of the zlint package is playing with lint.Lints. If there are users depending on that, this is going to break them.
Sounds reasonable to encapsulate that away.

No objections from me. Boulder/CFSSL are both using `LintCertificate` and not touching `lint.Lints`.

> I think we'll want to release a major revision once we merge in the next PR.

Agreed, once this PR is merged master will have breaking changes from the `LintStatus` move.

zakird:
Yes, zlint not zdns. Spent yesterday working on both. 🤦‍♂

cardonator:
Just to drop my thoughts here.

Thanks for doing this, Zakir! The directory structure looks great to me, I think it will work well for the stated goals and also make it easier for people who want to extend zlint for their own purposes to do so in a sensible way that lines up with the above folder structure. Hopefully, it will also make it easier for others to contribute to sets of lints they are concerned with. Really appreciate the effort here!

cpu:
Starting to chew through the diff in this branch. It seems like the context of the "move" is being lost and each lint is represented in the Github UI as a removal of the old code and then an addition of new code in the new location. 

That makes it hard to check that the lint logic itself hasn't changed since there's no side-by-side diff. Because of that I'm largely doing a skim to match the new location with the lint source, checking no lints were lost, and verifying the unit/integration tests pass. 

I think it would be a waste of effort to try and go back and remake this branch with separated commits for the move/update. Similarly doing a close review of the 1,000+ changed files vs their old version won't be doable. I just want to make sure I represent my review process honestly with this comment.

cpu:
@zakird I'll leave this one for you to merge once you think it's ready. I updated the PR description quickly to reflect the current directory structure instead of your initial proposal.

Issue #344, opened 2020-01-07
Support configurable lint sources for LintCertificate


cpu:
In https://github.com/zmap/zlint/pull/337 we split lints into directories by category. We should follow up on this change by implementing a breaking API change that will allow consumers to provide an options struct to the top level `zlint.LintCertificate` function that allows choosing lint categories/sources to use to lint the certificate.

We will also need to decide on default lint categories. My current preference is to match the existing ZLint behaviour and run **all** categories unless otherwise specified but I could be convinced if folks think just the RFC or CABF categories should be used by default.

cpu:
I've assigned myself this issue but may not have a chance to submit a PR for a little bit. If someone else is eager to take this let me know and I'll re-assign.

cardonator:
Personally, I agree with running all categories by default unless otherwise specified. Also, does it make sense to have include _and_ exclude category capability? (I'm specifically thinking of ETSI here where you might want to "run everything except ETSI".

cpu:
> Also, does it make sense to have include and exclude category capability? 

That sounds nice :+1:

cpu:
> My current preference is to match the existing ZLint behaviour and run all categories unless otherwise specified but I could be convinced if folks think just the RFC or CABF categories should be used by default.

Chatting about this more with @zakird I'm open to the idea of defaulting to all lint sources just for CI and having the library offer a more conservative default (that's his preference). I'm primarily worried about bit rot in CI if there are lint categories that aren't exercised during integration tests.

timfromdigicert:
If everything is run by default, and you get something you don't like, you can turn that off.

If everything isn't run by default, and you don't get something you needed, you're guaranteed not to notice.

weddi-eddy:
It'd be really cool to have a switch to be able to select what you want. By default, I'd say it does everything (which, perhaps there's a switch that simply lists the folders and can also list the lints in each folder; something like zlint --list-folders or something).  If you run zlint as per normal, it does the default folders, whatever is considered to be default.  If you want specifics, then you can add a switch that allows you to select the folders based on the results from the --list-folders switch I suggest above.

zlint --only-folders cabf,mozilla,ev

That would be easy on any scripts that are run and the same logic can be used for API.

weddi-eddy:
Thinking on this a bit more, I think there should be a way of easily telling what is included by default (for sanity's sake) so you know what ZLint is polling against.

zakird:
[...]

PR #364, opened 2020-01-17
lints: better test utils, avoid accessing lint.Lints directly

cpu:
This is a follow up from #337 addressing the use of `lint.Lints` throughout the unit tests. This in turn will unblock switching `lint.Lints` to be unexported which will give the project more freedom to restructure for #344.

At a high level:
* I removed the `testlint/testDef` directory. I'm not sure what it was used for historically but it appears unused today.
* I flattened the `testlint/testCerts` directory into just `testdata/` at the root of the project. All of the test files are certs and I think this is a little bit cleaner.
* I moved the existing test helpers from `util/testTools.go` to a dedicated `test` package. That makes it clearer that only unit tests should be consuming things from that package vs `util` which is used by lint logic.
* I reworked the test helpers, adding a new `test.TestLint` function that is used by the vast majority of the unit tests to load a cert from the test dir and execute a lint with a given name. Presently this function uses the `lint.Lints` map directly but now we have 1 place to update when that field is no longer exported instead of 200+.

This is a big diff, but it should be a functional no-op (only unit tests and test data are changed).


cpu:
>  Files changed 844

:sob: Sorry @zakird!

PR #370, opened 2020-01-22
Add reference to RFC 6818 to clarify explicitText

jsha:


sleevi:
This is unnecessary, given https://github.com/zmap/zlint/blob/e3ad0f9eba10b1fa0ee70e6581a627cdfeaa590c/lints/rfc/lint_ext_cert_policy_explicit_text_not_utf8.go#L66 ?

Not an objection, just trying to understand the relationship between things and the style for the lints.

jsha:
This may just be a matter of me being insufficiently familiar with the style for these lints, but I didn't think to look at the bottom of the file. My workflow was:

- Read the clause in RFC 5280
- Look for a matching lint in zlint
- Check that the lint matches what I expect
- Compare behavior against quoted text
- Get confused that what I read in the quoted text doesn't match what I just read in RFC 5280.

So, submitted this as a possible clarification for other folks in the same position, but I'm also fine with it being closed as not following common style.

A broader style fix to address the same confusion might be to move all the `func init`'s to the top of each lint, so the `Citation` field is closer to the text being cited. Or possible even make the citation text part of the `lint.Lint` object.

sleevi:
Yeah, I'm not sure the style either, which is why I meant to ask the question than the statement. I didn't notice it until expanding to view how it was implemented, and then saw it already referenced 6818.

Definitely, moving init() to be the start would serve a lot of self-documentation for those of us who read top-down :) I filed https://github.com/zmap/zlint/issues/371 as my poor attempt to explain your great idea :)

zakird:
I agree makes sense to move, but shouldn't get in the way of merging this correction.

PR #372, opened 2020-01-23
zlint: refactor lint reg., allow filtering lints used.

cpu:
This replaces the exported `lint.LintMap` field (`map[string]*Lint`) that was used by `RegisterLint` with a more robust solution based around a `Registry` interface. This allows ZLint users to include/exclude lints by name or source.

## Top Level API

The `lint.RegisterLint` function remains the same, meaning individual lints are not changed. Lints that call this function are added to the global registry. Clients can access this registry with `lint.GlobalRegistry()`. Similarly the top level `zlint.LintCertificate` remains unchanged. It lints the provided cert with all lints in the global registry.

The old `zlint.LintCertificateFiltered` function that accepted a lint name regex to filter the lints applied is replaced by a new function `zlint.LintCertificateEx` that allows specifying a `lint.Registry` explicitly. The same regex filtering can be done by pre-filtering the provided registry (See notes below).

The `lint.Source` type was changed from an int enum to a string enum. This makes it easier to work with as a consumer (e.g. via command line flags, and JSON output) and since the number of lints (and sources) is small the benefits to using an int enum type are minimal. The serialized form of Lints now includes the `Source` field in the output as `"source"`.

## Registry

The `Registry` interface also allows finding all lint names with `Names()`, finding all lint sources with `Sources()`, finding a specific lint by name with `ByName()`, and finding all lints for a given source
with `BySource()`.

The `zlint.EncodeLintDescriptionsToJSON` function is now implemented by the `Registry` interface as `WriteJSON`. This makes it easier to encode a subset of the Registry's lints by filtering the global registry.

~Like before (with the exported `map[string]*Lint`) the registry is not safe for concurrent updates. That's fine for the current ZLint codebase but is something we may want to consider addressing in the future.~ _Edit: I decided it made sense to add locking to future proof the implementation for thread safety, see 6072e24 The implementation in this branch is now safe for concurrent access/registration_

## Registry Filtering

Filtering of lints to be run is now done with the `lint.Registry.Filter` function and corresponding `lint.FilterOptions` type. This allows filtering a registry to include/exclude lints by name (or using a name regex), and to include/exclude lints by source.

By filtering the global registry and then providing it explicitly to `zlint.LintCertificateEx` callers have control over exactly what lints will be applied.

Filtering operations are applied with the following precedence: excludes by source > includes by source > excludes by name > includes by name.

E.g. excluding a source and then trying to include a lint in that excluded source by name will not work. The source exclusion happens first.

## ZLint CMD Updates

The `zlint` command (`cmd/zlint/main.go`) is updated to add four new command line flags:

1. `-list-lints-sources` - Prints a list of lint sources, one per line.
2. `-excludeSources` - Comma-separated list of lint sources to exclude.
3. `-includeSources` - Comma-separated list of lint sources to include.
4. `-nameFilter` - Regex used to match lint names to include (cannot be used at the same time as `-excludeSources` or `-includeSources)

Two existing flags are renamed:

1. `-include` becomes `-includeNames`
2. `-exclude` becomes `-excludeNames`.

Notably all three list flags (`-list-lints-json`, `-list-lints-schema` and `-list-lints-sources`) now operate **after** applying the include/exclude filters, allowing an easy way to find which lints/sources will be run with the filtered command line flags in use. 

## Integration Test Updates

Matching the `zlint` command the integration test (`integration/integration_test.go`) command line flags are updated to allow including/excluding lints by source.

Resolves https://github.com/zmap/zlint/issues/344


cpu:
Based on some out of band feedback I'm going to try and restructure the `Registry` into a `Linter` type and remove the `zlint.LintCertificate` / `zlint.LintCertificateEx` package functions, instead defining the lint operation as a function on the linter type, which can ben constructed similar to a registry in this branch. Will hopefully push another commit in the next week. If folks have additional feedback it may be best to wait until after that lands so we're all on the same page.

cardonator:
Thanks for all your hard work on this, @cpu. It's looking really good so far. 

With the large amount of changes in this PR, would it also be possible to add a flag that tells LintCertificate to only return lints that met some condition? I.E. in my case, I only care about lints that had a status that wasn't NA or Pass. It would be nice if LintCertificate could be instructed to only return those lints and not every lint that ran. This would be particularly helpful when using the CLI as the output of the tool is pretty hard to parse within a terminal.

cpu:
> Thanks for all your hard work on this, @cpu. It's looking really good so far.

Thanks for taking a look!

> would it also be possible to add a flag that tells LintCertificate to only return lints that met some condition?

@cardonator That seems like a reasonable request. I had to implement this sort of filtering in the CFSSL ZLint integration: https://github.com/cloudflare/cfssl/blob/644917271238216c94866f021e2b24ce54555848/signer/local/local.go#L183-L185

I'll see about rolling that in to this branch (or will file a follow-up issue if I don't get to it).

cpu:
> Based on some out of band feedback I'm going to try and restructure the Registry into a Linter type and remove the zlint.LintCertificate / zlint.LintCertificateEx package functions, instead defining the lint operation as a function on the linter type, which can be constructed similar to a registry in this branch.

I spent some time fiddling with this (WIP/broken code here: https://github.com/cpu/zlint/tree/cpu-refactors-lint-registration-again). My experience is that the idea doesn't play nicely with the way lint registration is handled, and the split between the `zlint` and `lint` packages. Ultimately I couldn't get something working that wasn't a big re-architecture and worked well for consumers. The linked branch comes close but since consumers don't need to import the `zlint` package in that design (e.g. see `cmd/zlint/main.go`) no lints are actually registered.

I'm still happy with the design in this PR w/ the `Registry` type. It strikes me as a good compromise between having the features we want and maintaining the existing approach to registering lints and the existing subpackages. I vote we move forward with this branch. If folks disagree I'd be happy to close this PR and let someone else take a crack at a proposed redesign.

cmd/zlint/main.go
```diff
@@ -62,32 +71,40 @@ func init() {
...
 		return
 	}
 
 	if listLintsSchema {
-		names := make([]string, 0, len(lint.Lints))
-		for lintName := range lint.Lints {
-			names = append(names, lintName)
-		}
-		sort.Strings(names)
+		names := registry.Names()
```
dadrian: This is fine now, but we should probably drop public support for ZSchema within ZLint.

lint/registration.go
```diff
@@ -0,0 +1,350 @@
...
+func (r *registryImpl) WriteJSON(w io.Writer) {
+	enc := json.NewEncoder(w)
+	enc.SetEscapeHTML(false)
+	for _, name := range r.Names() {
+		_ = enc.Encode(r.ByName(name))
+	}
+}
+
+// newRegistry constructs a registryImpl that can be used to register lints.
+func newRegistry() *registryImpl {
```
dadrian: I would make this function public, to make it easier for third-parties to create a registry of custom lints. AFAICT, right now to do so, they would have to reimplement Registry fully, or modify the global registry.

I like having the global registry since it's what most people want, most of the time, but having a separate registry also seems useful.

dadrian:
Overall, I think this is really good @cpu. If we can clear up CI, I'm happy to merge.

cmd/zlint/main.go
cpu: @dadrian Honestly I'm not 100% clear on what ZSchema is :laughing: I assumed it was important to Censys/ZMap but if it isn't I'm happy to rip it out now. We're already breaking back compat.

cpu:
> If we can clear up CI, I'm happy to merge.

@dadrian I think the CI failure was unrelated to the code. Yesterday I accidentally pushed my branch to the `zmap` repo first (this PR is from my fork). When I deleted the `zmap` copy of the branch I think [it confused Travis](https://travis-ci.org/zmap/zlint/builds/648569854?utm_source=github_status&utm_medium=notification):

```
$ git clone --depth=50 --branch=cpu-refactors-lint-registration https://github.com/zmap/zlint.git zmap/zlint
Cloning into 'zmap/zlint'...
warning: Could not find remote branch cpu-refactors-lint-registration to clone.
fatal: Remote branch cpu-refactors-lint-registration not found in upstream origin
```

It looks like CI is [happy again](https://travis-ci.org/zmap/zlint/builds/648958295) as of d4cbcd6.

cmd/zlint/main.go
cpu: I went ahead and ripped it out in 2a7674f. Can always revert that commit if folks would prefer a more gradual deprecation.

cpu:
> I'm happy to merge.

One other note: can I request this gets squash merged with the PR desc as the commit message? I think that will be the cleanest. Alternatively I can rebase and squash to one commit before merging but I think it's easier to let Github do the work.

cpu:
> I'll see about rolling that in to this branch (or will file a follow-up issue if I don't get to it).

Unfortunately I had to punt lint result filtering to a separate issue for now: https://github.com/zmap/zlint/issues/384 I'll try to come back around at that after a RC and reviewing the other open PRs.

cardonator:
It's all good, @cpu! Thanks for all your work on these large refactorings!

Issue #356, opened 2020-01-15
Lint for Mozilla PKI policy RSASSA-PSS encoding requirements


cpu:
See https://github.com/mozilla/pkipolicy/blob/master/rootstore/policy.md#511-rsa

> RSASSA-PSS with SHA-256, MGF-1 with SHA-256, and a salt length of 32 bytes.
>
> The encoded AlgorithmIdentifier MUST match the following hex-encoded bytes:
>
> 304106092a864886f70d01010a3034a00f300d06096086480165030402010500a11c301a06092a864886f70d010108300d06096086480165030402010500a203020120
> 
> RSASSA-PSS with SHA-384, MGF-1 with SHA-384, and a salt length of 48 bytes.
> 
> The encoded AlgorithmIdentifier MUST match the following hex-encoded bytes:
> 304106092a864886f70d01010a3034a00f300d06096086480165030402020500a11c301a06092a864886f70d010108300d06096086480165030402020500a203020130
>
> RSASSA-PSS with SHA-512, MGF-1 with SHA-512, and a salt length of 64 bytes.
>
> The encoded AlgorithmIdentifier MUST match the following hex-encoded bytes:
>
> 304106092a864886f70d01010a3034a00f300d06096086480165030402030500a11c301a06092a864886f70d010108300d06096086480165030402030500a203020140

cpu:
I've heard out of band from @mtgag that they will be working on submitting a PR with an implementation for this issue. If someone else would like to work on this please coordinate efforts.

cpu:
Resolved by https://github.com/zmap/zlint/pull/377

PR #377, opened 2020-02-03
lints: enforce Mozilla PKI policy RSASSA-PSS encoding requirements

mtgag:
Implementation of https://github.com/zmap/zlint/issues/356


cpu:
@mtgag If you wouldn't mind could you resolve the conflicts on this branch? I'll start reviewing it afterwards. Thanks!

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/zmap/zlint/v2/util"
+)
+
+type rsaPssAidEncoding struct{}
+
+var allowedAidEncodingsForPSS = [3]string{
+	"304106092a864886f70d01010a3034a00f300d06096086480165030402010500a11c301a06092a864886f70d010108300d06096086480165030402010500a203020120",
+	"304106092a864886f70d01010a3034a00f300d06096086480165030402020500a11c301a06092a864886f70d010108300d06096086480165030402020500a203020130",
+	"304106092a864886f70d01010a3034a00f300d06096086480165030402030500a11c301a06092a864886f70d010108300d06096086480165030402030500a203020140"}
+
```
cpu: I don't think it makes sense to store these values as strings of hex encoded bytes. It means you have to decode them again in `Execute` for each certificate that's linted. It would be better to express these values as a `[]byte`. I also think you'll want to express this as a map from OID to expected `[]byte` (see my later comments in `Execute` about the `for` loop based comparison).

This would also match how the `CheckAlgorithmIDParamNotNULL` function used by the `e_tbs_signature_rsa_encryption_parameter_not_null` lint is implemented: https://github.com/zmap/zlint/blob/5b2df5c915f1d73dbfcd4f2ac2a6241103fbc4a3/v2/util/algorithm_identifier.go#L14-L32

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return &lint.LintResult{Status: lint.Error, Details: "error reading signatureAlgorithm from TBS"}
+	}
+
+	for _, encoding := range allowedAidEncodingsForPSS {
+		expectedEncoding, _ := hex.DecodeString(encoding)
+		if bytes.Equal(signatureAlgoID, expectedEncoding) {
+			return &lint.LintResult{Status: lint.Pass}
+		}
+	}
```
cpu: I think it would be better to use the approach taken in https://github.com/zmap/zlint/blob/5b2df5c915f1d73dbfcd4f2ac2a6241103fbc4a3/v2/util/algorithm_identifier.go#L36 and https://github.com/zmap/zlint/blob/5b2df5c915f1d73dbfcd4f2ac2a6241103fbc4a3/v2/lints/rfc/lint_tbs_signature_rsa_encryption_parameter_not_null.go#L43

E.g. extract the signature OID from `RawSubjectPublicKeyInfo` with `cryptobyte`, and then use that as a lookup into a map of expected `[]byte`.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
```diff
@@ -0,0 +1,98 @@
...
+	"encoding/hex"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v2/lint"
+	"github.com/zmap/zlint/v2/util"
+)
+
+type rsaPssAidEncoding struct{}
+
+var allowedAidEncodingsForPSS = [3]string{
```
cpu: I think the name of this var should indicate that it's DER encoded bytes. e.g. like the similar `util.RSAAlgorithmIDToDER`.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	for _, encoding := range allowedAidEncodingsForPSS {
+		expectedEncoding, _ := hex.DecodeString(encoding)
+		if bytes.Equal(signatureAlgoID, expectedEncoding) {
+			return &lint.LintResult{Status: lint.Pass}
+		}
+	}
+
+	return &lint.LintResult{Status: lint.Error, Details: "RSASSA-PSS parameters are not properly encoded"}
```
cpu: It would be nice to provide more information in this error (e.g. what was the OID found that didn't match the expected byte pattern?).

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_mp_rsassa-pss_parameters_encoding_correct",
+		Description:   "The encoded AlgorithmIdentifier for RSASSA-PSS MUST match specific hex-encoded bytes",
```
cpu: I think it's worth noting in the name/desc that this lint is checking the signature algorithm encoding (e.g. vs a public key algorithm encoding, which is covered by `e_mp_rsassa-pss_in_spki`).

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
mtgag: I am not sure about the map approach. In the case of RSAPSS this would look something like that:

var RSASSAPSSAlgorithmIDAndParametersToDER = map[string][]byte{
	"RSASSA-PSS with SHA-256, MGF-1 with SHA-256, salt length 32 bytes": {0x30, 0x41, .... 0xa2, 0x03, 0x02, 0x01, 0x20}
}

because there is no direct mapping of an OID to DER encoding. The OID for PSS is always the same. I will change this to working with bytes instead of hex encoded strings. Ok?

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
mtgag: See also https://github.com/zmap/zlint/pull/377#discussion_r386237154. In the case of RSAPSS is not possible to extract an OID and check that it maps to the correct value. I believe to check whether the encoding is one of the three allowed ones is (in the case of RSAPSS) better. Ok?

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
mtgag: Added.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
mtgag: More info is provided now.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
mtgag: Changed name.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
mtgag: Changed to map for better readability, but the key of the map is not used.

cpu:
Thanks for the iteration @mtgag! I'll be travelling over the next few days but will endeavour to give the code & your replies my attention next week.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
cpu: > because there is no direct mapping of an OID to DER encoding. The OID for PSS is always the same.

Ahhhhhh! I understand now. Apologies, I mistakenly thought the OID would change between these cases and could be used to index the expected bytes for the DER encoding of the parameters. Thanks for your patience/explanation.

I think the best solution in this case is a hybrid between what you have now (a `map[string][]byte`) and what you had originally (a `[3]string`). Since the `map` keys aren't used anywhere in the code I think they're better expressed as comments. I would like to see the `[]byte` constants remain the same. So I think the best approach will be using `[3][]byte` and adding comments on each constant. E.g. something like:

```go
var RSASSAPSSAlgorithmIDToDER = [3][]byte{
        // RSASSA-PSS with SHA-256, MGF-1 with SHA-256, salt length 32 bytes
	{0x30, 0x41, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x0a, 0x30, 0x34, 0xa0, 0x0f, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x01, 0x05, 0x00, 0xa1, 0x1c, 0x30, 0x1a, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x08, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x01, 0x05, 0x00, 0xa2, 0x03, 0x02, 0x01, 0x20},
        // RSASSA-PSS with SHA-384, MGF-1 with SHA-384, salt length 48 bytes
	{0x30, 0x41, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x0a, 0x30, 0x34, 0xa0, 0x0f, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x02, 0x05, 0x00, 0xa1, 0x1c, 0x30, 0x1a, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x08, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x02, 0x05, 0x00, 0xa2, 0x03, 0x02, 0x01, 0x30},
        // RSASSA-PSS with SHA-512, MGF-1 with SHA-512, salt length 64 bytes
	{0x30, 0x41, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x0a, 0x30, 0x34, 0xa0, 0x0f, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x03, 0x05, 0x00, 0xa1, 0x1c, 0x30, 0x1a, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x08, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x03, 0x05, 0x00, 0xa2, 0x03, 0x02, 0x01, 0x40},
}
```

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
cpu: Makes sense thanks! I replied in the other comment thread.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
mtgag: Ok.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
```diff
@@ -0,0 +1,101 @@
...
+	// RSASSA-PSS with SHA-512, MGF-1 with SHA-512, salt length 64 bytes
+	{0x30, 0x41, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x0a, 0x30, 0x34, 0xa0, 0x0f, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x03, 0x05, 0x00, 0xa1, 0x1c, 0x30, 0x1a, 0x06, 0x09, 0x2a, 0x86, 0x48, 0x86, 0xf7, 0x0d, 0x01, 0x01, 0x08, 0x30, 0x0d, 0x06, 0x09, 0x60, 0x86, 0x48, 0x01, 0x65, 0x03, 0x04, 0x02, 0x03, 0x05, 0x00, 0xa2, 0x03, 0x02, 0x01, 0x40},
+}
+
+func (l *rsaPssAidEncoding) Initialize() error {
+	return nil
+}
+
+func (l *rsaPssAidEncoding) CheckApplies(c *x509.Certificate) bool {
+	return c.SignatureAlgorithmOID.Equal(util.OidRSASSAPSS)
```
sleevi: @cpu I may have missed it, but I didn't see anything in `rfc` or `cabf_br` that ensured that the the inner `signature` field matched the outer `signatureAlgorithm` field. Do you know of such a test?

The context is that the Mozilla Policy in https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/#51-Algorithms applies to both the `signatureAlgorithm` field of the Certificate, and the `signature` field of the TBSCertificate. This only tests the TBSCertificate (as does line 78)

I _suspect_ that this is perfectly fine for this PR, because it doesn't look like https://github.com/zmap/zcrypto/blob/master/x509/x509.go even exposes the ability to access the outer structure. That is, the `x509.Certificate` that is exported doesn't expose this information from https://github.com/zmap/zcrypto/blob/2177ed15981c76d13681ec89db9a1efd8e436230/x509/x509.go#L1458 , so this wouldn't be an easy fix. But just noting it in case we want to file a follow-up.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct_test.go
```diff
@@ -0,0 +1,70 @@
...
+			InputFilename:  "rsassapssWithSHA384EmptyHashParams.pem",
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "Standard RSASSA-PSS with SHA512",
+			InputFilename:  "rsassapssWithSHA512.pem",
+			ExpectedResult: lint.Pass,
+		},
+		{
+			Name:           "Standard RSASSA-PSS with SHA512 but the hash parameters are empty instead of NULL",
```
sleevi: Suggestion: These variations all only exercise a single mis-encoding (empty hash parameters). Would it be useful to include explicit hash parameters that are more esoteric, such as a different mask length?

I'm not sure if I'm underestimating the difficulty, so I thought I'd check here.

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct.go
cpu: > I didn't see anything in rfc or cabf_br that ensured that the the inner signature field matched the outer signatureAlgorithm field. Do you know of such a test?

@sleevi I don't know of any coverage for that within the existing lints :-( Thanks for raising this line of thought.

> This only tests the TBSCertificate (as does line 78)

Indeed, I think the older RSA parameter encoding lint is the same: https://github.com/zmap/zlint/blob/fd40f579253ea1ebfb18a585ab5cd8e7dcde61aa/v2/lints/rfc/lint_tbs_signature_rsa_encryption_parameter_not_null.go#L43-L44

> so this wouldn't be an easy fix. But just noting it in case we want to file a follow-up.

I think a follow-up would make the most sense. Probably expressed as two issues: one on the ZLint repo discussing the need to lint the `Certificate.signature`, and one on the ZCrypto repo discussing the need to export the outer structure. Would you like me to file those?

sleevi: > I think a follow-up would make the most sense. Probably expressed as two issues: one on the ZLint repo discussing the need to lint the `Certificate.signature`, and one on the ZCrypto repo discussing the need to export the outer structure. Would you like me to file those?

Done:
https://github.com/zmap/zcrypto/issues/213
https://github.com/zmap/zlint/issues/419

cpu: Thank you!

v2/lints/mozilla/lint_mp_pss_parameters_encoding_correct_test.go
mtgag: Added a test case for irregular (maybe invalid or unsupported is a better name for this?) salt length. Other test cases (e.g. irregular trailerField) are indeed more work to implement also because the crypto libraries do not support it out-of-the-box.

cpu: @mtgag Thanks for adding another test case. I think the coverage in-branch is sufficient to see the lint merged. It would be nice to have more malformed test cases but it sounds like the difficulty would mean holding up the PR longer and I'd prefer to see it merged and iterated on as time allows.

Since @sleevi gave this branch a +1 without requesting the test case change as blocking feedback I'm going to go ahead and merge. @mtgag If you think you'd be up to adding the more difficult test cases would you mind filing an issue to act as a marker/discussion point?

Thanks all!

PR #380, opened 2020-02-05
Qcstmt repeated

mtgag:


lints/etsi/lint_qcstatem_repeated_stmt.go
```diff
@@ -0,0 +1,92 @@
...
+		foundOidListSlice = append(foundOidListSlice, &oid)
+	}
+	return &lint.LintResult{Status: lint.Pass}
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_qcstatem_repeated_stmt",
+		Description:   "Checks that none of the ETSI QcStatements appear multiple times within the QcStatements.",
+		Citation:      "(Implicit requirement)",
```
sleevi: I've gone through RFC 3739 and ETSI EN 319 412-5, and I can't find anything to support this being an implicit requirement.

That is, there's nothing like [Section 4.2 of RFC5280](https://tools.ietf.org/html/rfc5280#section-4.2) stating:
```
   A certificate MUST NOT include more
   than one instance of a particular extension. 
```

Indeed, within RFC 3739, this is expected as a possibility, since the `semanticsIdentifier` may be used to define the semantics for a sub-group of attributes, and logically one could imagine the same identifier also being used to describe semantics for a different sub-group of attributes.

I think an example of this might be the `QcEuLimitValue` supporting liability limits in multiple currencies, since that field is not a `SEQUENCE`

That said, I wholly expect that TSPs will mess this up, and so I think it's good that you included it. I'd defer to @cpu / @dadrian here, but given that there's no text to support this, I think it might be best as a WARNING. Before spending too much time here, I wanted to double check how to handle such ambiguities in zlint

zakird: I'm going to open a separate issue to discuss this. This has come up repeatedly and I think we need make a few decisions as project and document.

sleevi: Thanks. To close out this comment, that discussion is https://github.com/zmap/zlint/issues/412 

While @zakird hasn't closed that discussion out, it does seem that there's a growing consensus that in the absence of a SHOULD-level requirement in the relevant specs (in this case, RFC 3739 or x19 412-5, we should hold off.

I took a look again through v2.2.3 (the 2020-01 on-approval document, which introduces the `esi4-qcStatement-7` statement for relevant jurisdiction), and went through Section 3.2.6 of RFC 3739, and I'm not seeing anything that might indicate a statement may only appear once.

This might make a candidate for a 'community' section, if the project goes that way. I'm going to set Changes Requested, so at least I don't forget I've looked at and pinged this. It seems the most likely outcome is either not accepting or moving to community.

cpu:

PR #381, opened 2020-02-05
lint: Psd2 policy recommended

mtgag:
Implementation of a lint for recommended psd2 policy


lints/etsi/lint_qcstatem_psd2_policy_recommended.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	return &lint.LintResult{Status: lint.Pass}
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "w_qcstatem_psd2_policy_recommended",
+		Description:   "Check the requirement that EU Qualified Certificates include at least one of the QCP qualifiers.",
+		Citation:      "FETSI EN 319 412-4",
```
sleevi:
```suggestion
		Citation:      "ETSI EN 319 412-4: 4.2, ETSI EN 319 412-4: 4.3",
```

lints/etsi/lint_qcstatem_psd2_policy_recommended.go
```diff
@@ -0,0 +1,51 @@
...
+		return &lint.LintResult{Status: lint.Warn, Details: "EU qualified certificate missing QCP policy identifier"}
+	}
+
+	return &lint.LintResult{Status: lint.Pass}
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "w_qcstatem_psd2_policy_recommended",
+		Description:   "Check the requirement that EU Qualified Certificates include at least one of the QCP qualifiers.",
```
sleevi:
```suggestion
		Description:   "EU Qualified certificates SHALL include QCStatements and SHOULD include one of the certificate policy identifiers defined in clause 5.3 of ETSI EN 319 411-2",
```

lints/etsi/lint_qcstatem_psd2_policy_recommended.go
```diff
@@ -0,0 +1,51 @@
...
+	return nil
+}
+
+func (l *qcStatemPsd2PolicyRecomm) CheckApplies(c *x509.Certificate) bool {
+	isPresent := util.HasCertAnyEtsiQcStatement(c)
+	return isPresent
+}
+
+func (l *qcStatemPsd2PolicyRecomm) Execute(c *x509.Certificate) *lint.LintResult {
+	if util.HasCertAnyEtsiQcStatement(c) && !util.HasCertAnyEtsiQcpPolicy(c) {
```
sleevi: There's a challenge here, because the ETSI QC Statements defined in 319 412-5 do not necessarily imply a certificate is a Qualified Certificate. 412-5 makes that clear in [Section 1: Scope](https://www.etsi.org/deliver/etsi_en/319400_319499/31941205/02.02.01_60/en_31941205v020201p.pdf#page=6) that they can be used in combination with any certificate profile.

That is:
- If (QCP Present), then (QC Statement MUST be present)
- If (QC Statement present), then (QCP MAY be present)

However, this is even messier, because 412-4 Section 4.3 only states that the policies from 5.3 "SHOULD" be present, not MUST. This is messy, because GEN-6.6.1-05 of 319 411-2 allows a TSP-specific policy OID to be used in lieu of the 5.3 OIDs.

So I'm not even sure this makes sense as a warning, because the SHOULD is effectively weakened to a MAY by GEN-6.6.1-05.

So I don't think we want this as a lint, but what we do want is that if they assert any of the OIDs from 5.3 of ETSI EN 319 411-2, then they're bound to include QCStatements by GEN-6.6.1-02 (although whether or not that is a hard requirement is iffy, given that only 'appropriate' QCStatements are required). Additionally, an approach like https://github.com/zmap/zlint/blob/fd40f579253ea1ebfb18a585ab5cd8e7dcde61aa/v2/util/ev.go#L21 would be needed to track the qualified certificate profiles that are CA specific.

Does that all make sense?

cpu:

PR #387, opened 2020-02-12
Psd2 policy mandatory

mtgag:
implementation of a lint for several requeriments regarding policies for PSD2


lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
```diff
@@ -0,0 +1,80 @@
...
+	if util.HasCertAnyEtsiQcStatement(c) && len(c.PolicyIdentifiers) == 0 {
+		return &lint.LintResult{Status: lint.Error, Details: "EU Qualified Certificate must at least contain one policy identifier (ETSI EN 319 411-2: GEN-6.6.1-05)"}
+	}
+	return &lint.LintResult{Status: lint.Pass}
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_qcstatem_psd2_policy_mandatory",
+		Description:   "Tests various requirements regarding policy identifiers of EU Qualified Certificates",
```
sleevi: As a description, this reads like "tests stuff". Line 66 is a good example of a descriptive statement, but we should try to find a way to tighten this up.

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
```diff
@@ -0,0 +1,80 @@
...
+		Citation:      "ETSI EN 319 412-4, ETSI EN 319 412-5, ETSI EN 319 411-2",
```
sleevi: Let's reference specific sections or normative requirements here:
```suggestion
		Citation:      "ETSI EN 319 411-2: GEN-6.6.1-05, ETSI EN 319 412-4: Clause 4.3, ETSI EN 319 412-5: Clause 5",
```

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (l *qcStatemPsd2PolicyMandatory) CheckApplies(c *x509.Certificate) bool {
+	isPresent := util.HasCertAnyEtsiQcStatement(c)
+	return isPresent
+}
+
+func policyAndTypeAreConsistent(c *x509.Certificate, policy asn1.ObjectIdentifier, policyStr string, qcType asn1.ObjectIdentifier, typeStr string, mustHaveQcStmtFour bool) string {
+
+	if util.HasCertPolicy(c, policy) && !util.HasCertEtsiQcType(c, qcType) {
+		return "EU Qualified Certificate has policy " + policyStr + " but doesn't have the corresponding QcType '" + typeStr + "' (ETSI EN 319 412-4: \"Policy identifiers included in the certificate policies extension of EU Qualified Certificates shall be consistent with the EU Qualified Certificate Statements\")"
```
sleevi: I don't see where the requirement to include a QcType exists? The current language being quoted doesn't seem to support this.

This also feels a bit duplicative of the citation. Could you explain a bit more?

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	_, isQscdStmtPresent := util.IsQcStatemPresent(c, &util.IdEtsiQcsQcSSCD)
+	if util.HasCertPolicy(c, policy) && mustHaveQcStmtFour && !isQscdStmtPresent {
+		return "EU Qualified Certificate has policy indicating private key storage in QSCD, but the corresponding QC Statement 4 is missing (ETSI EN 319 412-4: \"Policy identifiers included in the certificate policies extension of EU Qualified Certificates shall be consistent with the EU Qualified Certificate Statements\")."
```
sleevi:
```suggestion
		return "Policy indicating QSCD is present, but etsi4-qcStatement-4 is missing (ETSI EN 319 412-5: Clause 5, ETSI EN 319 412-5, Clause 4.2.2)"
```

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	if util.HasCertPolicy(c, policy) && !mustHaveQcStmtFour && isQscdStmtPresent {
+		return "EU Qualified Certificate does not feature the policy indicating private key storage in QSCD, but the corresponding QC Statement 4, indicating this, is present (ETSI EN 319 412-4: \"Policy identifiers included in the certificate policies extension of EU Qualified Certificates shall be consistent with the EU Qualified Certificate Statements\")."
```
sleevi:
```suggestion
		return "ETSI EN 319 411-2: GEN-6.6.1-04:  The qcStatement for QSCD (esi4-qcStatement-4) shall not be included in certificates that are not issued according to [QCP-n-qscd] or [QCP-l-qscd] requirements"
```

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
```diff
@@ -0,0 +1,80 @@
...
+	util.AppendToStringSemicolonDelim(&errStr, policyAndTypeAreConsistent(c, util.IdEtsiPolicyQcpNaturalQscd, "QCP-n-qscd", util.IdEtsiQcsQctEsign, "eSign", true))
+	util.AppendToStringSemicolonDelim(&errStr, policyAndTypeAreConsistent(c, util.IdEtsiPolicyQcpLegal, "QCP-l", util.IdEtsiQcsQctEseal, "eSeal", false))
+	util.AppendToStringSemicolonDelim(&errStr, policyAndTypeAreConsistent(c, util.IdEtsiPolicyQcpLegalQscd, "QCP-l-qscd", util.IdEtsiQcsQctEseal, "eSeal", true))
+	if errStr != "" {
+		return &lint.LintResult{Status: lint.Error, Details: errStr}
+	}
+	_, psd2Present := util.IsQcStatemPresent(c, &util.IdEtsiPsd2Statem)
+	_, QcTypePresent := util.IsQcStatemPresent(c, &util.IdEtsiQcsQcType)
+	if psd2Present && !QcTypePresent {
+		return &lint.LintResult{Status: lint.Error, Details: "EU Qualified Certificate has PSD2 QcStatement but not the mandatory QcType Statement (ETSI EN 319 412-5: \"When the certificate is issued in accordance with Annex III or Annex IV of Regulation (EU) No 910/2014 [i.8], this statement shall be present\")."}
```
sleevi: I'm having trouble finding supporting text that this is an error. That is, there's nothing preventing the PSD2 QCStatement from being present in a non-qualified certificate. It will not have the legal value for purposes of PSD2, but the QCStatement itself doesn't require that the certificate be qualified (at least, AFAICT from reading TS 119 495).

You've already got the requirement that QcType be present on line 38, so I think this check can just be deleted entirely (lines 60-64). Have I misunderstood a requirement?

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	if util.HasCertAnyEtsiQcStatement(c) && len(c.PolicyIdentifiers) == 0 {
+		return &lint.LintResult{Status: lint.Error, Details: "EU Qualified Certificate must at least contain one policy identifier (ETSI EN 319 411-2: GEN-6.6.1-05)"}
```
sleevi: I don't believe this is correct, because GEN-6.6.1-05 allows for "and/or" the policy identifiers from clause 5.3 or a TSP-specific OID.

So a certificate that includes a TSP-specific OID and a QCStatement, which is valid and permitted, would trigger an error here. So I don't think this is correct.

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
mtgag: Modify this implementation to:

if util.HasCertPolicy(c, policy) && util.IsQcStatemPresent(c, &util.IdEtsiQcsQcType) && !util.HasCertEtsiQcType(c, qcType) {

which additionally checks if QcType is set in the certificate. Only then must it be consistent. Ok?

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
mtgag: No. I will delete this check and also chage the name of the lint. Actually the whole lint is not really related to PSD2. I guess this slipped in is due to the very first attempts that targeted at PSD2.

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
mtgag: This checks if any (len(c.PolicyIdentifiers) == 0) policy identifier is set which I believe is correct (at least one policy identifier must be set). However we will modify this implementation to:

if len(c.PolicyIdentifiers) == 0 {
        return &lint.LintResult{Status: lint.Error, Details: "EU Qualified Certificate must at least contain one policy identifier (ETSI EN 319 411-2: GEN-6.6.1-05)"}
}

Ok? Also when this lint is triggered needs to be changed to:

func (l *qcStatemPsd2PolicyMandatory) CheckApplies(c *x509.Certificate) bool {
    return util.IsQcStatemPresent(c, &util.IdEtsiQcsQcCompliance)
}

this seems more correct. I will refactor this lint and merge main. Would it be possible to recheck?

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
sleevi: I guess what I'm trying to understand is where the requirement is to have a QcType coming from? 

412-4, Clause 4.2 basically just says "Do what clause 4/5 of 412-5" say

412-5, Clause 4.2.3 doesn't mandate the use of QCType, and the interpretation of whether it's an EU Qualified certificate is contingent upon the presence of a QcCompliance as described in Clause 4.2.1. However, even if Clause 4.2.1 is present (QcCompliance), that's still not sufficient, because that depends on whether QcCCLegislation (Clause 4.2.4) is present.

Further, as it pertains to QCP-w's common policy (EN 319 411-2, v2.2.2, Clause 5.5.5), it's unclear whether that is intended to meet the requirements of Annex IV (how 319 412-5 v2.2.3, Table 2 describes it) or whether it's intended to meet articles 3 (38) and 45 of the eIDAS regulation (how 319 411-2, v2.2.2 clause 5.5.5 describes it)

So I think this is very much a separate check from PSD2, but here's my understanding of the check:
- If QcCompliance is asserted, without QcCCLegislation, then:
  - If QcType is present, it's compliant with Annex IV
  - If QcType is absent, it MAY be a Lint violation, or it MAY be a signal that the certificate complies with Directive 1999/99/EC (per Table 1 of 319 412-5).
    - If the QCP-w policy is asserted, then the absence of QcType is a bug, because the QCP-w asserts compliance with eIDAS
    - If there is no QCP-w policy asserted, then it depends on the CA's certification date
- If QcCompliance is asserted, with QcCCLegislation, then:
  - QcType is undefined whether it's required or not, depending on the relevant MOU between the EU and the relevant CCLegislation

But I definitely think teasing this out should be a separate lint, because we have multiple dimensions here to worry about. For example, QcCompliance is clearly required when asserting the policy OIDs from Clause 5.5 of 319 411-2, but it's ambiguous in other cases. The semantics of QcType depend on that assertion, since QcType can be used without QcCompliance to indicate a non-qualified certificate.

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
sleevi: Great. Sorry it took me so long to trace through the requirements. The profile for Qualified certificates versus the permissiveness for non-qualified definitely makes this messy to figure out.

lints/etsi/lint_qcstatem_psd2_policy_mandatory.go
sleevi: But the existence of an ETSI QcStatement doesn't make a certificate a Qualified certificate. Many of the QCStatements can be used for non-qualified certificates, and don't require a policy OID to be set.

So I don't think your `len(c.PolicyIdentifiers) == 0` lint would be correct here, because that'd be fine.

I think we probably want to discuss a bit more as far as what particular thing we'd like this lint to check, then work through the LintApplies. Figuring out whether or not a certificate is intended to be qualified is... obnoxiously messy (as the earlier comment captures), so it seems we should sort that out.

v2/lints/etsi/lint_policy_and_qctype_are_consistent.go
```diff
@@ -0,0 +1,93 @@
...
+func (l *consistentPolicyAndQCType) CheckApplies(c *x509.Certificate) bool {
+	// Checks if the QcStatement id-etsi-qcs-QcCompliance is present. If yes, then this lint is triggered.
+	_, isEUQualified := util.IsQcStatementPresent(c, &util.IdEtsiQcsQcCompliance)
+	return isEUQualified
+}
+
+func policyAndTypeAreConsistent(c *x509.Certificate, policy asn1.ObjectIdentifier, policyStr string, qcType asn1.ObjectIdentifier, typeStr string) string {
+
+	// Check #1: Policy in question is present.
+	// if the EU qualified certificate does not have the policy in question, there will be no error regarding consistency.
```
sleevi, line 40:
```suggestion
	// If the policy is missing, not enough information exists to determine if there is an inconsistency.
```

v2/lints/etsi/lint_policy_and_qctype_are_consistent.go
```diff
@@ -0,0 +1,93 @@
...
+	_, isQcTypePresent := util.IsQcStatementPresent(c, &util.IdEtsiQcsQcType)
+
+	if isQcTypePresent && !util.HasCertEtsiQcType(c, qcType) {
+		return fmt.Sprintf("EU Qualified Certificate has policy %s but does not have the corresponding QcType '%s' (ETSI EN 319 412-4: \"Policy identifiers included in the certificate policies extension of EU Qualified Certificates shall be consistent with the EU Qualified Certificate Statements\")", policyStr, typeStr)
+	}
+
+	// Check #3: Consistency of QCStatement QcSSCD (id-etsi-qcs-QcSSCD) and policy (bidirectional).
+	_, isQscdStatementPresent := util.IsQcStatementPresent(c, &util.IdEtsiQcsQcSSCD)
+
+	qscdPolicyIsPresent := util.HasCertPolicy(c, util.IdEtsiPolicyQcpNaturalQscd) || util.HasCertPolicy(c, util.IdEtsiPolicyQcpLegalQscd)
```
sleevi, line 55: Suggestion: It may make sense to lift this to a utility function - `isQcsdPolicy` ?

It seems like you could check only based on `policy`, since you've already checked that it's present on line 41, and that would at least align with now you structured this function so that it 'could' support CA-specific OIDs in the future.

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
```diff
@@ -0,0 +1,165 @@
...
+		ExpectedResult lint.LintStatus
+	}{
+		{
+			Name:           "evAllGood.pem",
+			InputFilename:  "evAllGood.pem",
+			ExpectedResult: lint.NA,
+		},
+		{
+			Name:           "QcStmtPsd2Cert80EuQcCertWoAnyPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert80EuQcCertWoAnyPolicy.pem",
```
sleevi, line 37: Asserts QcsCompliance w/o asserting any certificatePolicies and w/o asserting any QcType

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
```diff
@@ -0,0 +1,165 @@
...
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert81EuQcCertWoAnyPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert81EuQcCertWoAnyPolicy.pem",
```
sleevi, line 42: Asserts QcsCompliance w/o asserting any certificatePolicies and w qct-eSeal

1. Does this test something useful / will this be used in a future test? AFAICT, your len(certificatePolicies) == 0 means that this covers the exact same code as Cert80, and so it's not clear if it is as useful
2. If it is useful, should it have that appended to the file name?

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
```diff
@@ -0,0 +1,165 @@
...
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert82EuQcCertWoAnyPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert82EuQcCertWoAnyPolicy.pem",
```
sleevi, line 47: Asserts QcsCompliance w/o asserting any certificatePolicies and w qct-eSign

1. Does this test something useful / will this be used in a future test? AFAICT, your len(certificatePolicies) == 0 means that this covers the exact same code as Cert80, and so it's not clear if it is as useful
2. If it is useful, should it have that appended to the file name?

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
```diff
@@ -0,0 +1,165 @@
...
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert83EuQcCertWoAnyPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert83EuQcCertWoAnyPolicy.pem",
```
sleevi, line 52: Asserts QcsCompliance w/o asserting any certificatePolicies and w qct-Web

1. Does this test something useful / will this be used in a future test? AFAICT, your len(certificatePolicies) == 0 means that this covers the exact same code as Cert80, and so it's not clear if it is as useful
2. If it is useful, should it have that appended to the file name?

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
```diff
@@ -0,0 +1,165 @@
...
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert84InconsistentPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert84InconsistentPolicy.pem",
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert85InconsistentPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert85InconsistentPolicy.pem",
```
sleevi, line 62: It wasn't clear to me how this differed from the test on 56, other than the QCT being different (eseal vs esign). It probably benefits from clearer file names, but since this does cover different code paths, I'm good here :)

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
```diff
@@ -0,0 +1,165 @@
...
+			ExpectedResult: lint.Pass,
+		},
+		{
+			Name:           "QcStmtPsd2Cert87EuQcCertWoQcPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert87EuQcCertWoQcPolicy.pem",
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert88EuQcCertWoQcPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert88EuQcCertWoQcPolicy.pem",
```
sleevi, line 77: Could you describe how this is distinct from 87

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
```diff
@@ -0,0 +1,165 @@
...
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert97InconsistentPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert97InconsistentPolicy.pem",
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "QcStmtPsd2Cert98InconsistentPolicy.pem",
+			InputFilename:  "QcStmtPsd2Cert98InconsistentPolicy.pem",
```
sleevi, line 127: It would be useful to name 96-98 to highlight their uniqueness

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
mtgag: I will remove this code part. This is already covered and caught by 
e_sub_cert_cert_policy_empty
e_sub_cert_certificate_policies_missing

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
mtgag: Done.

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
mtgag: Done.

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
mtgag: Done. Added comment to each certificate.

v2/lints/etsi/lint_policy_and_qctype_are_consistent_test.go
mtgag: Removed this test. This is covered by other lints.

v2/lints/etsi/lint_policy_and_qctype_are_consistent.go
mtgag: Implemented suggestion.

v2/util/qc_stmt.go
```diff
@@ -117,6 +119,76 @@ func checkAsn1Reencoding(i interface{}, originalEncoding []byte, appendIfCompari
...
+	if len(s.GetErrorInfo()) != 0 {
+		return nil
+	}
+	if !s.IsPresent() {
+		return result
+	}
+	qcType := s.(Etsi423QcType)
+	result = append(result, qcType.TypeOids...)
+	return result
+}
```
sleevi, line 138: Should we be ensuring these new functions use the new ZCrypto QCStatement parsing? I suspect longer-term, most of this file will be able to go away now that ZCrypto landed support in https://github.com/zmap/zcrypto/pull/217 (which was added to zlint in https://github.com/zmap/zlint/pull/435 ), so I'm trying to figure out the best way to do that.

**Option 1**: Obsolete/update qc_stmt.go in favor of the ZCrypto parsing, and then continue
**Option 2**: Incrementally adjust for this lint, and then later finish the rest of the conversion.

@cpu @dadrian Do you have thoughts here? I'm slightly inclined to **Option 2** to keep forward momentum, but I realize that leaves the file in an inconsistent state.

Based on the ZCrypto change, my understanding for Option 2 would be that this would just be
```suggestion
func GetEtsiQcTypes(c *x509.Certificate) []asn1.ObjectIdentifier {
  var result []asn1.ObjectIdentifier
  for _, qcTypeStmt := range c.QCStatements.ParsedStatements.Types {
    result = append(result, qcTypeStmt.TypeIdentifiers...)
   }
   return result
}
```

However, I'd want to hear from the maintainers before going that route. And there's probably some 'nil' handling needed for both QCStatements and ParsedStatements ;)

v2/util/qc_stmt.go
```diff
@@ -117,6 +119,76 @@ func checkAsn1Reencoding(i interface{}, originalEncoding []byte, appendIfCompari
...
+
+func HasCertAnyEtsiQcStatement(c *x509.Certificate) bool {
+	ext := GetExtFromCert(c, QcStateOid)
+	if ext == nil {
+		return false
+	}
+	return IsAnyEtsiQcStatementPresent(ext.Value)
+}
```
sleevi, line 146: If/when IsAnyEtsiQcStatementPresent is updated (e.g. to take an x509.Certificate) this function wouldn't be needed.

It could just do a case over c.QCStatements.StatementIDs

v2/util/qc_stmt.go
sleevi: I'm not sure how I missed https://github.com/zmap/zlint/issues/434 but yeah, that's what I'm flagging here.

Should that be resolved first is, I think, the question at hand :)

cpu:

PR #389, opened 2020-02-12
Psd2 roles

mtgag:
implementation of a lint to check correct roles for PSD2


lints/etsi/lint_qcstatem_psd2_roles.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+func isValidPSD2Role(oid *asn1.ObjectIdentifier, name *string) bool {
+	return ((oid.Equal(util.IdEtsiPsd2RolePspAs) && *name == "PSP_AS") ||
+		(oid.Equal(util.IdEtsiPsd2RolePspPi) && *name == "PSP_PI") ||
+		(oid.Equal(util.IdEtsiPsd2RolePspAi) && *name == "PSP_AI") ||
+		(oid.Equal(util.IdEtsiPsd2RolePspIc) && *name == "PSP_IC"))
+}
+
+func (l *qcStatemPsd2Roles) Execute(c *x509.Certificate) *lint.LintResult {
```
sleevi: @dadrian / @cpu : Structurally, could you provide guidance on what granularity of lints is expected?

In [TS 119 495](https://www.etsi.org/deliver/etsi_ts/119400_119499/119495/01.04.01_60/ts_119495v010401p.pdf), there are three discrete requirements regarding roles:
* A role must be present (GEN-5.1-2, GEN-5.2.2-1)
* The role must be one of the four OIDs defined (GEN-5.2.2-2)
* The role name must match the abbreviated form for the role OID (GEN-5.2.2-3, REG-5.2.2-5)

This PR currently checks the latter two requirements together, which is not unreasonable, considering there's a 1:1 mapping with OID and name.

However, I wasn't sure whether that was OK, or whether it made sense to write these as three discrete lints, given the three discrete technical requirements. Thoughts?

sleevi: Oh, and I should mention, this is a stand-in since it applies to several other PRs (As somewhat documented in #410 ), which also touch on fields with multiple discrete requirements. This was the smallest one I could find, so I thought it'd be easy to check with this first, as if we split this, then we should also split up any other larger one, presumably.

cpu:

Issue #392, opened 2020-02-12
Fix module incompatibility for v2.0.0-rc1


cpu:
We neglected to consider the Go module semantic import requirements ([docs](https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning)) and the `v2.0.0-rc1` tag (and the code w/ breaking API changes in master) aren't compatible. E.g. importing to CFSSL gives:

```
~/go/src/github.com/cloudflare/cfssl$ go get github.com/zmap/zlint@v2.0.0-rc1
go: finding github.com/zmap/zlint v2.0.0-rc1
go: finding github.com/zmap/zlint v2.0.0-rc1
go get github.com/zmap/zlint@v2.0.0-rc1: github.com/zmap/zlint@v2.0.0-rc1: invalid version: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

~I'm working on figuring out the best path forward. Likely it will involve creating a `v2` directory in the root and moving the new backwards incompat code there.~ I really hate duplicating code in-tree and we don't intend to support the v1 codebase. Since the README specifies we require Go 1.13+ I think we can use the "Major branch" approach to a v2 release.

Will address before finalizing 2.0.0.

cpu:
Resolved by https://github.com/zmap/zlint/commit/fd40f579253ea1ebfb18a585ab5cd8e7dcde61aa and published as [`v2.0.0-rc4`](https://github.com/zmap/zlint/releases/tag/v2.0.0-rc4)

PR #393, opened 2020-02-12
project: switch to v2 import/module path.

cpu:
This fixes the project's module compatibility w.r.t releasing module code version 2 or higher. 

We're following the "Major branch" approach described in the [docs](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher
) and will accept that this may break compat for users of old (unsupported) Go versions that `go get -u`.

Imports of `"github.com/zmap/zlint/v2"` in downstream code will get the code in master with backwards incompat changes. Existing imports of `"github.com/zmap/zlint"` will get v1.1.0 indefinitely (or until there is a newer minor or point release in the v1 stream). I tested briefly with some forks.

Resolves https://github.com/zmap/zlint/issues/392


zakird:
I don't have a good sense if this is the right way to do this in Go. It's kinda ugly but that might just be how Go is. Will defer to @dadrian .

cpu:
> It's kinda ugly but that might just be how Go is.

Agreed :-( My understanding is changing the import paths to add the major version is unavoidable and we just have to decide on whether this approach "major branch" is the best path forward (my feeling, and Jsha's) or whether we should pursue the "Major subdirectory" approach. I experimented a bit with the latter and couldn't get it working (likely my error).

> Will defer to @dadrian .

WFM, thanks!

cpu:
> or whether we should pursue the "Major subdirectory" approach. I experimented a bit with the latter and couldn't get it working (likely my error).

@dadrian figured out what I was doing wrong (_hint: simple import path errors and some misunderstandings about where the `go` command needs to be run relative to the mod :man_facepalming:_). 

The PR desc in https://github.com/zmap/zlint/pull/398 makes a compelling case for why we should use the major subdirectory approach. I'm going to close my PR in favour of helping land 398.

Issue #383, opened 2020-02-10
Update newLint.sh and template


cpu:
Now that lints are split up inside of `lints/` it probably makes sense to update `newLint.sh` to ask what folder the new lint should be created in.

cpu:
Resolved by #397

PR #397, opened 2020-02-13
Update newLint.sh script and contributing guide. 

pkoro:
#383

### DIFF
--- a/newLint.sh
+++ b/newLint.sh
@@ -1,9 +1,10 @@
 # Script to create new lint from template
 
-USAGE="Usage: $0 <ARG1> <ARG2>
+USAGE="Usage: $0 <ARG1> <ARG2> <ARG3>
 
-ARG1: File_name/TestName (no 'lint_' prefix)
-ARG2: Struct_name"
+ARG1: Path_name
+ARG2: File_name/TestName (no 'lint_' prefix)
+ARG3: Struct_name"
 
 if [ $# -eq 0 ]; then
     echo "No arguments provided..."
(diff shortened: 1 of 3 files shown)


newLint.sh
```diff
@@ -17,18 +18,32 @@ if [ $# -eq 1 ]; then
...
+if [ $# -eq 2 ]; then
+    echo "Not enough arguments provided..."
+    echo "$USAGE"
+    exit 1
+fi
+
+if [ ! -d lints/$1 ]
 then
-   echo "File already exists. Can't make new file."
+   echo "Directory $1 does not exist. Can't make new file."
```
cpu:
```suggestion
   echo "Directory 'lints/$1' does not exist. Can't make new file."
```

template
```diff
@@ -12,7 +12,7 @@
  * permissions and limitations under the License.
  */
 
-package lints
+package PACKAGE
 
 import (
 	"github.com/zmap/zcrypto/x509"
```
cpu: Would you mind also adding the missing `github.com/zmap/zlint/lint` import to the template?

newLint.sh
```diff
@@ -17,18 +18,32 @@ if [ $# -eq 1 ]; then
...
-cp template lints/lint_$FILENAME.go
+if [ -e lints/$1/lint_$2.go ]
+then
+   echo "File already exists. Can't make new file."
+   exit 1
+fi
+
+PATHNAME=$1
+LINTNAME=$2
+FILENAME=${LINTNAME:2}
```
cpu: Am I correct in understanding the `LINTNAME`/`FILENAME` split here is being used to remove the `e_`, `w_` etc prefix from lint filenames? If so can you add a comment to that effect? I think it's a little bit subtle as is.

newLint.sh
pkoro: Done!

newLint.sh
pkoro: Yes you are right. I added one comment regarding this.

PR #398, opened 2020-02-13
Fix v2 with go.mod

dadrian:
Adopt the **Major Subdirectory** approach from
https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher

### Effects:

* Moves all code into a `v2` subdirectory. Consumers of ZLint v2 will need to change their imports in go to `github.com/zmap/zlint/v2` and update `go.mod` accordingly.
* Old versions of ZLint are still fetchable via `go.mod` and version pinning, e.g. `require github.com/zmap/zlint@1.1.0`
* People using `go get` without modules are going to have all their code break. However, this was going to happen regardless of what we do with our directories because we made breaking changes to our code.

### Patching old versions

To patch an old version (pre-2.0), we would need to branch off of one of the old tags. To avoid this, we could attempt to maintain support for 1.1.0 by implementing another point release at the top-level directory on top of `v2`, and then exposing the old API's above the v2 directory (see https://github.com/rsc/quote/blob/master/quote.go as an example). However, I don't believe we plan on supporting old versions at all, so there's no reason to do this. On the off chance we do need to cut a point release for 1.1.0, we can use Git.

### Why not use the Major Branch approach

When not maintaining support for v1 side-by-side with v2, the major branch approach is identical to the Major Subdirectory approach from a Git standpoint---point release for v1 would need to be done on branches. However, it does then have the side effect that the import path for your code no longer matches the directory structure of your code. Effectively, we could `mv` all of the `v2` directory back up to the top-level, claim the name `...v2/`, and use `v2` in all of our import paths. While this might look cleaner from a Git repo standpoint, my gut sense is that we may as well match the directories to the import path, since that's slightly easier to grok.


dadrian:
You should be able to import this code in a module by running `go get github.com/zmap/zlint/v2@dadrian-v2.0.0-RC2` in a repo with Go modules enabled. 

Example `go.mod`:
```go
module dadrian.io/test-zlint

go 1.13

require github.com/zmap/zlint/v2 v2.0.0-rc1.0.20200213173722-6ee8eead21f6
```
Example `main.go`
```go
package main

import (
	"fmt"
	"github.com/zmap/zlint/v2/lint"
)

func main() {
	registry := lint.GlobalRegistry()
	fmt.Println(registry)
}

**Comment by dadrian** [comment]
Note: CI still needs to be updated to reflect the fact all the files it expected to be in the root of the repo are now in `v2`

**Comment by cpu** [comment]
> Why not use the Major Branch approach

:+1: I'm convinced. I think it will also make it easier to work on a `v3` in parallel like we probably 
 should have done for `v2`.

Thanks for figuring this out!

File: .gitignore [hunk 1]
```diff
@@ -114,10 +114,10 @@ _testmain.go
...
-cmd/zlint/zlint
-/zlint-gtld-update
-cmd/zlint-gtld-update/zlint-gtld-update
+v2/zlint
+v2/cmd/zlint/zlint
+v2/zlint-gtld-update
+v2/cmd/zlint-gtld-update/zlint-gtld-update
 
 ### Integration test data ###
 data
```
cpu:
```suggestion
v2/data
```

cpu: I think `dist` needs the same treatment -> `v2/dist`.

cpu:
@dadrian One other thing I thought of: I think the Goreleaser config needs to be told to use the subdir:

```
diff --git .goreleaser.yml .goreleaser.yml
index 2b84be0..c303c31 100644
--- .goreleaser.yml
+++ .goreleaser.yml
@@ -4,6 +4,7 @@ before:
     - go mod tidy
 builds:
   -
+    dir: v2
     main: ./cmd/zlint/main.go
     binary: zlint
     env:
```

After this lands I'll cut a v2.0.0-rc2 to test definitively but I suspect without the `dir` it will break.

PR #399, opened 2020-02-14
Ev orgideext present recommended

mtgag:
implementation of a lint to check mandatory presence of CAB/F OI extension if subject:OI is present


mtgag:
Closing this pull request because this is already implemented in lint lint_ev_organization_id_missing.go.

Issue #413, opened 2020-02-27
Lint for nameConstraints for dNSName and iPAddress


sleevi:
Presently, there are not lints on the well-formedness of the dNSName and iPAddress nameConstraints. There are lints for other name types, such as https://github.com/zmap/zlint/blob/fd40f579253ea1ebfb18a585ab5cd8e7dcde61aa/v2/lints/rfc/lint_name_constraint_on_edi_party_name.go or https://github.com/zmap/zlint/blob/fd40f579253ea1ebfb18a585ab5cd8e7dcde61aa/v2/lints/rfc/lint_name_constraint_on_x400.go , but none for dNSName or iPAddress.

Broadly, many of the dNSName checks applied to the SAN would also apply to the `nameConstraints`.

For example, see [this discussion in the CA/B Forum](https://cabforum.org/pipermail/servercert-wg/2020-February/001674.html), which pointed out that https://github.com/awslabs/certlint correctly [detects the malformed name constraint](https://crt.sh/?id=2929505&opt=cablint) on https://crt.sh/?sha256=A56C1BD2C47FFC701023FAD8F65CBE794946B779475848DCA30760A06DD72FD9 but ZLint [does not](https://crt.sh/?id=2929505&opt=zlint)

sleevi:
Erp, I missed the similarity to #136 due to also mentioning iPAddresses here.

I'm ok duping them, or keeping this separate to also lint the well-formedness of IP addresses.

rolandshoemaker:
Looks like `zmap/zcrypto/x509` is already enforcing IP constraint validity [during parsing](https://github.com/zmap/zcrypto/blob/master/x509/x509.go#L1655-L1665).

Stdlib `crypto/x509` does its own DNS name validity check, but [disagrees that leading periods are invalid](https://golang.org/src/crypto/x509/x509.go#L1245). See https://github.com/golang/go/issues/16347 for discussion of why this was allowed in stdlib.

sleevi:
@rolandshoemaker Thanks for pointing out the iPAddress. While that makes sure that the base value is parsable, I was thinking about other checks, such as making sure it's not an internal IP address. I filed https://github.com/cabforum/documents/issues/160 so that these requirements could be made more explicit, where they are presently implicit.

That said, the Golang change seems to be about compatibility with existing misissuance, while given ZLint's concern on correctness, seems not to be too compelling. That said, I filed https://www.rfc-editor.org/errata/eid5997 to try and clarify.

[...]

PR #414, opened 2020-03-04
lints: disallow reserved iPAddresses in NCs

rolandshoemaker:
Addresses the IP portion of #413. Given it should be impossible for an applicant to display control of an IANA reserved range, any NC with an iPAddress subtree that intersects any of these ranges should be considered invalid.

Been a while since I spent any time looking at zlint, so entirely possible I've done something stupid here.


v2/util/ip.go
```diff
@@ -73,6 +73,19 @@ func IsIANAReserved(ip net.IP) bool {
 	return false
 }
 
+// IntersectsIANAReserved checks if a CIDR intersects any IANA reserved CIDRs
+func IntersectsIANAReserved(net net.IPNet) bool {
+	if !net.IP.IsGlobalUnicast() {
+		return true
+	}
+	for _, reserved := range reservedNetworks {
+		if reserved.Contains(net.IP) || net.Contains(reserved.IP) {
```
zakird: Does this handle the case where `net.IP == reserved.IP`?

v2/lints/cabf_br/lint_ext_nc_intersects_reserved_ip.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type NCReservedIPNet struct{}
+
+func (l *NCReservedIPNet) Initialize() error {
+	return nil
+}
+
+func (l *NCReservedIPNet) CheckApplies(c *x509.Certificate) bool {
+	return c.NotAfter.After(util.NoReservedIP) && util.IsExtInCert(c, util.NameConstOID)
```
zakird: Is this the date after which CAs cannot issue certificates or the date after which certificates can no longer contain reserved IPs?

v2/util/ip.go
rolandshoemaker: Any two networks that have the same base number should contain each other, [example](https://play.golang.org/p/tO-H4Mr6cWs).

v2/lints/cabf_br/lint_ext_nc_intersects_reserved_ip.go
rolandshoemaker: The latter I believe, I wasn't entirely sure if it's the best choice, but I thought it made sense to keep in line with the `e_ext_san_contains_reserved_ip` lint.

PR #415, opened 2020-03-04
remove lisp reserved range since no longer IANA reserved

zakird:
lisp range is no longer reserved here: https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,21 +5,14 @@ ZLint
 [![Go Report Card](https://goreportcard.com/badge/github.com/zmap/zlint)](https://goreportcard.com/report/github.com/zmap/zlint)
 
 ZLint is a X.509 certificate linter written in Go that checks for consistency
-with standards (e.g. [RFC 5280]) and other relevant PKI requirements (e.g.
-[CA/Browser Forum Baseline Requirements][BR v1.4.8]).
-
-It can be used as a command line tool or as a library integrated into CA
-software.
+with standards (e.g. [RFC 5280]) and PKI requirements (e.g., [CA/Browser Forum
+Baseline Requirements][BR v1.4.8]). It can be used as a command line tool or as
+a library integrated into CA software. [ZLint
+v2.0.0](https://github.com/zmap/zlint/tree/v2.0.0) is the latest stable
+release.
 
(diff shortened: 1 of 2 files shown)


zakird:
When I was looking at this list I noticed that the loopback ranges`::1/128`, `::/128`, and `127.0.0.1/8` aren't in this set of blocks. @sleevi these cannot be used, right? @cpu do we have a lint elsewhere for these? I'm not finding one, but I feel like this has to be me missing something.

sleevi:
> When I was looking at this list I noticed that the loopback ranges`::1/128`, `::/128`, and `127.0.0.1/8` aren't in this set of blocks. @sleevi these cannot be used, right? @cpu do we have a lint elsewhere for these? I'm not finding one, but I feel like this has to be me missing something.

Correct. They're reserved by protocol, so should not be allowed (e.g. https://tools.ietf.org/html/rfc4291#section-2.5.3 )

zakird:
Thanks. Once this is merged, I'll add those in if there's truly no loopback lint already.

cpu:
>  @cpu do we have a lint elsewhere for these? I'm not finding one, but I feel like this has to be me missing something.

@zakird These addresses seem to be caught by the call to `ip.IsGlobalUnicast()` at the start of `util.IsIANAReserved`:
https://github.com/zmap/zlint/blob/3329bb69d206198706abc7c84eccb59011bd9de5/v2/util/ip.go#L63-L65

I confirmed this with a tiny unit test:

```go
// util/ip_test.go 
package util

import (
	"net"
	"testing"
)

func TestBadIP(t *testing.T) {
	if !IsIANAReserved(net.ParseIP("127.0.0.1")) {
		t.Error("loopback address was not considered reserved")
	}
	if !IsIANAReserved(net.ParseIP("::1/128")) {
		t.Error("loopback address was not considered reserved")
	}
}
```

```bash
$ go test --test.run TestBadIP -v ./util/
=== RUN   TestBadIP
--- PASS: TestBadIP (0.00s)
PASS
ok  	github.com/zmap/zlint/v2/util	0.004s
```

It probably makes sense to make this more explicit in the code (with a comment, or by adding the ranges explicitly to `reservedNetworks`.

PR #421, opened 2020-03-17
util: rewrite test/prepend_testcerts_openssl.sh, update testdata

cpu:
The old version of `test/prepend_testcerts_openssl.sh` needed to be run from a specific directory, didn't pass `shellcheck`, and would unconditionally prepend the OpenSSL text output to all certs in the testdata dir (even if they already had it).

The version in 04c9d99 should be safer and suitable to be integrated with CI. It also supports taking a glob as the first argument and only prepending certs that need it and have a filename that matches the glob. I kept it in BASH rather than rewriting it in Go since we're shelling out to `openssl` and I find that's fiddly work in Go and the overall complexity of the script is low.

b250164 contains the results of running the script against the full `testdata` directory. Many testcerts were missing OpenSSL text output and this commit fixes them across the board.

06f8b73 is the one exception where `openssl` fails to parse the certificate. For now I've added a note about this to the `.pem` file such that the `prepend_testcerts_openssl.sh` script no longer emits a warning. I haven't done any investigation work into the testcert/testcase and why it fails to parse.

db871ab updates CI to run the `test/prepend_testcerts_openssl.sh` script for each branch's build and exit with an error code if there are diffs. This will make sure we don't introduce any new test certs after b250164 that forget to include some text before the PEM contents.

### DIFF
--- a/v2/test/prepend_testcerts_openssl.sh
+++ b/v2/test/prepend_testcerts_openssl.sh
@@ -34,7 +34,6 @@ for f in "$CERTS_DIR"/*.pem; do
   fi
 
   # Prepend the test cert with its -text OpenSSL output.
-  openssl x509 -in "$f" -text -noout | \
-    cat - "$f" > "$TMP_DIR/$CERT_NAME.new" && \
-    mv "$TMP_DIR/$CERT_NAME.new" "$f"
+  openssl x509 -text -in "$f" -outform PEM -out "$TMP_DIR/$CERT_NAME.new" \
+    && mv "$TMP_DIR/$CERT_NAME.new" "$f"
 done
--- a/v2/test/prepend_testcerts_openssl.sh
+++ b/v2/test/prepend_testcerts_openssl.sh
@@ -1,4 +1,40 @@
 #!/bin/bash
-for f in ../testlint/testCerts/*; do
-    openssl x509 -in $f -text -noout | cat - $f > /tmp/out && mv /tmp/out $f
+
+set -e -o pipefail
+
+BASE_DIR=$(dirname "$0")
+CERTS_DIR="$BASE_DIR/../testdata"
+TMP_DIR=$(mktemp -d -t zlint-XXXX)
+
+# Trap EXIT to cleanup the TMP_DIR
+trap '{ rmdir --ignore-fail-on-non-empty $TMP_DIR; }' EXIT
+
+# For every .pem file in the $CERTS directory, prepend 0penSSL text output if
+# required.
+for f in "$CERTS_DIR"/*.pem; do
+  # Skip any files that don't begin with a PEM header. These are assumed to
+  # already have the OpenSSL text output prepended.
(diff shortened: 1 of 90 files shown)


v2/test/prepend_testcerts_openssl.sh
```diff
@@ -1,4 +1,40 @@
...
+  # If the certificate has errors parsing with OpenSSL print a warning to stderr
+  # and continue. Sometimes our test data is too weird to parse and that's OK.
+  if ! openssl x509 -in "$f" -noout || false; then
+    echo "error parsing $f with OpenSSL" >&2
+    continue
+  fi
+
+  # Prepend the test cert with its -text OpenSSL output.
+  openssl x509 -in "$f" -text -noout | \
+    cat - "$f" > "$TMP_DIR/$CERT_NAME.new" && \
```
sleevi: Why not just

`openssl x509 -text -in "$f" -outform PEM -out "$TMP_DIR/$CERT_NAME.new" && mv "$TMP_DIR/$CERT_NAME.new" "$f"`

Am I missing an error condition that this covers?

v2/test/prepend_testcerts_openssl.sh
```diff
@@ -1,4 +1,40 @@
...
+
+# Trap EXIT to cleanup the TMP_DIR
+trap '{ rmdir --ignore-fail-on-non-empty $TMP_DIR; }' EXIT
+
+# For every .pem file in the $CERTS directory, prepend 0penSSL text output if
+# required.
+for f in "$CERTS_DIR"/*.pem; do
+  # Skip any files that don't begin with a PEM header. These are assumed to
+  # already have the OpenSSL text output prepended.
+  if [[ ! $(head -n1 "$f") =~ "-----BEGIN" ]]; then
```
sleevi: Just noting:

I was originally going to suggest it might be useful to have a --force option or the like, to force the header to be recomputed even for existing items, if/as more fields are supported (e.g. to make it more useful)

That said, I think I talked myself out of it, because the simplicity of line 25 is worth preserving, and if/when that happens, this 'could' just be a simple bash loop to do the same. That is, it's more likely to be a one-off, so there's no point in building that functionality into this tool and complicating this tool.

v2/test/prepend_testcerts_openssl.sh
cpu: Good question :thinking: I cribbed this from the existing script without thinking about it too hard. I tried your approach using `-out` and `-outform` on a branch with b250164 reverted and it looks like it operates the same way.

Switched to your version in 25b5f4a, thanks!

v2/test/prepend_testcerts_openssl.sh
cpu: Good point about wanting to update the text in the future. I'm also happy to defer that to the day when the need presents itself.

cpu:
Since this is test-only I'm going to go ahead and merge it. Thanks!

PR #426, opened 2020-04-02
gTLD map: capture 2020-04-01 removal of xn--estv75g

ezekiel:


cpu:
:wave: Thanks for the PR, but it's a duplicate of https://github.com/zmap/zlint/pull/425

cpu:
One other thought: I think it would be better to let the @tld-update-bot open these PRs. That way we know for sure that no transliteration errors occurred between the upstream JSON and the committed diff. The bot runs every 24hrs at aproximately 1pm eastern time but if there's a situation that requires the new data be pulled in faster I think it would be better to open an issue instead of a PR and a maintainer can kick the bot. It's not likely to be any slower than opening a PR manually because those still require a maintainer to merge.

Hope that makes sense, thanks again for the PR!

Issue #368, opened 2020-01-22
Add lints for DirectoryString encoding as Printable or UTF8


jsha:
https://tools.ietf.org/html/rfc5280#page-24

   When encoding attribute values of type DirectoryString, conforming
   CAs MUST use PrintableString or UTF8String encoding, with the
   following exceptions:

      (a)  When the subject of the certificate is a CA, the subject
           field MUST be encoded in the same way as it is encoded in the
           issuer field (Section 4.1.2.4) in all certificates issued by
           the subject CA.  Thus, if the subject CA encodes attributes
           in the issuer fields of certificates that it issues using the
           TeletexString, BMPString, or UniversalString encodings, then
           the subject field of certificates issued to that CA MUST use
           the same encoding.

      (b)  When the subject of the certificate is a CRL issuer, the
           subject field MUST be encoded in the same way as it is
           encoded in the issuer field (Section 5.1.2.3) in all CRLs
           issued by the subject CRL issuer.

      (c)  TeletexString, BMPString, and UniversalString are included
           for backward compatibility, and SHOULD NOT be used for
           certificates for new subjects.  However, these types MAY be
           used in certificates where the name was previously
           established, including cases in which a new certificate is
           being issued to an existing subject or a certificate is being
           issued to a new subject where the attributes being encoded
           have been previously established in certificates issued to
           other subjects.  Certificate users SHOULD be prepared to
           receive certificates with these types.

Clause (c) makes it hard to make this an error or warning, since we can't know if a name had previously been established. Perhaps this could be a notice-level lint?

PR #432, opened 2020-05-13
ci: remove vendor dir, Go 1.13.x -> 1.14.x, fix integration test data

cpu:
## chore: remove vendor dir.

Vendoring has lost favour compared to relying on the Go 1.13+ [proxy/module checksum behaviour](https://proxy.golang.org/).

## ci: go 1.13.x -> go1.14.x

Switch to the latest stable Go release stream.  Also remove setting `GO111MODULE` and `GOFLAGS` in the makefile env. The former is already the default since Go 1.12.x and the latter isn't required because we removed the vendor dir.

## ci: update expected integration test data for Go 1.14.

New lints without result cases in our integration test data (e.g. `e_ext_nc_intersects_reserved_ip`, `e_mp_rsassa-pss_in_spki`) are added with the expected set `{}`.

Four existing lints have their expected error result tallies updated:

1. "e_sub_cert_locality_name_must_not_appear"
  old: fatals: 0    errs: 23   warns: 0    infos: 0
  new: fatals: 0    errs: 13   warns: 0    infos: 0

2. "e_sub_cert_province_must_not_appear"
  old: fatals: 0    errs: 16   warns: 0    infos: 0
  new: fatals: 0    errs: 8    warns: 0    infos: 0

3. "e_sub_cert_street_address_should_not_exist"
  old: fatals: 0    errs: 8    warns: 0    infos: 0
  new: fatals: 0    errs: 0    warns: 0    infos: 0

4. "e_sub_cert_postal_code_must_not_appear"
  old: fatals: 0    errs: 8    warns: 0    infos: 0
  new: fatals: 0    errs: 0    warns: 0    infos: 0

These four lints previously returned an error result for certificates that had a Subject Organization/GivenName/Surname that were encoded as a BMPString. Go < 1.14.x's ASN.1 package did not support this encoding type (See https://github.com/golang/go/pull/26690) and so the lints assumed the field was absent, resulting in a false positive. In Go 1.14.x+ the field is correctly decoded and the error result is no longer applicable.

As one concrete example consider the certificate with fingerprint [`008082596e15c4a1092b4c1ca44fb7d600669116d15f8bf04730b77b5630f4e9`](https://censys.io/certificates/008082596e15c4a1092b4c1ca44fb7d600669116d15f8bf04730b77b5630f4e9/openssl) and the `e_sub_cert_locality_name_must_not_appear` lint. This certificate has a subject organization name encoded as a BMPSTRING, and all other subject fields encoded as PRINTABLESTRING.

With ZLint built with Go 1.13.x this results in the `cert.Subject.Organization` field being considered empty but the `cert.Subject.Locality` is not. The `e_sub_cert_locality_name_must_not_appear` lint thus finds a non-empty `cert.Subject.Locality` but no `cert.Subject.Organization`, `cert.Subject.GivenName` or `cert.Subject.Surname` and an error result result is (incorrectly) returned:
https://github.com/zmap/zlint/blob/bb6c7a74f1901fecf545c51af97e9e874a0260d9/v2/lints/cabf_br/lint_sub_cert_locality_name_must_not_appear.go#L34-L38

With ZLint built with Go 1.14.x the `cert.Subject.Organization` field encoding is supported and so the decoded value is not empty. This in turn short-circuits the conditional in the `e_sub_cert_locality_name_must_not_appear` lint and a pass result is returned.


dadrian:
> In Go 1.14.x+ the field is correctly decoded and the error result is no longer applicable.

Always good to see results getting "more correct!". Do we have a lint that checks for BMPString encoding? I'm not sure what the BR's say about it, but it might be worth at least a `notice`.

cpu:
Thanks for the quick :mag:'s!

cpu:
> Do we have a lint that checks for BMPString encoding? I'm not sure what the BR's say about it, but it might be worth at least a notice.

I don't think there is one. Legacy BMPString encoding is mentioned as a SHOULD NOT in https://github.com/zmap/zlint/issues/368 within the context of DirectoryString encoding. I'm not sure off-hand if there are other contexts to lint for this string encoding.

sleevi:
@dadrian https://tools.ietf.org/html/rfc5280#section-4.1.2.6

>  When encoding attribute values of type DirectoryString, conforming
>   CAs MUST use PrintableString or UTF8String encoding, with the
>   following exceptions:

(a) applies to CA
(b) applies to CAs (CRL issuers)

>       (c)  TeletexString, BMPString, and UniversalString are included
>           for backward compatibility, and SHOULD NOT be used for
>           certificates for new subjects.  However, these types MAY be
>           used in certificates where the name was previously
>           established, including cases in which a new certificate is
>           being issued to an existing subject or a certificate is being
>           issued to a new subject where the attributes being encoded
>           have been previously established in certificates issued to
>           other subjects.  Certificate users SHOULD be prepared to
>           receive certificates with these types.

PR #436, opened 2020-05-24
Check tbsCertificate signature algorithm matches certificate

### LINKED ISSUE #419: Lint the signature field of a tbsCertificate matches the signatureAlgorithm field of a certificate
RFC 5280, [Section 4.1.1.2](https://tools.ietf.org/html/rfc5280#section-4.1.1.2) states:
```
   This field MUST contain the same algorithm identifier as the
   signature field in the sequence tbsCertificate (Section 4.1.2.3).
```

While the outer `signatureAlgorithm` is fully malleable to attackers, adding a lint to ensure the consistency between the inner and outer algorithms would align with RFC 5280.

This is most likely blocked on https://github.com/zmap/zcrypto/issues/213 to expose this information.

rolandshoemaker:
Per RFC 5280 section 4.1.1.2, couldn't find an existing lint.

Resolves #419


rolandshoemaker:
Just to add a little context after some discussions elsewhere, this lint is probably of value to a CA that is using zlint, since it may catch creating a certificate that is somewhat... dubious post signature, but is likely of significantly less value as a lint for larger corpora, like CT, since the certificate signatureAlgorithm field is not covered by the signature and is as such malleable. This means anyone can submit a bunch of doppelgänger certificates with an altered algorithmIdentifier sequence which would trigger this lint (an example being the 10 certs caught using the corpus integration tests, which all appear to be altered versions of their original certificates.

I'd probably be find abandoning this change given this...

cpu:
Hey @rolandshoemaker! Thanks for the PR.

I haven't had a chance to review it yet but wanted to leave a comment.

> I'd probably be find abandoning this change given this...

I think it's worth keeping if the tradeoffs you mention are documented (in code comments or elsewhere). We actually have an outstanding issue for this lint that @sleevi filed: https://github.com/zmap/zlint/issues/419 I think we were favouring an approach that would put more of the ASN.1 work into ZCrypto (https://github.com/zmap/zcrypto/issues/213) WDYT?

rolandshoemaker:
Ah interesting, I completely missed that issue.

In terms of exposing the signature that seems viable, but I'm not sure it's what you'd entirely want for this particular lint. Using `asn1.Unmarshal` means throwing away extraneous elements at the end of the SEQUENCE, so for instance an algorithmIdentifier with a trailing VideotextString will _look_ just like one without, but won't actually be byte-for-byte equal. We could probably get access to those bytes, but it'd require adding a handful of `asn1.RawContent` and alias types all over the place.

v2/lints/rfc/lint_tbs_signature_alg_matches_cert_signature_alg.go
```diff
@@ -0,0 +1,92 @@
...
+		return &lint.LintResult{Status: lint.Fatal, Details: "error reading certificate.signatureAlgorithm"}
+	}
+
+	// parse out tbsCertificate signature
+	input = cryptobyte.String(c.RawTBSCertificate)
+	var tbsCert cryptobyte.String
+	if !input.ReadASN1(&tbsCert, cryptobyte_asn1.SEQUENCE) {
+		return &lint.LintResult{Status: lint.Fatal, Details: "error reading tbsCertificate"}
+	}
+	if !tbsCert.SkipOptionalASN1(cryptobyte_asn1.Tag(0).Constructed().ContextSpecific()) {
```
cpu, line 65: small nit: maybe worth pulling out `cryptobyte_asn1.Tag(0).Constructed().ContextSpecific()` as an unexported package level `var` to avoid needing to construct it each time?

cpu: I think after another bit of thought it would be less readable.

v2/lints/rfc/lint_tbs_signature_alg_matches_cert_signature_alg.go
```diff
@@ -0,0 +1,92 @@
...
+		return &lint.LintResult{Status: lint.Fatal, Details: "error reading certificate.tbsCertificate"}
+	}
+	var certSigAlg cryptobyte.String
+	if !cert.ReadASN1(&certSigAlg, cryptobyte_asn1.SEQUENCE) {
+		return &lint.LintResult{Status: lint.Fatal, Details: "error reading certificate.signatureAlgorithm"}
+	}
+
+	// parse out tbsCertificate signature
+	input = cryptobyte.String(c.RawTBSCertificate)
+	var tbsCert cryptobyte.String
```
sleevi, line 61: Why not just stash tbsCert on line 51? That would avoid the cryptobyte.String() conversion over RawTBSCertificate

rolandshoemaker:
Thanks for the reviews, think I hit everything.

Issue #384, opened 2020-02-11
Allow filtering ResultSet by Status threshold


cpu:
Originally requested by @cardonator in https://github.com/zmap/zlint/pull/372#issuecomment-580441107 I had hoped to land a feature for this with #372 but ran out of time and don't want to block the RC tag.

Filtering lint results to only include those with a status above a configured threshold is a common requirement for downstream users of `zlint`, e.g. the `cfssl` integration supports this operation: https://github.com/cloudflare/cfssl/blob/644917271238216c94866f021e2b24ce54555848/signer/local/local.go#L183-L185

Rather than duplicate equivalent code in consumers I think it makes sense for ZLint to provide the feature. Personally I think it makes the most sense to implement this as a method on the `ResultSet` type (vs as another arg to `zlint.LintCertificateEx`) since it's a property of results and not lints being run. 

I had fiddled with a bit of code to do that in a discarded refactor branch: https://github.com/cpu/zlint/commit/22d711099ea5e4aaedde72450d0ae9e321ff0cf6 It would need to be rebased on master, and have tests written but might be a useful starting point to resolve this issue.

PR #437, opened 2020-05-25
Create options for human-readable output formats

acaird:
Added a `-summary` and `-longsummary` options to `zlint`.

The `-summary` option produces output that looks like:
```
+-------+--------------+
| LEVEL | # OCCURANCES |
+-------+--------------+
| info  |            0 |
| warn  |            1 |
| error |            7 |
| fatal |            0 |
+-------+--------------+
```
and the `-longsummary` produces output that looks like:
```
+-------+--------------+------------------------------------------+
| LEVEL | # OCCURANCES |                 DETAILS                  |
+-------+--------------+------------------------------------------+
| info  |            0 |  -                                       |
| warn  |            1 | w_ext_san_critical_with_subject_dn       |
| error |            7 | e_cert_policy_iv_requires_country        |
|       |              | e_ca_key_cert_sign_not_set               |
|       |              | e_ca_organization_name_missing           |
|       |              | e_ca_crl_sign_not_set                    |
|       |              | e_ca_country_name_missing                |
|       |              | e_sub_ca_crl_distribution_points_missing |
|       |              | e_sub_cert_not_is_ca                     |
| fatal |            0 |  -                                       |
+-------+--------------+------------------------------------------+
```

The only significant changes to zlint was to export `statusLabelToLintStatus` from `result.go` and to pull in `tablewriter`.

Thanks for considering these changes.


cardonator:
Wow, this is awesome. Seriously. Thanks for contributing that. I have been wanting something like this for a while but haven't had any time to work on it.

zakird:
I've committed additional changes to this PR, so tagging @cpu or @dadrian to take a final pass and merge.

v2/cmd/zlint/main.go
```diff
@@ -210,3 +220,83 @@ func setLints() (lint.Registry, error) {
 
 	return lint.GlobalRegistry().Filter(filterOpts)
 }
+
+func outputSummary(zlintResult *zlint.ResultSet, longSummary bool) {
```
cpu, line 224: I think some of this summary code could be hoisted into the `zlint.ResultSet` type. I think there's some similarity here with https://github.com/zmap/zlint/issues/384 for example. The textual summary parts should stay here of course :-)

v2/go.mod
```diff
@@ -1,6 +1,7 @@
 module github.com/zmap/zlint/v2
 
 require (
+	github.com/olekukonko/tablewriter v0.0.4
```
cpu, line 4: I'm a little bit hesitant to see a third party dependency added for something that's limited in functionality to the `zlint` binary. Do you think it would be difficult to implement the table output using `fmt` and the stdlib? I might be overlooking the reason you choose this dep.

v2/cmd/zlint/main.go
cpu: It would also probably help write unit tests for the code. I think this logic is intricate enough to merit a few unit tests and it will probably be easier to test outside of the CLI app.

acaird: thanks @cpu 

I'm less comfortable messing in `zlint.ResultSet` but would give it a try if you want.

I can also add some unit tests and update this PR, it will be a little while before I have time, though.

v2/go.mod
acaird: I agree.  I picked this because it was easy, although in the end the output was simplistic enough that is does seem like overkill and not worth the risk of relying on a third-party dependancy. 

I'll make a new PR without this dependancy.

acaird:
I'm not entirely convinced that removing the tablewriter dependency is worth it, but the new output looks like:

```
cat testdata/indivValAllBad.pem | ./zlint -summary
| LEVEL | # OCCURRENCES |
+-------+---------------+
|  info |             0 |
|  warn |             1 |
| error |             7 |
| fatal |             0 |
```

and

```
cat testdata/indivValAllBad.pem | ./zlint -longsummary
| LEVEL | # OCCURRENCES |                  DETAILS                  |
+-------+---------------+-------------------------------------------+
|  info |             0 |                                        -  |
|  warn |             1 |        w_ext_san_critical_with_subject_dn |
| error |             7 |                      e_sub_cert_not_is_ca |
|       |               |                e_ca_key_cert_sign_not_set |
|       |               |  e_sub_ca_crl_distribution_points_missing |
|       |               |                     e_ca_crl_sign_not_set |
|       |               |                 e_ca_country_name_missing |
|       |               |         e_cert_policy_iv_requires_country |
|       |               |            e_ca_organization_name_missing |
| fatal |             0 |                                        -  |
```

acaird:
Tests and some consideration of moving the filtering elsewhere will come later.

Thanks for all the feedback!

v2/cmd/zlint/main.go
```diff
@@ -51,14 +55,16 @@ var ( // flags
 func init() {
 	flag.BoolVar(&listLintsJSON, "list-lints-json", false, "Print lints in JSON format, one per line")
 	flag.BoolVar(&listLintSources, "list-lints-source", false, "Print list of lint sources, one per line")
+	flag.BoolVar(&summary, "summary", false, "Prints a short human-readable summary report")
+	flag.BoolVar(&longsummary, "longsummary", false, "Prints a human-readable summary report with details")
```
cpu, line 59: small nit: I think naming the flag `longSummary` instead of `-longsummary` will better match existing options (`-nameFilter`, `-includeSources`, etc).

v2/cmd/zlint/main.go
```diff
@@ -210,3 +220,139 @@ func setLints() (lint.Registry, error) {
...
+	threshold := lint.Pass
+
+	// Make the list of lint levels that matter
+	for _, i := range lint.StatusLabelToLintStatus {
+		if i <= threshold {
+			continue
+		}
+		lintLevelsAboveThreshold[int(i)] = i
+	}
+	// Set all of the levels to 0 events to they are all displayed
```
cpu, line 240: typo: 
```suggestion
	// Set all of the levels to 0 events so they are all displayed
```

v2/cmd/zlint/main.go
```diff
@@ -210,3 +220,139 @@ func setLints() (lint.Registry, error) {
...
+	// in the -summary table
+	for _, level := range lintLevelsAboveThreshold {
+		resultCount[level] = 0
+	}
+	// Count up the number of each event
+	for key, value := range zlintResult.Results {
```
cpu, line 246: small nit: I think this loop could be better if you used more descriptive var names. Something like:

```go
name, result := range zlintResults.Results {
   ...
}
```

v2/cmd/zlint/main.go
```diff
@@ -210,3 +220,139 @@ func setLints() (lint.Registry, error) {
...
+					fmt.Printf("+\n")
+				}
+			}
+		}
+	}
+	return hlengths
+}
+
+func printTableBody(hlengths []int, lines [][]string) {
+
```
cpu, line 343: nit: extra `\n`

v2/cmd/zlint/main.go
cpu: > I'm less comfortable messing in zlint.ResultSet but would give it a try if you want.

As an alternative, what about breaking out a `struct` type that lives in this package's scope for now? I think `outputSummary` is quite long as-is and would benefit from a little bit of refactoring. It's doing three things right now: initializing some related state, displaying short output, and displaying long output. Breaking up the initialization and the display seems like a good place to start.

Off the bat I think there's a bit of related state that could be grouped into struct members: `resultCount`, `resultDetails`, `lintLevelsAboveThreshold` and `sortedLevels`. Something like:

```go
type resultsTable struct {
  resultCount map[lint.LintStatus]int
  resultDetails map[lint.lintStatus][]string
  lintLevelsAboveThreshold map[int]lint.Status
  sortedLevels []int
}
```

You could then split out all of the `map` initialization from `outputSummary` into a constructor for the `resultsTable` type, something like:
```go
func newResultsTable(threshold lint.LintResult, results *zlint.ResultSet) resultsTable {
  // init struct members here w/ loops from the `outputSummary` code
  ...
}
```

That would remove the initialization concern from `outputSummary` and let it be re-implemented as a method on the `resultsTable` type that can strictly concern itself with outputting the results:

```go
func (t resultsTable) outputSummary(longSummary bool) {
   ....
}
```

WDYT?

acaird:
@cpu thanks for the feedback, I'll clean it up and write some tests, hopefully I'll get to it all within the next week or so

v2/cmd/zlint/main.go
acaird: fixed in b3c3f69

v2/cmd/zlint/main.go
acaird: changed in 832bc81

v2/cmd/zlint/main.go
acaird: fixed in b3c3f69

v2/cmd/zlint/main.go
acaird: fixed in  b3c3f69

cpu:
:wave: @acaird just wanted to check in on this branch. Do you feel like you'll have a chance to address the last bits of feedback in the next few weeks? If you think you might not have the time I definitely understand and would be happy to adopt this PR to get it across the line. Let me know what you would prefer!

acaird:
:wave: @cpu.  I haven't given up yet!  If I can't get something worthwhile by Tuesday I'll put it up for adoption.  Sorry for the delays... :/

cpu:
>  If I can't get something worthwhile by Tuesday I'll put it up for adoption. Sorry for the delays... :/

Please don't apologize, take your time :-) I just want to make sure my feedback isn't the only thing keeping this from landed in tree. Thanks for the update!

acaird:
Writing tests in `main` is not ideal; I can keep working on it or I can split the functions and their requirements into a file that isn't `main` and test that.  Thoughts?

cpu:
> I can split the functions and their requirements into a file that isn't main and test that. Thoughts?

@acaird Splitting would be my pref. but it's lightly held. If you find it easier to keep working through main that's OK with me. I don't want to hold you up :-)

v2/formattedoutput/formattedOutput.go
```diff
@@ -0,0 +1,159 @@
...
+)
+
+type resultsTable struct {
+	resultCount              map[lint.LintStatus]int
+	resultDetails            map[lint.LintStatus][]string
+	lintLevelsAboveThreshold map[int]lint.LintStatus
+	sortedLevels             []int
+}
+
+func (r resultsTable) newRT(threshold lint.LintStatus, results *zlint.ResultSet, longSummary bool) resultsTable {
```
dadrian, line 22: Doesn't technically matter in this case, since all of the member variables are effectively pointers, but this function should operate on pointer receivers.

dadrian, line 22:
```suggestion
func (r *resultsTable) newRT(threshold lint.LintStatus, results *zlint.ResultSet, longSummary bool) resultsTable {
```

v2/cmd/zlint/main.go
```diff
@@ -156,6 +161,10 @@ func doLint(inputFile *os.File, inform string, registry lint.Registry) {
 			log.Fatalf("can't format output: %s", err)
 		}
 		os.Stdout.Write(out.Bytes())
+	} else if summary {
+		formattedoutput.OutputSummary(zlintResult, false)
+	} else if longSummary {
+		formattedoutput.OutputSummary(zlintResult, true)
```
cpu, line 167: tiny nit: It looks like providing both `-summary` and `-longSummary` will short-circuit to only displaying the short summary. I think it would be a little bit nicer to either error if both flags are enabled or to flip these two `else if` cases so that the long summary takes precedent if both are provided.

v2/formattedoutput/formattedOutput.go
acaird: thanks @dadrian, I made that change

v2/cmd/zlint/main.go
acaird: I'm conflicted by this.  

Why do you think `longSummary` should take precedence?  My thinking was that shorter output is better, as as not to overwhelm the person running it.

Now that I'm looking at that whole if/else if/else block, I'm not too fond of it.  The original one was `prettyPrint` else `dump the json`.  It makes sense that those are exclusive, but it makes less sense to me that they are all exclusive.  I can imagine someone wanting all 3 of json output (pretty or not, but not both), a short summary, and a long summary.  That argues for pulling `summary` and `longSummary` out of that block and putting each of them in their own `if` statement.

Thoughts?

cpu: > I'm conflicted by this.

That's fair, I don't think it's meaningful enough to derail forward progress. We can ignore this for now.

> Why do you think longSummary should take precedence? My thinking was that shorter output is better, as as not to overwhelm the person running it.

My thought was that the short summary output is a subset of the long summary output so when both are requested it made sense to give the full output. It's not a strongly held position :-)

> That argues for pulling summary and longSummary out of that block and putting each of them in their own if statement.
> Thoughts?

That seems fine to me :+1:

cpu:
@zakird Thanks for merging, was just typing out a squash commit message.

zakird:
oh 🤦  sorry about that!

cpu:
No worries :-) Happy it's all merged up :tada:

acaird:
Thanks for merging it.  I'm glad it's useful.  I'm sorry about the tests, they are still forthcoming.

Issue #434, opened 2020-05-14
Cleanup util QCStatement code, use ZCrypto parsing/types


cpu:
I think a lot of the code in [`util/qc_stmt.go`](https://github.com/zmap/zlint/blob/master/v2/util/qc_stmt.go) in ZLint is now duplicated by code that landed upstream in ZCrypto w/ https://github.com/zmap/zcrypto/commit/13cf40b46ed3697eba66054aca90c73bbbc991d4

I think it makes sense for someone familiar with the ETSI ESI requirements to remove the ZLint code that isn't required anymore in favour of using the ZCrypto version. If there are types/behaviour in the ZLint parsing that isn't present in the ZCrypto version it would make sense to upstream it to that repo and then remove it from ZLint.

mtgag:
I think this is on our scope... I will try to incorporate this in the ETSI related lints also in the main project. These pull requests are not forgotten but unfortunately postponed. Hope to get back to it soon.

sleevi:
I started work on this, but I think there's still some work to be done before we can fully get there.

A few running notes from what I've found so far:
* `lint_qcstatem_qccompliance_valid` - can be deleted
* `lint_qcstatem_etsi_type_as_statem` - trivial
* `lint_qcstatem_etsi_present_qcs_critical` - trivial
* `lint_qcstatem_mandatory_etsi_statems` - Actually a buggy lint, but trivial :)
* `lint_qcstatem_qclimitvalue_valid` - because of how [ZCrypto](https://github.com/zmap/zcrypto/blob/4d171263147247189025e53b13fd2d5828d59755/x509/qc_statements.go#L118-L132) has to deal with CHOICE types, we don't have enough information in the [MonetaryValue](https://github.com/zmap/zcrypto/blob/4d171263147247189025e53b13fd2d5828d59755/x509/qc_statements.go#L58-L63) to know which CHOICE was done, and so can't be coverted
* `lint_qcstatem_qcpds_lang_case` - trivial
* `lint_qcstatem_qcpds_valid` - ZCrypto is too liberal in how it parses, both in terms of string type and trailing data, so this causes some bad certs to now pass where the hand-rolled parser rejects

I haven't finished everything, but I'm working on getting something together, and will also file issues once that's done for some of the ETSI lints that are not-quite-right (but can be "easily" fixed)

PR #448, opened 2020-06-05
Zcrypt qc

mtgag:
Adresses  this issue: https://github.com/zmap/zlint/issues/434


v2/lints/etsi/lint_qcstatem_qclimitvalue_valid.go
```diff
@@ -53,38 +45,36 @@ func isOnlyLetters(s string) bool {
 }
 
 func (l *qcStatemQcLimitValueValid) Execute(c *x509.Certificate) *lint.LintResult {
+	errString := util.ErrorStringBuilder{Delimiter: "; "}
+
+	if len(c.QCStatements.ParsedStatements.Limit) != 1 {
```
sleevi, line 50: When the QCStatement parsing was landed to ZCrypto, I couldn't find any language in the relevant specifications to assert that one and only one instance of a given QCStatement could be present. The `UNIQUE` in the `QC-STATEMENT` information object only means that there will only be one unique statement within the information object set, not that there can be only one unique instantiation. 

X.681 reiterates this in the introduction and explicitly in 3.4.8

v2/lints/etsi/lint_qcstatem_qclimitvalue_valid.go
```diff
@@ -53,38 +45,36 @@ func isOnlyLetters(s string) bool {
...
-	if len(errString) == 0 {
-		qcLv, ok := s.(util.EtsiQcLimitValue)
-		if !ok {
-			return &lint.LintResult{Status: lint.Error, Details: "parsed QcStatem is not a EtsiQcLimitValue"}
+	if qcLv.Amount < 0 {
+		errString.Append("amount is negative")
+	}
+
+	// Check whether the alphabetic currency code was set
+	if len(qcLv.Currency) > 0 {
```
sleevi, line 60: I was talking with @dadrian this morning about this. I think we need to/should change ZCrypto first. As mentioned on #434 , we need to disambiguate which CHOICE was selected. The easiest way to do this is to update ZCrypto such that `Currency` is a `*string` and `CurrencyAmount` is a `*int`, emphasizing effectively that either are optional, so that we can nil-test to distinguish between an 'empty' currency and a valid one.

I was also discussing whether it would be appropriate to have ZCrypto more closely align with the ASN.1 grammar here so that it can do things like check string type (`PrintableString`) and constraint length.

But I think we should resolve those ZCrypto issues, which is why I started poking at #434 - to suss them out.

v2/lints/etsi/lint_qcstatem_qcpds_lang_case.go
```diff
@@ -54,28 +46,23 @@ func isOnlyLowerCaseLetters(s string) bool {
...
-		pds := s.(util.EtsiQcPds)
-		for i, loc := range pds.PdsLocations {
-			if !isOnlyLowerCaseLetters(loc.Language) {
-				util.AppendToStringSemicolonDelim(&wrnString, fmt.Sprintf("PDS location %d has a language code containing invalid letters", i))
-			}
+	warnString := util.ErrorStringBuilder{Delimiter: "; "}
 
-		}
+	if len(c.QCStatements.ParsedStatements.PDSLocations) != 1 {
+		return &lint.LintResult{Status: lint.Error, Details: "invalid number of PdsLocations objects"}
```
sleevi, line 52: Same remark re: uniqueness; AFAICT, ETSI does not actually specify that there should be only one instance.

If I'm botching the ASN.1, I'm definitely happy to address! I'll reach out to Arno to see about raising it as a formal question.

v2/util/qc_stmt.go
```diff
@@ -117,23 +147,50 @@ func checkAsn1Reencoding(i interface{}, originalEncoding []byte, appendIfCompari
...
+	IdEtsiQcsQcRetentionPeriod,
+	IdEtsiQcsQcSSCD,
+	IdEtsiQcsQcEuPDS,
+	IdEtsiQcsQcType,
+}
+
+func IsAnyEtsiQcStatementPresent(c *x509.Certificate) bool {
+	statements := c.QCStatements
+	statementIDs := statements.StatementIDs
+	for _, oid := range statementIDs {
```
sleevi, line 162: It's worth calling out you use different syntax in Line 173.

Should this just be:
```suggestion
	for _, oid := range c.QCStatements.StatementIDs {
```

v2/util/qc_stmt.go
```diff
@@ -117,23 +147,50 @@ func checkAsn1Reencoding(i interface{}, originalEncoding []byte, appendIfCompari
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func IsQCStatementPresent(c *x509.Certificate, statementOid string) bool {
+	statementIDs := c.QCStatements.StatementIDs
+	for _, oid := range statementIDs {
```
sleevi, line 174:
```suggestion
	for _, oid := range c.QCStatements.StatementIDs {
```

v2/lints/etsi/lint_qcstatem_qclimitvalue_valid.go
mtgag: The motivation for this was to avoid ambiguity rather than to address an ASN.1 constraint. If more than one QcEuLimitValue exist an application that uses the certificate would have difficulties to decide which one to use. Some implementations would probably use the first occurence.

If that is too strict or out of scope it is fine to delete this.

cpu:

PR #449, opened 2020-06-07
Fix .onion tests to only apply to EV certificates

### LINKED ISSUE #440: Bug: .onion names are no longer prohibited by the Baseline Requirements
[`lint_san_dns_name_onion_not_ev_cert.go`](https://github.com/zmap/zlint/blob/fd40f579253ea1ebfb18a585ab5cd8e7dcde61aa/v2/lints/cabf_br/lint_san_dns_name_onion_not_ev_cert.go) has a lint that returns an error if a `.onion` name appears in the SAN and the certificate is not EV.

However, as of BRs [1.6.9](https://cabforum.org/2020/02/20/ballot-sc27v3-version-3-onion-certificates/), with the adoption of [CA/Browser Forum Ballot SC27v3](https://cabforum.org/2020/02/20/ballot-sc27v3-version-3-onion-certificates/), `.onion` names are now permitted in BR certificates, provided they are subject to additional constraints (differing from the EV Guidelines)

The following tasks are needed:
* Either Lints can be expanded to support a `RetiredDate` that is similar to their [`EffectiveDate`](https://github.com/zmap/zlint/blob/99579098a16c10d1d704b8b8149dd8c35329107f/v2/lint/base.go#L59-L62), or the EV-only lint be retired to 2020-03-27, the effective date they became permitted by the BRs
  * The effective date may need to be modified, such that it ensures that prior to the effective date, no `.onion` cert existed (EV or BRs) - that is, it should have been revoked.
  * While if issued after the effective date, but before the retired date, it should be EV.
* A new lint, for the BRs, to ensure that the label preceeding the `.onion` address is a:
  > "valid Version 3 Onion Address, as defined in section 6 of the Tor Rendezvous Specification - Version 3 located at https://spec.torproject.org/rend-spec-v3"
  as per the BRs, v1.6.9, Appendix C
  * The effective date for this lint would be 2020-03-27

sleevi:
Before this change, ZLint would reject .onion names in non-EV certs
via the `lint_san_dns_name_onion_not_ev_cert` lint, and if that
was suppressed, then complain about the missing Tor Service
Descriptor extension. As of CA/Browser Forum Ballot SC27, it's
allowed for v3 onion names to appear in DV/OV/IV certificates, and
the Tor Service Descriptor extension is neither required nor
prohibited for these.

This change corrects the Tor Service Descriptor tests to properly
account for it being mandatory for EV, while optional for DV/OV/IV.
This does not introduce new lints to ensure that the address is
itself a well-formed V2 (if EV) or V3 (all types) address, which
will come in a follow-up change.

Closes #440

### DIFF
--- a/v2/lints/cabf_br/lint_ext_tor_service_descriptor_hash_invalid.go
+++ b/v2/lints/cabf_br/lint_ext_tor_service_descriptor_hash_invalid.go
@@ -24,8 +24,6 @@ import (
 	"github.com/zmap/zlint/v2/util"
 )
 
-const onionTLD = ".onion"
-
 type torServiceDescHashInvalid struct{}
 
 func (l *torServiceDescHashInvalid) Initialize() error {
@@ -33,10 +31,14 @@ func (l *torServiceDescHashInvalid) Initialize() error {
 	return nil
 }
 
-// CheckApplies returns true if the certificate is a subscriber certificate that
-// contains a subject name ending in `.onion`.
+// CheckApplies returns true if the TorServiceDescriptor extension is present
+// or if the certificate is an EV subscriber certificate with one or more
+// subject names ending in `.onion`.
 func (l *torServiceDescHashInvalid) CheckApplies(c *x509.Certificate) bool {
-	return util.IsSubscriberCert(c) && util.CertificateSubjInTLD(c, onionTLD)
+	ext := util.GetExtFromCert(c, util.BRTorServiceDescriptor)
+	return ext != nil || (util.IsSubscriberCert(c) &&
+		util.CertificateSubjInTLD(c, util.OnionTLD) &&
+		util.IsEV(c.PolicyIdentifiers))
 }
 
 // failResult is a small utility function for creating a failed lint result.
@@ -86,7 +88,7 @@ func lintOnionURL(onion string) *lint.LintResult {
 // Execute will lint the provided certificate. An lint.Error lint.LintResult will be
 // returned if:
 //
-//   1) There is no TorServiceDescriptor extension present.
+//   1) There is no TorServiceDescriptor extension present and it's required
 //   2) There were no TorServiceDescriptors parsed by zcrypto
 //   3) There are TorServiceDescriptorHash entries with an invalid Onion URL.
 //   4) There are TorServiceDescriptorHash entries with an unknown hash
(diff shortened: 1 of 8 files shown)


sleevi:
Updated comment to reflect #440 still needs the new lint. I figured that would be more appropriate as a separate PR, but happy to fold it in here based on reviewer load. I thought this was already a big enough change to warrant separating as its own PR.

v2/util/time.go
```diff
@@ -57,6 +57,7 @@ var (
 	MozillaPolicy22Date         = time.Date(2013, time.July, 26, 0, 0, 0, 0, time.UTC)
 	MozillaPolicy24Date         = time.Date(2017, time.February, 28, 0, 0, 0, 0, time.UTC)
 	MozillaPolicy27Date         = time.Date(2020, time.January, 1, 0, 0, 0, 0, time.UTC)
+	CABFBRs_1_6_9_Date          = time.Date(2020, time.March, 27, 0, 0, 0, 0, time.UTC)
```
sleevi, line 60: One thing I debated: What to name this. The naming here for CAB related things is a bit weird. Up until CABV201Date, it *seems* like the naming for CABV actually tied to the version of the document (in which case, this should be `CABV169Date`

Except 201 was a ballot number, not a version number, since the EV Guidelines and the BRs are separately versioned.

I opted here to be explicit about document-and-version, in line with the ETSI date, but I could also see an argument for a descriptive date, like `OnionV3Date`, so that folks don't have to constantly figure out what changed in BRs 1.6.9 vs 1.7.0.

Alternatively, naming the Ballots, e.g. as `CABFSC27`, at least makes it easier to quickly look up what the diff was (and across documents).

Naming is hard, but I wanted to draw attention to this to see where folks felt.

v2/lints/cabf_br/lint_ext_tor_service_descriptor_hash_invalid.go
```diff
@@ -184,14 +186,19 @@ func (l *torServiceDescHashInvalid) Execute(c *x509.Certificate) *lint.LintResul
...
-		if _, found := descriptorMap[eTLDPlusOne]; !found {
-			return failResult(
-				"%s subject domain name %q does not have a corresponding "+
-					"TorServiceDescriptorHash for its eTLD+1",
-				onionTLD, subjDomain)
+	// For EV certificates, every `.onion` name is required to have a
+	// TorServiceDescriptorHash, so check if any of the onion subjects in the
+	// certificate don't have a TorServiceDescriptorHash for the eTLD+1 in the
+	// descriptorMap.
+	// See also https://github.com/cabforum/documents/issues/190
```
sleevi, line 193: We don't really have any normative guidance on what to do if this extension appears in a non-EV cert. My interpretation is that despite it being unnecessary for V3 certs, because it's optional, it should be legal to include this for some but not all v3 onion names.

Similarly, if/when we fix it for EV, I imagine that it'll be required for V2 names, but optional for V3 names. While prohibited-for-v3 makes the linting slightly easier, I suspect it makes CAs' lives harder, so I doubt that'll happen in the EVGs.

v2/lints/cabf_br/lint_ext_tor_service_descriptor_hash_invalid_test.go
```diff
@@ -16,7 +16,7 @@ func TestTorDescHashInvalid(t *testing.T) {
 	}{
 		{
 			Name:           "Onion subject, no service descriptor extension, before util.CABV201Date",
-			InputFilename:  "dnsNameOnionTLD.pem",
+			InputFilename:  "onionSANEVBefore201.pem",
 			ExpectedResult: lint.NE,
 		},
 		{
```
sleevi, line 22: Note: `onionSANMissingServDescHash.pem` is updated to be an EV certificate, which does mean that coverage for the optionality (for DV) certs is missed. I was torn on whether to add a DV cert (or, more aptly, restore the existing cert as a DV test and add a new cert as an EV test). I wasn't sure to what degree we wanted to cover this with testing.

cpu: I'm comfortable with the coverage you have in branch. Thanks for calling this out :+1:

v2/util/time.go
cpu: > Except 201 was a ballot number, not a version number, since the EV Guidelines and the BRs are separately versioned.

I think this one was my fault :sweat: In retrospect I agree it seems better to match to the document version number instead of referencing a ballot here.

> I opted here to be explicit about document-and-version,

That feels right for this case to me too :+1:

PR #450, opened 2020-06-10
Lint that .onion addresses are well-formed

### LINKED ISSUE #440: Bug: .onion names are no longer prohibited by the Baseline Requirements

sleevi:
Adds a new lint, identified as e_san_dns_name_onion_invalid,
that makes sure that the .onion addresses present within a
certificate are well-formed v2 or v3 addresses, according to
the v2 or v3 Rendezvous specifications.

Closes #440

### DIFF
--- a/v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
+++ b/v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
@@ -53,6 +53,19 @@ import (
 	"github.com/zmap/zlint/v2/util"
 )
 
+var (
+	// Per 2.4 of Rendezvous v2:
+	// Valid onion addresses contain 16 characters in a-z2-7 plus ".onion"
+	onionV2Len = 16
+
+	// Per 1.2 of Rendezvous v3:
+	// A hidden service's name is its long term master identity key.  This is
+	// encoded as a hostname by encoding the entire key in Base 32, including a
+	// version byte and a checksum, and then appending the string ".onion" at the
+	// end. The result is a 56-character domain name.
+	onionV3Len = 56
+)
+
 type onionNotValid struct{}
 
 func (l *onionNotValid) Initialize() error {
@@ -88,12 +101,8 @@ func (l *onionNotValid) Execute(c *x509.Certificate) *lint.LintResult {
 			}
 		}
 		onionDomain := labels[len(labels)-2]
-		switch len(onionDomain) {
-		// Onion v2 address. These are only permitted for EV, per BRs Appendix C.
-		// Per 2.4 of Rendezvous v2:
-		// Valid onion addresses contain 16 characters in a-z2-7 plus ".onion"
-		// This is handled later, during the Base32 decode
-		case 16:
+		if len(onionDomain) == onionV2Len {
+			// Onion v2 address. These are only permitted for EV, per BRs Appendix C.
 			if !util.IsEV(c.PolicyIdentifiers) {
 				return &lint.LintResult{
 					Status: lint.Error,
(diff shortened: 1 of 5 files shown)


v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
```diff
@@ -0,0 +1,140 @@
...
+		// version byte and a checksum, and then appending the string ".onion" at the
+		// end. The result is a 56-character domain name.
+		case 56:
+		default:
+			return &lint.LintResult{
+				Status:  lint.Error,
+				Details: fmt.Sprintf("%q is not a v2 or v3 Tor address", subj),
+			}
+		}
+		_, err := base32.StdEncoding.DecodeString(strings.ToUpper(onionDomain))
```
cpu, line 119: Can you help me understand why the `strings.ToUpper` is required here? The example addrs in the onion spec seem to be all lowercase (e.g. `pg6mmjiyjmcrsslvykfwnntlaru7p5svn6y2ymmju6nubxndf4pscryd.onion`). If this is to normalize what could be a case insensitive `onionDomain` from the cert would it be less confusing to use `strings.ToLower`?

v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
```diff
@@ -0,0 +1,140 @@
...
+		labels := strings.Split(subj, ".")
+		if len(labels) < 2 {
+			return &lint.LintResult{
+				Status: lint.Error,
+				Details: fmt.Sprintf("certificate contained a %s domain with too "+
+					"few labels: %q", util.OnionTLD, subj),
+			}
+		}
+		onionDomain := labels[len(labels)-2]
+		switch len(onionDomain) {
```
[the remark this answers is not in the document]
sleevi: Yeah, I realized that in hindsight and debated changing, but wasn't sure if someone started reviewing. Switched to use a var block with constants for the lengths (I think) and a conditional if, to make it a little easier to read.

v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
```diff
@@ -0,0 +1,140 @@
...
+		// end. The result is a 56-character domain name.
+		case 56:
+		default:
+			return &lint.LintResult{
+				Status:  lint.Error,
+				Details: fmt.Sprintf("%q is not a v2 or v3 Tor address", subj),
+			}
+		}
+		_, err := base32.StdEncoding.DecodeString(strings.ToUpper(onionDomain))
+		if err != nil {
```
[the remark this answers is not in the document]
sleevi: Ah, right! Done

v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
sleevi: Good question! I should have commented this, as it was otherwise magic.

The Base-32 specification (or at least, StdEncoding) works on RFC 4648, which defines the characterspace for Base-32 using purely uppercase characters. Lowercase would be distinct (as they are in base-64)/malformed. I added a comment to this effect, to make it clearer why it's being uppercased; any mixed case or lower case would be invalid base32

v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
```diff
@@ -0,0 +1,145 @@
...
+		} else {
+			return &lint.LintResult{
+				Status:  lint.Error,
+				Details: fmt.Sprintf("%q is not a v2 or v3 Tor address", subj),
+			}
+		}
+		// Base-32 uses an exclusively uppercase alphabet, per spec, but DNS is
+		// case-insensitive. Convert to uppercase to ensure well-formed names will
+		// decode.
+		if _, err := base32.StdEncoding.DecodeString(strings.ToUpper(onionDomain)); err != nil {
```
sleevi, line 125: Oh shoot, re-reading the docs while updating this comment, Golang will ignore newlines, which means you can still smuggle in invalid values.

I'll update this to first pass through a regex.

sleevi: OK, swapped to a regex, which makes the base32 encoding unnecessary. This is because the alignment (16, 56) means that the inputs are integral multiples of 40 input bits -> 8 output bytes, which means that if they match the regex, they are by definition correctly encoded.

This avoids golang allowing padding (e.g. a 12-byte input with 4 bytes of = padding) and newlines

v2/lints/cabf_br/lint_san_dns_name_onion_invalid.go
cpu: That makes sense, thank you

cpu:
Thanks again @sleevi. I hope this helps encourage some CAs to issue DV certs for `.onion` subjects.

PR #457, opened 2020-07-22
added a test using an existing CA cert

cardonator:
@cpu hopefully reusing an existing CA cert for this test is okay :smile:


cpu:
> hopefully reusing an existing CA cert for this test is okay

@cardonator Absolutely! Thanks for adding the test, I really appreciate it.

PR #460, opened 2020-08-04
README to suggest checking x509.ParseCertificate output

zakird:


README.md
```diff
@@ -126,7 +131,12 @@ import (
...
 var certDER []byte = ...
-parsed, _ := x509.ParseCertificate(certDER)
+parsed, err := x509.ParseCertificate(certDER)
+if err != nil {
+	// If x509.ParseCertificate fails, the certificate is too broken to lint.
+	// This should be treated as ZLint rejecting the certificate
+	log.Fatal("unable to parse certificate:", err)
+}
 
 registry, _ := lint.GlobalRegistry().Filter(lint.FilterOptions{
```
sleevi, line 141: Should this also be updated, to highlight the need to re-examine the filter list?

Just curious if this filter can end up causing things to 'fail open' (e.g. a bad filter syntax causing someone to think they're linting, but they aren't)

sleevi: Er, nevermind. I guess in that case, registry will be nil, and so line 134/144 will end up running all lints, thus failing closed.

zakird: Yeah, that's a good call. Will push update.

zakird: Even so, you probably want to know.

PR #464, opened 2020-08-31
Add citation for sub-CAs to ca_digital_signature_not_set

aarongable:
Adds an additional reference to the BRs to the ca_digital_signature_not_set
lint, to make it clear that it applies to both Root and Subordinate CA certs.


zakird:
Thanks for the contribution! Will merge once tests (inevitably) pass.

PR #469, opened 2020-09-03
Align Validity Period definition with RFC 5280

### LINKED ISSUE #467: e_tls_server_cert_valid_time_longer_than_398_days calculation is off by 1 second
RFC 5280 section 4.1.2.5 (and the BRs by reference) defines the Validity Period as `the period of time from notBefore through notAfter, inclusive.` This means that a certificate whose notBefore and notAfter field values are the same has a validity period of 1 second. Expanding from this base case, this means that certificates with a notAfter of exactly 398 days (34,387,200 seconds) greater than the notBefore value are actually valid for 34,387,201 seconds and thus are valid for 399 days per the rounding rules found in section 6.3.2 of the BRs.

It appears that the `e_tls_server_cert_valid_time_longer_than_398_days` lint does a simple time difference between the notBefore and notAfter and does not add 1 second to this interval to determine the certificate's Validity Period. To accurately detect certificates that do not comply with the validity period restrictions in the BRs, this additional second needs to be added to the interval and used for comparison.

CBonnell:
Fixes #467.

### DIFF
--- a/v2/lints/apple/lint_e_server_cert_valid_time_longer_than_398_days.go
+++ b/v2/lints/apple/lint_e_server_cert_valid_time_longer_than_398_days.go
@@ -42,13 +42,13 @@ func (l *serverCertValidityTooLong) Execute(c *x509.Certificate) *lint.LintResul
 	// "We recommend that certificates be issued with a maximum validity of 397 days."
 	warnValidity := 397 * dayLength
 
-	// Calculate the date that the certificate NotAfter must be less than
-	errEndDate := c.NotBefore.Add(maxValidity)
-	warnEndDate := c.NotBefore.Add(warnValidity)
+	// RFC 5280, section 4.1.2.5: "The validity period for a certificate is the period
+	// of time from notBefore through notAfter, inclusive."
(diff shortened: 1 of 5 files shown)


zakird:
This looks correct to me offhand, but would be good to get a second pair of eyes from @sleevi.

zakird (approved):

sleevi:
Yeah, the fractional validity depends on the accuracy of the client clock, given that UTCTime/GeneralizedTime in the abstract can represent fractional seconds, even if 5280 forbids them. That is, a cert with a notAfter (HHMMSS) at 000000Z and checked at 000000.001Z is expired, but if checked at 000000.000Z, valid, and the validity period was effectively (.001 seconds),

Thus, in theory, the validity period is 0 seconds (some infinitesimal slice of a second), but since machines aren’t that accurate, it’s typically 1ns when computed.

This LGTM as a change, and definitely underscores the SHOULD 397 days to avoid such issues entirely.

cpu:
@zakird @sleevi Do one of you want to try the [release instructions](https://github.com/zmap/zlint/blob/master/CONTRIBUTING.md#publishing-a-release) to cut a v2.2.0-rc2 release with this fix?

Issue #471, opened 2020-09-10
LINT for "The curious case of the Dangerous Delegated Responder Cert"


RufusJWB:
Would it be possible to implement a lint for certificates that fall under [The curious case of the Dangerous Delegated Responder Cert](https://www.mail-archive.com/dev-security-policy@lists.mozilla.org/msg13493.html)? For example this [ICA](https://crt.sh/?id=648772148&opt=zlint) is affected by the problem but seems to run through ZLINT without even a warning.

zakird:
We'd be happy to accept a lint that checks this.

RufusJWB:
I'll do my very best. Never programmed in GO before

cpu:
Happy to help review.

PR #472, opened 2020-09-11
lints: add lint for OCSP responder certificates

RufusJWB:
Fixes https://github.com/zmap/zlint/issues/471

@sleevi could you please review this PR? Thank you!


RufusJWB:
@cpu @zakird Could one of you please review this PR?

zakird (changes requested):

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
```diff
@@ -0,0 +1,64 @@
...
+		return true
+	}
+	return false
+}
+
+func (l *OCSPIDPKIXOCSPNocheckExtNotIncluded) Execute(c *x509.Certificate) *lint.LintResult {
+	if !util.HasEKU(c, x509.ExtKeyUsageOcspSigning) {
+		// This point should never be reached, because this LINT only applies to certificates with this EKU
+		return &lint.LintResult{Status: lint.Fatal}
+	}
```
sleevi, line 40: Do we need to duplicate this? It seems like we shouldn't, but if there's some reason I've overlooked, it also seems like

```
if !CheckApplies(c) {
```

would be better. But I don't think it's needed :)

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if !util.IsExtInCert(c, util.OscpNoCheckOID) {
+		if util.IsServerAuthCert(c) {
+			// If the certificate is a TLS certificate, it is clear, that the BRGs apply and we have an ERROR
+			return &lint.LintResult{Status: lint.Error}
+		}
+
+		// If the certificate is not a TLS certificate, the BRGs apply, if one of the sibling certificates or the parent certificate has the Server Auth EKU
+		return &lint.LintResult{Status: lint.Warn, Details: "Check the sibling and parent certificate for the Server Auth EKU. If one of them contains the Server Auth EKU, this is an ERROR and not a WARN"}
```
sleevi, line 49: I'm not sure I understand this part of the body.

BRs 4.9.9 apply to OCSP Responder Certificates, independent of any additional EKUs at play here.

My best guess here is that this is trying to cover a scenario where a Root issues Intermediate 1 and Intermediate 2. Intermediate 1 has `id-kp-serverAuth` and Intermediate 2 has `id-kp-emailProtection`.

You're wanting to make sure this lint runs for the OCSP Responder for Intermediate 1, and I'm guessing that this is trying to exclude Intermediate 2's OCSP Responder, right? With the assumption being that the issuing CA runs all lints (and all lint categories) for all certificates within their issuance hierarchy (e.g. they lint all of their S/MIME certificates just like they lint their TLS certificates, it's just we don't have S/MIME lints).

Just wanting to double check my understanding so far - did I get this right?

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
zakird: Agree, I think that this can be dropped.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: @sleevi Yes, you are totally correct. Let me try to give an example. Basically I want to achieve the following results:

https://crt.sh/?id=174851777&opt=zlint --> should be flagged as "Error"
https://crt.sh/?id=3065731585&opt=zlint --> should be flagged as "Pass"
https://crt.sh/?id=6487721485&opt=zlint --> should be flagged as "Warn" because one would need to check the intermediate CA above (https://crt.sh/?id=21606057) to see if it is capable of issuing TLS certificates and therefore the BRGs apply or not

I have tried to cover those three examples in https://github.com/RufusJWB/zlint/blob/master/v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go

I hope my understanding of this whole issue is correct.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: I updated the code with the proposal by @sleevi . I would like to let this check in here, even if it is redundant because it is defensive programming and one should never make assumptions on the value fed into a function as long as it not validated in the method itself.

sleevi: It’s actually possible to be too defensive. I totally hear you about not wanting to make assumptions, but as engineers, we also write code with defined or guaranteed behaviors, according to interface contracts.

The reason why I wasn’t sure, but Zakir confirmed my memory was correct, is that the current program defines it an invalid sequence to call `Execute` without calling `CheckApplies`. I totally understand not wanting to assume, but it’s a bit like how the code guarantees `Initialize` will be called beforehand, or like how it’s “undefined” (or “invalid” or “wrong”, depending on language) to use memory before it’s been initialized, or to write to a file handle you haven’t opened. Some things it’s OK to take for granted, because that’s the defined contract, and if anything crashes, it’s because the calling code is wrong, not this code.

I ultimately defer to @zakird and @cpu how strongly they feel here, because I do seem to recall there might be a few of the newer lints not consistently following this, but I think we can get away without this, saying it’s “programmer error” to call if `CheckApplies` returns false.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
sleevi: Gotcha. Yeah, this is a tricky case!

We don’t have issuer context presently, which was something we discussed when implementing the checks to make sure the signature hash algorithm of a subscriber cert is appropriate for the issuing CAs private key (e.g. SHA-384 with P-256). While we could fake a little there, based on the size of r/s, we don’t have that same luxury here.

Under today’s model, it’ll require a human to sort out. If we warned/errored on everything, that would reflect that in the BRs it’s a MUST, but we know it would have a false positive like in the case of Siemens. For a case like that, the only way to know whether it’s a false positive or not would be to inspect all certificate paths, or at least, all known/possible issuer certs as inputs to this cert. If we made it an error, CAs would get in the habit of ignoring errors, or needing to suppress errors, and we don’t want to encourage that.

I’m mixed on 43-46, and wondering if it may be unnecessary. The reason is that it seems like it’s (potentially) linting a different property, or would at least be covered by that property, of not having a server cert be an OCSP responder cert to begin with. However, I think I can also see the argument being made here, which is that such a cert would be unambiguously in scope, and so it doesn’t hurt to make that `Fatal` when we’re for sure.

I think my substantive feedback is more that we should document this more, as it at least wasn’t immediately clear and obvious on the first pass, but assuming I’ve got that understanding right, I think you’re right and that it makes sense as the best we can do here. I do think we can also perhaps structure this a little easier; for example, we know unambiguous if the extension **is** present, then it’s a `PASS`, so we could return that result then, which would let us unnest this logic. That would make it easier to reason about the state tree, and also I suspect make it easier to document.

As it’s late here, I’ll pick up tomorrow with concrete code suggestions, unless this all makes sense and you wanted to give a go at a restructuring and updated documentation.

If you do decide to update, one big thing that threw me was the remark about ‘sibling cert’, since it’s not the sibling that decides, but whether the parent is “technically capable”. I’ll try to come up with some more concrete suggestions tomorrow, if that doesn’t immediately make sense.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: I think there is a significant difference between relying on the compiler / runtime environment to correctly initialize the memory or call the constructor of an object and on relying on the programmer to call two methods in the correct order. But in the end, I'm following the advice of the maintainer :-)

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: Thank you for your feedback.

1. Regarding the `FATAL` I don't think it fits here because acc. to https://github.com/zmap/zlint/blob/master/CONTRIBUTING.md#adding-new-lints `FATAL` is reserved for 

> when there is an unresolvable error in `zlint`

Therefore I would propose to keep it as `Error`

2. Regarding the proposed restructuring, I did try to reflect it in my commit https://github.com/RufusJWB/zlint/commit/d49bcea722c2a909b7a4b17ebe2b5abdea487498 . Do you think the code is better to understand now?

v2/lint/base.go
```diff
@@ -83,8 +83,11 @@ func (l *Lint) CheckEffective(c *x509.Certificate) bool {
 // CheckEffective()
 // Execute()
 func (l *Lint) Execute(cert *x509.Certificate) *LintResult {
-	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) {
-		return &LintResult{Status: NA}
+	if l.Source == CABFBaselineRequirements {
+		if l.Name != "e_ocsp_id_pkix_ocsp_nocheck_ext_not_included" && !util.IsServerAuthCert(cert) {
```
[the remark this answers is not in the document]
RufusJWB: I implemented your proposal and started to change the first lint which was not working anymore: e_ca_key_usage_not_critical . How do you think about the code change?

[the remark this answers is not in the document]
RufusJWB: One more question: If you like the code change, do you think I should add the additional filter to all CheckApplies methods of all CAB BRG lints or only to those that fail the integration test without it?

cpu:
> @cpu @.zakird Could one of you please review this PR?

I won't have a chance to give this a review until the weekend.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
```diff
@@ -0,0 +1,58 @@
...
+		return &lint.LintResult{Status: lint.Pass}
+	}
+
+	if util.IsServerAuthCert(c) {
+		// If the certificate is a TLS certificate, it is clear, that the BRGs apply and we have an ERROR
+		return &lint.LintResult{Status: lint.Error}
+	}
+
+	// If the certificate is not a TLS certificate, the BRGs only apply, if the parent certificate has the Server Auth EKU or could possibly issue a certificate that has the Server EKU
+	return &lint.LintResult{Status: lint.Warn, Details: "If the parent (issuing) certificate contains the Server Auth EKU or no EKU at all this is an ERROR and not a WARN"}
```
sleevi, line 46: However, I'm not entirely certain we need to handle this case, and I'd be curious for feedback from other @zmap/zlint-maintainers .

Today, the CABF BR lints [are](https://github.com/zmap/zlint/blob/aa4e2619db00dbdea768e9afa70262f0d6af3417/v2/util/ca.go#L47) [run](https://github.com/zmap/zlint/blob/aa4e2619db00dbdea768e9afa70262f0d6af3417/v2/lint/base.go#L86):
- If the cert has no EKUs
- If the cert has >=1 EKU, and contains `id-kp-anyExtendedKeyUsage`
- If the cert has >=1 EKU, and contains `id-kp-serverAuth`

That means if you have a hierarchy, where you have
- Root CA
- Intermediate CA 1 (only has `id-kp-emailProtection`)
- Intermediate CA 2 (no EKU extension)
- Leaf Cert (`id-kp-emailProtection`)

Then we run all of the BR lints against `Intermediate CA 2`, which very likely will have some that create warnings/failures, even though it's beneath a technically-constrained sub-CA.

This PR updates the BR checks to *also* run if:
- The cert has >=1 EKU, and contains `id-kp-ocspSigning`

And this logic here is trying to handle the scenario of:
- Root CA
- Intermediate CA 1 (only has `id-kp-emailProtection`)
- OCSP Signer (has `id-kp-ocspSigning`)

Trying to handle "OCSP Signer" seems odd, because we don't handle "Intermediate CA 2" any specially. But I guess the argument is that "Intermediate CA 2" is generally unlikely from publicly trusted hierarchies, because different root programs (e.g. Microsoft) have long required that intermediates have EKUs, so while "Intermediate CA 2" is unlikely/rare, "OCSP Signer" is far more likely/common.

I have a _slight_ preference for making this consistently an error, consistent with how the other BR checks apply in the absence of the EKU, and that effectively means "assume OCSP signer certs are subject to the BRs", but I'm curious for other thoughts, especially from @cardonator (to understand how CAs with complex hierarchies have integrated and dealt with the existing false-positives)

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func TestDelegatedResponder(t *testing.T) {
+	inputPath := "ocspidpkixocspnocheckextnotincluded_delegated_responder.pem"
+	expected := lint.Pass
+	out := test.TestLint("e_ocsp_id_pkix_ocsp_nocheck_ext_not_included", inputPath)
+	if out.Status != expected {
+		t.Errorf("%s: expected %s, got %s", inputPath, expected, out.Status)
+	}
+}
```
sleevi, line 68: This isn't a hard block, but a number of our newer tests uses a loop, which makes it a lot easier to distinguish the test case / expectations from the boiler plate.

https://github.com/zmap/zlint/blob/aa4e2619db00dbdea768e9afa70262f0d6af3417/v2/lints/apple/lint_ct_sct_policy_count_unsatisfied_test.go#L31 is an example, if you'd like to model, which makes it slight easier to maintain and refactor.

sleevi (commented):
Sorry, I realize it may not be the most efficient that I keep asking questions about the "not directly the lint" bits, but trying to make sure I've got all of that paged in and correct, since I suspect the lint is easy once we get the supporting bits sorted :)

v2/lints/cabf_br/lint_ca_key_usage_not_critical.go
```diff
@@ -33,6 +33,12 @@ func (l *caKeyUsageNotCrit) Initialize() error {
 }
 
 func (l *caKeyUsageNotCrit) CheckApplies(c *x509.Certificate) bool {
+
+	// this check applies not, if the certificate only includes the OCSP EKU but is not a server auth certificate. Compare with https://github.com/zmap/zlint/pull/472#discussion_r489499211
+	if !(util.IsServerAuthCert(c) && !util.IsOCSPCert(c)) {
+		return false
+	}
```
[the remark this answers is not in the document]
cardonator: I agree @sleevi I'm not sure this condition is right. @RufusJWB can you explain the intention here?

v2/util/ca.go
```diff
@@ -44,6 +44,18 @@ func IsSubscriberCert(c *x509.Certificate) bool {
...
+	if len(cert.ExtKeyUsage) == 0 {
+		return true
+	}
+	for _, eku := range cert.ExtKeyUsage {
+		if eku == x509.ExtKeyUsageAny || eku == x509.ExtKeyUsageOcspSigning {
+			return true
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
cardonator: I'm curious about this as well. Are there other lints you are hoping to exclude from running in their `CheckApplies` if the cert has an OCSP EKU?

It seems somewhat duplicative and maybe we should consider picking one or the other (EKU/KU pulled from a `Has*` function OR `Is*Cert` boolean functions) and not have both methods of collecting the same information. I suspect the Is functions were born out of concern that contributors/integrators might not know all the possible EKU/KU values that could potentially exist on a certificate so we "blessed" a few commonly used ones, but that knowledge could be solved with documentation.

v2/lint/base.go
```diff
@@ -83,8 +83,11 @@ func (l *Lint) CheckEffective(c *x509.Certificate) bool {
 // CheckEffective()
 // Execute()
 func (l *Lint) Execute(cert *x509.Certificate) *LintResult {
-	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) {
-		return &LintResult{Status: NA}
+	if l.Source == CABFBaselineRequirements {
+		if !(util.IsServerAuthCert(cert) || util.IsOCSPCert(cert)) {
```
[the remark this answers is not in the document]
RufusJWB: This proposal makes it easier to read. I'll accept the suggestion.

v2/util/ca.go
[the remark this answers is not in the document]
RufusJWB: > I suspect the Is functions were born out of concern that contributors/integrators might not know all the possible EKU/KU values that could potentially exist on a certificate so we "blessed" a few commonly used ones, but that knowledge could be solved with documentation.

I created the `IsOCSPCert` function to have one central place which decides, whether a certificate is capable of signing OCSP responses or not. I created it analogously to the `IsServerAuthCert` which was already there.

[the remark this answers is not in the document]
RufusJWB: @sleevi 

> This contrasts with https://github.com/zmap/zlint/pull/472/files#diff-0e6a1fbcef97b2e7645afdbe65060787 where you just call HasEKU

I changed the method `Initialize` in https://github.com/RufusJWB/zlint/commit/66af7e6beebb4f185efaa28644294f3db8818e74 to use the helper method from the https://github.com/RufusJWB/zlint/blob/master/v2/util/ca.go file. Now we only have one place to clarify, what makes a certificate an OCSP responder certificate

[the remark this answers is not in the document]
RufusJWB: Do we all agree on the definition of what makes a certificate an OCSP responder certificate as it is implemented in the lines from above?

v2/lints/cabf_br/lint_ca_key_usage_not_critical.go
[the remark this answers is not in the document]
RufusJWB: > I agree @sleevi I'm not sure this condition is right. @RufusJWB can you explain the intention here?
This condition is the negotiation of the condition which I introduced based on https://github.com/zmap/zlint/pull/472#discussion_r489550435 . But I agree, it is very hard to read and should be rewritten.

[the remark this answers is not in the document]
RufusJWB: I have rewritten it to a simple check if the certificate is capable of ServerAuth or not. This makes everything easier to understand and should still be correct. https://github.com/RufusJWB/zlint/commit/8a16e16f28fa7a900206c52a11a056e8264f176a What do you think?

v2/util/ca.go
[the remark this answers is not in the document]
sleevi: No, I don’t think that’s correct, from spec or implementation 😅 That’s part of why I was trying to figure out the “why” with my [earlier comment](https://github.com/zmap/zlint/pull/472#discussion_r490533146), because I wasn’t sure if I was missing some context.

Specifically, I don’t think omitting EKU or having AnyEKU makes it an OCSP responder. From memory, a number of libraries require the explicit EKU, although I don’t recall from 6960 if that’s explicitly stated. It certainly “feels” wrong, but at least that makes it easier to sort through now that it’s clear the previous difference (between Has and Is) wasn’t an intentional decision.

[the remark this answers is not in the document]
RufusJWB: In [RFC6960 4.2.2.2](https://tools.ietf.org/html/rfc6960#section-4.2.2.2) it is stated:

> Therefore, a certificate's issuer MUST do one of the following:
> 
>    - sign the OCSP responses itself, or
> 
>    - explicitly designate this authority to another entity
> 
>    OCSP signing delegation SHALL be designated by the inclusion of
>    id-kp-OCSPSigning in an extended key usage certificate extension
>    included in the OCSP response signer's certificate.  This certificate
>    MUST be issued directly by the CA that is identified in the request.

So it seems to be clear: A certificate is a delegated OCSP responder certificate, if and only if it includes the id-kp-OCSPSigning extension.

But what about the case, when a CA signs the OCSP response itself? Would this CA certificate be considered as an OCSP responder certificate too? For the purpose of this lint, it is not necessary to decide this question, as the requirement to be tested only applies to delegated OCSP responders. --> I'll rename the method to IsDelegatedOCSPResponderCert .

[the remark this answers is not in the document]
RufusJWB: Can we set this conversation on resolved?

v2/lint/base.go
```diff
@@ -83,14 +83,19 @@ func (l *Lint) CheckEffective(c *x509.Certificate) bool {
 // CheckEffective()
 // Execute()
 func (l *Lint) Execute(cert *x509.Certificate) *LintResult {
-	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) {
+	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) && !util.IsDelegatedOCSPResponderCert(cert) {
+		// OCSP certificates also fall under certain BRG lints, even if they don't have the ServerAuth EKU set
```
sleevi, line 87:
```suggestion
		// The Baseline Requirements only apply to those technically capable of TLS issuance, as well as the
		// infrastructure supporting them, such as OCSP responders. This is checked early, to avoid having to
		// duplicate this check in every lint.
```

This just tries to have the comment match the conditional, in terms of negation, as well as explain the early check.

This is mostly a minor comment.

v2/lint/base.go
```diff
@@ -83,14 +83,19 @@ func (l *Lint) CheckEffective(c *x509.Certificate) bool {
...
 	}
+
 	if !l.Lint.CheckApplies(cert) {
 		return &LintResult{Status: NA}
-	} else if !l.CheckEffective(cert) {
-		return &LintResult{Status: NE}
+	} else {
+		if !l.CheckEffective(cert) {
+			return &LintResult{Status: NE}
+		}
```
sleevi, line 96: I think, with the new edits, you can restore this to what it was, right? This rewrite has the same processing as the original code, and the original code is more concise, so we'd prefer that as the idiom. Unless I'm overlooking an edge case here?

v2/lints/cabf_br/lint_ca_key_usage_not_critical.go
```diff
@@ -33,6 +33,12 @@ func (l *caKeyUsageNotCrit) Initialize() error {
 }
 
 func (l *caKeyUsageNotCrit) CheckApplies(c *x509.Certificate) bool {
+
+	// this check doesn't apply, if the certificate is not a server auth certificate
+	if !util.IsServerAuthCert(c) {
+		return false
+	}
+
 	return c.IsCA && util.IsExtInCert(c, util.KeyUsageOID)
```
sleevi, line 42: We can keep this code concise as well.

```suggestion
	// Key Usage is only required to be critical for Root CA and Subordinate CA certificates, and does not
	// currently apply to OCSP responder certificates.
	return c.IsCA && util.IsExtInCert(c, util.KeyUsageOID) && util.IsServerAuthCert(c)
```

This is also trying to remove the double negation from the original comment ("does **not** ... if **not**"), and tries to explain the exceptional situation (i.e. what isn't covered by the code), to make it more obvious what the code is functionally trying to accomplish.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name: "e_ocsp_id_pkix_ocsp_nocheck_ext_not_included",
+		Description: "OCSP signing Certificate MUST contain an extension of type id-pkixocsp-nocheck, as" +
+			" defined by RFC6960",
+		Citation:      "BRs: 4.9.9",
+		Source:        lint.CABFBaselineRequirements,
+		EffectiveDate: util.MozillaPolicy21Date,
```
sleevi, line 63:
```suggestion
		EffectiveDate: util.CABEffectiveDate,
```

This has been part of the BRs since 1.0

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
sleevi: @RufusJWB Any thoughts on this?

v2/testdata/ocspidpkixocspnocheckextnotincluded_SMIME_CA_correct.pem
```diff
@@ -0,0 +1,50 @@
+-----BEGIN CERTIFICATE-----
```
sleevi, line 1: I notice the test files also lack their "openssl x509" output. You can use https://github.com/zmap/zlint/blob/aa4e2619db00dbdea768e9afa70262f0d6af3417/v2/test/prepend_testcerts_openssl.sh to generate that output for the .pem files.

v2/util/ca.go
```diff
@@ -44,6 +44,11 @@ func IsSubscriberCert(c *x509.Certificate) bool {
 	return !IsCACert(c) && !IsSelfSigned(c)
 }
 
+func IsDelegatedOCSPResponderCert(cert *x509.Certificate) bool {
+	// compare https://tools.ietf.org/html/rfc6960#section-4.2.2.2
```
sleevi, line 48:
```suggestion
	// See https://tools.ietf.org/html/rfc6960#section-4.2.2.2
	// OCSP Delegated Responders require the EKU explicitly and id-kp-anyExtendedKeyUsage is not allowed.
```

sleevi (commented):
Thanks @RufusJWB! I think we're in the home stretch! A few minor comments (mostly about comments), and just some tweaking for the test files.

I notice they're rather large certificates; is there any thought given to making them synthetic certs with smaller keys? This isn't a hard blocker, by any means. Personally, I found https://github.com/google/der-ascii useful for generating test certs; it was easy to take one of the existing certs and just hand-modify the extensions and re-encode. This is also on the 'minor' comment, and I think in the worst case, one of the maintainers could easily swap them out before merging, so this isn't a hard blocker.

Not 100% I got the wrapping correct with my suggestions below (curse GitHub WebUI not showing 80 column limits), so feel free to adjust as appropriate.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
RufusJWB: @sleevi I followed your proposal and refactored test test into a loop. To make this whole thing more re-useable and minimize the amount of duplicated code, I moved the definition of the `TestCase` struct and of the `RunTest` func into https://github.com/RufusJWB/zlint/blob/master/v2/test/helpers.go . Now they can be re-used wherever needed in the code.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: I think there was some uncertainty in your original post on the effective date of this rule. MozillaPolicy21Date should be more conservative, than CABEffectiveDate as it is 1st of February 2013 and CABEffectiveDate is 1st of July 2012. If you are sure about CABEffectiveDate I'll be happy to accept your suggestion.

sleevi: Right, the effective date of this requirement was BRs 1.0, but that wasn't enforced by Mozilla consistently on all CAs until later. That was the uncertainty.

ZLint has historically gone by when the requirement was introduced (e.g. BRs 1.0), rather than when a particular program enforced the (existing) requirement. [Baseline Requirements, v1.0](https://cabforum.org/wp-content/uploads/Baseline_Requirements_V1.pdf#page=24) has this requirement in 13.2.5, so it's been a requirement since the get-go.

v2/go.sum
```diff
@@ -28,6 +28,7 @@ github.com/zmap/zcrypto v0.0.0-20191112190257-7f2fe6faf8cf h1:Q9MiSA+G9DHe/TzG8p
 github.com/zmap/zcrypto v0.0.0-20191112190257-7f2fe6faf8cf/go.mod h1:w7kd3qXHh8FNaczNjslXqvFQiv5mMWRXlL9klTUAHc8=
 github.com/zmap/zcrypto v0.0.0-20200513165325-16679db567ff h1:0DDYlvtXPb8EMtQPZ2TJDcM+adqtzy77QOndkCW79JQ=
 github.com/zmap/zcrypto v0.0.0-20200513165325-16679db567ff/go.mod h1:TxpejqcVKQjQaVVmMGfzx5HnmFMdIU+vLtaCyPBfGI4=
+github.com/zmap/zlint v1.1.0 h1:Vyh2GmprXw5TLmKmkTa2BgFvvYAFBValBFesqkKsszM=
```
sleevi, line 31: I don't think this is right, is it?

RufusJWB: To be honest, this was done by the "go" toolchain automatically and I simply committed it, even if I don't really understand, what it does. If you think, we don't need it, I'll revert it.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: yes, that makes sense. This lint checks compliance with the BRGs not if the BRGs were already required by the root store programs. I've accepted the suggestion.

v2/testdata/ocspidpkixocspnocheckextnotincluded_SMIME_CA_correct.pem
RufusJWB: @sleevi where do you want the "openssl x509" output to be stored? Do you have an example from a different test?

v2/lints/cabf_br/lint_ca_key_usage_not_critical.go
RufusJWB: I think we will in a later point in time need a discussion, what makes a certificate a TLS capable certificate. The function `IsServerAuthCert` was introduced with [commit](https://github.com/zmap/zlint/pull/171/commits/2caf99bdc64da040f4877cebff2f199a17f59253) and is not justified with references to the BRGs, Root Store Policies or RFCs.

In the Mozilla Root Store Policy 1.1 we find:

> 2.  Intermediate certificates which have at least one valid, unrevoked chain up
>     to such a CA certificate and which are technically capable of issuing 
>     working server or email certificates. Intermediate certificates that are not 
>     considered to be technically capable will contain either:
> 
>     * an Extended Key Usage (EKU) extension which does not contain any of
>       these KeyPurposeIds: anyExtendedKeyUsage, id-kp-serverAuth,
>       id-kp-emailProtection; and/or:
>     * name constraints which do not allow Subject Alternative Names (SANs) of
>       any of the following types: dNSName, iPAddress, SRVName, rfc822Name
> 
> 3.  End-entity certificates which have at least one valid, unrevoked chain up
>     to such a CA certificate through intermediate certificates which are all in
>     scope, such end-entity certificates having either:
> 
>     * an Extended Key Usage (EKU) extension which contains one or more of these
>       KeyPurposeIds: anyExtendedKeyUsage, id-kp-serverAuth,
>       id-kp-emailProtection; or:
>     * no EKU extension.

In the BRGs 7.1.2.3 we have:

> f. `extKeyUsage` (required)
> 
>    Either the value `id-kp-serverAuth` [RFC5280] or `id-kp-clientAuth` [RFC5280] or both values MUST be present. `id-kp-emailProtection` [RFC5280] MAY be present. Other values SHOULD NOT be present. The value `anyExtendedKeyUsage` MUST NOT be present.
> 

and in 7.1.2.2

> g. `extkeyUsage` (optional/required)
> 
>    For Cross Certificates that share a Subject Distinguished Name and Subject Public Key with a Root Certificate operated in accordance with these Requirements, this extension MAY be present. If present, this extension SHOULD NOT be marked critical. This extension MUST only contain usages for which the issuing CA has verified the Cross Certificate is authorized to assert. This extension MAY contain the `anyExtendedKeyUsage` [RFC5280] usage, if the Root Certificate(s) associated with this Cross Certificate are operated by the same organization as the issuing Root Certificate.
> 
>    For all other Subordinate CA Certificates, including Technically Constrained Subordinate CA Certificates:
> 
>    This extension MUST be present and SHOULD NOT be marked critical[^**].
> 
>    For Subordinate CA Certificates that will be used to issue TLS certificates, the value `id-kp-serverAuth` [RFC5280] MUST be present. The value `id-kp-clientAuth` [RFC5280] MAY be present. The values `id-kp-emailProtection` [RFC5280], `id-kp-codeSigning` [RFC5280], `id-kp-timeStamping` [RFC5280], and `anyExtendedKeyUsage` [RFC5280] MUST NOT be present. Other values SHOULD NOT be present.
> 
>    For Subordinate CA Certificates that are not used to issue TLS certificates, then the value `id-kp-serverAuth` [RFC5280] MUST NOT be present. Other values MAY be present, but SHOULD NOT combine multiple independent usages (e.g. including `id-kp-timeStamping` [RFC5280] with `id-kp-codeSigning` [RFC5280]).

We should double check, if the `IsServerAuthCert` is reflecting this.

v2/lint/base.go
RufusJWB: fixed.

v2/lint/base.go
```diff
@@ -83,14 +83,19 @@ func (l *Lint) CheckEffective(c *x509.Certificate) bool {
 // CheckEffective()
 // Execute()
 func (l *Lint) Execute(cert *x509.Certificate) *LintResult {
-	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) {
+	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) && !util.IsDelegatedOCSPResponderCert(cert) {
```
RufusJWB, line 86: @sleevi @cpu @zakird The more I think about this line, I wonder if we want to delete it in total and replace it with checks in the `CheckApplies` function of every lint. This would also support the upcoming SMIME-BRGs and the drafted Code Signing BRGs, especially since they will share a lot of requirements. Otherwise we will have more and more exceptions and strange combinations of OID checks.

sleevi: I don’t support doing that is this change, but I also think we should hold off on that refactoring until we get to that point. That seems like a lot of churn for unclear benefit, at least at present, although I can agree with you in the hypothetical.

One of the issues we had with QWACs is that it’s not merely sufficient for something to be a certificate; lints need an invested community of maintainers who are familiar with the specifications, the interpretations, and the use case. I suspect few, if any, of the maintainers are interested in Code Signing. The S/MIME WG activity seems too early to tell if it will produce something useful for Internet mail clients, given that right now it appears to be narrowly focusing on Enterprise/Gov’t cases and/or letting them dominate the use cases. As such, it may end up that those Guidelines aren’t adopted by major mail clients, or their audits not accepted/not seen as relevant, if they don’t end up being useful, and that seems like it could similarly impact the ability to support or maintain lints.

Given that historically ZLint has exclusively focused on TLS certs, I think the broader meta-question about other profiles is one we should hold off on, until we’ve got a body of lints and active maintainer interest. In the worst case, the structure of lints was designed to facilitate mixins of other lints in the interim, and wouldn’t be (negatively) affected by the current checks.

RufusJWB: I'll split an issue based on this conversation, so we don't forget about it and close this conversation here.

v2/go.sum
cpu: I think this part of the diff should be reverted. It doesn't make sense for the module to depend on itself (and at an older version!)

v2/testdata/ocspidpkixocspnocheckextnotincluded_SMIME_CA_correct.pem
cpu: @RufusJWB You can run the `prepend_testcerts_openssl.sh` script and it will literally add the output to the front of these `.pem` files where it will be ignored by the `x509` package at parse-time but will be present for human inspection.

cpu: Also I thought I landed a CI task that failed if there were `.pem` files missing this prepended output. Sounds like it might not be working correctly :-(

v2/util/ca.go
```diff
@@ -44,6 +44,12 @@ func IsSubscriberCert(c *x509.Certificate) bool {
 	return !IsCACert(c) && !IsSelfSigned(c)
 }
 
+func IsDelegatedOCSPResponderCert(cert *x509.Certificate) bool {
+	// See https://tools.ietf.org/html/rfc6960#section-4.2.2.2
+	// OCSP Delegated Responders require the EKU explicitly and id-kp-anyExtendedKeyUsage is not allowed.
```
cpu, line 49: Can you move this comment to be a [godoc comment](https://blog.golang.org/godoc) on the function definition?

v2/util/time.go
```diff
@@ -54,6 +54,7 @@ var (
 	OnionOnlyEVDate             = time.Date(2015, time.May, 1, 0, 0, 0, 0, time.UTC)
 	CABV201Date                 = time.Date(2017, time.July, 28, 0, 0, 0, 0, time.UTC)
 	AppleCTPolicyDate           = time.Date(2018, time.October, 15, 0, 0, 0, 0, time.UTC)
+	MozillaPolicy21Date         = time.Date(2013, time.February, 1, 0, 0, 0, 0, time.UTC)
```
cpu, line 57: I don't see this new exported constant referenced anywhere in this branch. Should it be removed?

cpu (changes requested):
Apologies for the delayed review. In all honesty I'm finding it challenging to keep up with open source in the current climate :-( Thanks to everyone else that put in their $0.02 already.

I think the base lint logic looks good. I've left some comments about the tertiary code.

v2/go.sum
RufusJWB: reverted

v2/util/ca.go
RufusJWB: @cpu Could you double-check the new documentation and mark this conversation as 'resolved' if the changed documentation is okay?

v2/util/time.go
RufusJWB: @cpu It was referenced in an older version. Since it is correct, I don't see a reason to remove it, as it might be needed later by someone else. But of course, I could also remove it. As you like.

v2/test/helpers.go
```diff
@@ -22,10 +23,29 @@ import (
...
 	"github.com/zmap/zlint/v2/lint"
 )
 
+type TestCase struct {
+	Name           string
+	Filename       string
+	ExpectedResult lint.LintStatus
+}
+
+func RunTest(lintName string, testCases []TestCase, t *testing.T) {
```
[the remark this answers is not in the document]
RufusJWB: I have marked it as `Helper` in case we decide to keep it.

v2/lints/apple/lint_ct_sct_policy_count_unsatisfied_test.go
```diff
@@ -28,11 +28,7 @@ func TestSCTCountPolicyUnsatisified(t *testing.T) {
 	// arguments that were used to create the test file.
 	//
 	// [0]: https://gist.github.com/cpu/6d26b2718f29e184ff88a90f02d7cbcb
-	testCases := []struct {
-		Name           string
-		Filename       string
-		ExpectedResult lint.LintStatus
-	}{
+	testCases := []test.TestCase{
```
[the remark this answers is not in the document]
RufusJWB: @cpu I reverted this change

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
cardonator: Regarding false positives, we do have a list of lints that tend to fail for not legitimate reasons that we review on a regular basis but don't prevent us from issuing a new ICA (or EE cert). The lints are still logged for review but they don't prevent us from issuing. Obviously, recent iterations of our system don't even allow us to create the hierarchy you mentioned above 😄 but there are several other scenarios in which a lint would fail for a not legitimate reason, particularly on ICAs and even more particularly when linting prior to the actual signing taking place.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
RufusJWB: I rolled back the refactoring and puit it into a different branch https://github.com/RufusJWB/zlint/blob/474-refactor-tests/v2/test/helpers.go Later we can merge it in a separate PR

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: @cardonator would the current distinction with `warning` and `error` be okay for you?

v2/lint/base.go
cardonator: IMO it is clearly under zlint's purview to lint any industry regulated certificate type, but totally agree that it requires invested maintainers to make it valuable. As other cert types become more useful for various purposes, one would hope that invested CAs would desire to contribute to the project (eventually) 🙂

v2/lints/cabf_br/lint_ca_key_usage_not_critical.go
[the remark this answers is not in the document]
RufusJWB: I have updated all lints that report false positives in the test corpus to check for server auth explicitly. Shall I update all other lints too? @cardonator  @cpu  @sleevi

v2/testdata/ocspidpkixocspnocheckextnotincluded_SMIME_CA_correct.pem
RufusJWB: @sleevi I used this handy der2ascii tool to reduce the size of the certificates significantly. Obviously the signature is now broken, but I think this is okay for this use case, isn't it?

RufusJWB: @cpu It seems that something is wrong with `prepend_testcerts_openssl.sh`

> z002m76a@md1r0uxc:/mnt/c/Repositories/zlint/v2/test$ ./prepend_testcerts_openssl.sh
> ./prepend_testcerts_openssl.sh: line 17: warning: command substitution: ignored null byte in input
> ./prepend_testcerts_openssl.sh: line 17: warning: command substitution: ignored null byte in input
> ./prepend_testcerts_openssl.sh: line 17: warning: command substitution: ignored null byte in input
> z002m76a@md1r0uxc:/mnt/c/Repositories/zlint/v2/test$   

any thoughts on this?

RufusJWB:
@cpu @zakird @sleevi do you have a chance of finalizing the review?

v2/util/time.go
sleevi: @RufusJWB Unless/until it's needed, it's better to remove. YAGNI and such :)

sleevi (changes requested):
@RufusJWB Thanks for doing this, and for continuing to plough through!

I think some things that stand out:
* As @cpu mentioned re: `config.json`, having some output of the new errors and warnings helps ensure a manual verification that things are correct. Are you able to provide some results from a test run that highlights these failures? Apologies if I'm missing it.
* You've modified a large number of `cabf_br` checks to change the `c.IsCA` check to only consider `util.IsServerAuthCert`.
  * I understand this motivation is driven by the change to to `v2/lint/base.go` now including OCSP signing certificates in scope, where previously, it only considered TLS certificates.
  * However, if an OCSP signing certificate asserts `cA:True`, then it's by definition, a subordinate CA certificate. If it's in scope for the BRs, then all of the BR requirements on Sub-CAs also apply.
  * As best I can guess, this is the same issue within your lint: uncertainty about whether or not something is in scope or not. However, as the underlying issue (which this is trying to lint) highlighted, if a CA "Root", which is in scope for the BRs, has issued an Intermediate Foo, which say asserts S/MIME and OCSP Signing, then "Foo" is a responder for "Root", and Subject to the BRs. Beyond pkix-no-check, this would also apply to all of the rules on Sub-CAs (i.e. the lints being disabled)

I suspect this is fundamentally a question about whether it'd be preferable to have false negatives (e.g. skipping lints) or false positives (e.g. including lints for certs that are out of scope). I *think* we might be better served by opting for false positives, since those at least get human attention and review, rather than false negatives, which suggest all is right for violations.

@zakird may have thoughts here, since he drew my attention to it.

I'll take a separate pass over the specific lint, but wanted to give this high-level feedback first.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
```diff
@@ -0,0 +1,73 @@
...
+			result := test.TestLint(lintName, tc.Filename)
+			if result.Status != tc.ExpectedResult {
+				t.Errorf("expected result %v was %v", tc.ExpectedResult, result.Status)
+			}
+		})
+	}
+}
+
+func TestOCSPIDPKIXOCSPNocheckExtNotIncluded(t *testing.T) {
+	testCases := []TestCase{
```
sleevi, line 43: It took me a while to figure out what the properties are that you're testing, since you're using "real" certs that have a number of fields.

Can you confirm my understanding:
* S/MIME CA Wrong
  * Includes OCSPSigning EKU, but lacks nocheck
  * `CheckApplies` is true (because EKU)
  * Result is a warning (because not clear it's in scope; in practice, it is, because the issuer is in-scope)
* S/MIME CA Correct
  * Omits OCSPSigning EKU entirely
  * `CheckApplies` is false (because EKU)
* TLS CA Wrong
  * Includes OCSPSigning EKU, omits nocheck
  * `CheckApplies` is true (because EKU)
  * Result is an error (because TLS serverAuth EKU is present)
* TLS CA Correct
  * Includes TLS serverAuth EKU, omits OCSP EKU
  * `CheckApplies` is false for this lint (because no OCSP EKU)
* Delegated Responder
  * Only includes OCSP Signing EKU, has nocheck
  * `CheckApplies` is true (because EKU)
  * Result is pass

If we look at how the code is structured, to obtain 100% coverage we need tests to ensure:
* OCSP EKU absent, (NA)
* OCSP EKU present, nocheck present (Pass)
* OCSP EKU present, nocheck absent, TLS EKU present (Error)
* OCSP EKU present, nocheck absent (Warning)

You have five tests, and it seems the "TLS CA Correct" and "SMIME CA Correct" are just testing the same thing. Am I overlooking something?

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
```diff
@@ -0,0 +1,73 @@
...
+		{
+			Name:           "SMIME CA Wrong",
```
sleevi, line 45: It took me a while to work through "why" this was wrong, by manually inspecting the certs and trying to compare logic. I think it might be helpful here to use Name to highlight what the essential properties are (e.g. in the list I just commented above), so it's easy to work through what the cert does or doesn't cover.

This is, admittedly, easier when they have the PEM data themselves.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
```diff
@@ -0,0 +1,73 @@
...
+			Filename:       "ocspidpkixocspnocheckextnotincluded_TLS_CA_correct.pem",
+			ExpectedResult: lint.NA,
+		}, {
+			Name:           "Delegated Responder",
+			Filename:       "ocspidpkixocspnocheckextnotincluded_delegated_responder.pem",
+			ExpectedResult: lint.Pass,
+		},
+	}
+
+	lintName := "e_ocsp_id_pkix_ocsp_nocheck_ext_not_included"
```
sleevi, line 70: Is this variable needed? Can it just be moved to line 72?

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
```diff
@@ -0,0 +1,73 @@
...
+ */
+
+import (
+	"testing"
+
+	"github.com/zmap/zlint/v2/lint"
+	"github.com/zmap/zlint/v2/test"
+)
+
+type TestCase struct {
```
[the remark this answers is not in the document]
sleevi: Note: This feels like premature abstraction, but this is an area where I defer more to the Go experts on local style and how much abstraction is appropriate.

sleevi (changes requested):

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
```diff
@@ -0,0 +1,73 @@
...
+			}
+		})
+	}
+}
+
+func TestOCSPIDPKIXOCSPNocheckExtNotIncluded(t *testing.T) {
+	testCases := []TestCase{
+		{
+			Name:           "SMIME CA Wrong",
+			Filename:       "ocspidpkixocspnocheckextnotincluded_SMIME_CA_wrong.pem",
```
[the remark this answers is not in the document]
RufusJWB: On my Windows 10 I can run all tests without any problem. I think I obtained them as pem-files from cert.sh, but I never double-checked. I had a look into them and you are right, they are binary files and not base64 ascii. Which would also explain, why prepend_testcerts_openssl.sh is not working for me. Do you think it makes sense to convert them or should I prepare test certificates from scratch only for this specific test case? @sleevi

v2/util/time.go
RufusJWB: I removed it.

v2/test/helpers.go
```diff
@@ -1,3 +1,4 @@
+// todo: This package should be renamed to something like "TestUtils"
```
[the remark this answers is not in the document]
RufusJWB: As a newbie in GO I had problems in understand the difference between the `test` and the `testing` package.

RufusJWB:
Based on issue https://github.com/zmap/zlint/issues/476 I'm wondering if I should refactor this LINT in two separate LINTs: one for certificates that fall clearly under the BRGs and one for those, where it depends on the issuing CA. How do you think about this? @robstradling @sleevi ?

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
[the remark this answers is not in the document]
RufusJWB: I created 32 new artificial PEM files with OpenSSL which includes the requested plain-text dumps. We should also have a full test coverage for every possible combination of EKUs now.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included.go
RufusJWB: I have split the one lint in two. There is one lint for `warning` and one for `error` now. I have also created test cases for all possible (32) combinations of the EKU OCSPSigning, serverAuth, emailProtection and extendedKeyUsage with OCSPNoCheck.

v2/test/helpers.go
```diff
@@ -22,10 +23,29 @@ import (
...
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			result := TestLint(lintName, tc.Filename)
+			if result.Status != tc.ExpectedResult {
+				t.Errorf("expected result %v was %v", tc.ExpectedResult, result.Status)
+			}
+		})
+	}
+}
+
```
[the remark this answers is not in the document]
RufusJWB: I changed it as you have proposed it

v2/integration/config.json
```diff
@@ -746,6 +746,10 @@
     },
     "e_cert_sig_alg_not_match_tbs_sig_alg": {
       "ErrCount": 10
+    },
+    "e_ocsp_id_pkix_ocsp_nocheck_ext_not_included": {
+      "ErrCount": 78,
+      "WarnCount": 95
```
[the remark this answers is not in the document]
RufusJWB: I did some sample testing.

For `Error`:
- https://censys.io/certificates/01a070613aaeb102c44468a231559e3b66d1982ec6435891909a9e6b366423d2
- https://censys.io/certificates/01f8971121f4103d30be4235cd7dc0eee6c6ae12fca7750848ea0e2e13fc2428
- https://censys.io/certificates/01fd614879394b3ae149252fc15698674159ce5550669ff912892e19835a1e4c
- https://censys.io/certificates/03b5ffe7db4d85717cb2ed07119dd63a8d592c2fd04f2988d1c074f0fffaaa2e
- https://censys.io/certificates/04ecba8f92bff7458c4a5e7c69261fc7e2ef52d5af54fbdd92b17141bbe0651f

For `Warn`:
- https://censys.io/certificates/0051c98924f239bd49ab8c8f6802f8b9cf4ad41eb3272f1a34c0cd6d61dcb3d4
- https://censys.io/certificates/007182136c03682d9b05d699ed6ed4fb78208b98f416fde2aa03f59421998ef5
- https://censys.io/certificates/01e87af0c5b740c8b1ccca07192a4284923b3f27672f6e1a64d7a61dac503627
- https://censys.io/certificates/05bfb6605d48516a571baf9a7ff75376130470da5ee7ff684c2672eaa0c0c8ad
- https://censys.io/certificates/0664f1eeeed1a7355700201d2f4431dc43b0bceab129a8cabfd4ed87d04341ef

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
RufusJWB: I have created test certificates for all 32 possible combinations and updated the tests respectively

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
RufusJWB: Solved

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_test.go
RufusJWB: did so

RufusJWB:
@sleevi @zakird @cpu @cardonator I know, that you are all very busy, but I think we are on the home run with this PR. Do you see a chance of reviewing it once more? I did try to implement all your advice and recommendation.

cpu:
> @sleevi @zakird @cpu @cardonator I know, that you are all very busy, but I think we are on the home run with this PR. Do you see a chance of reviewing it once more? I did try to implement all your advice and recommendation.

Thanks @RufusJWB, I agree this is looking much closer to a mergeable state now. Thanks for sticking with all of the review feedback, I know there has been a lot! I left a couple of comments but I don't think there is anything substantial left to figure out.

v2/integration/config.json
```diff
@@ -746,6 +746,12 @@
     },
     "e_cert_sig_alg_not_match_tbs_sig_alg": {
       "ErrCount": 10
+    },
+    "e_ocsp_id_pkix_ocsp_nocheck_ext_not_included_server_auth": {
+      "ErrCount": 78
```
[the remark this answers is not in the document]
RufusJWB: Please compare with https://github.com/zmap/zlint/pull/472#discussion_r501708198

v2/lint/source.go
```diff
@@ -25,17 +25,18 @@ import (
...
-	MozillaRootStorePolicy   LintSource = "Mozilla"
-	AppleCTPolicy            LintSource = "Apple"
-	ZLint                    LintSource = "ZLint"
-	AWSLabs                  LintSource = "AWSLabs"
-	EtsiEsi                  LintSource = "ETSI_ESI"
+	UnknownLintSource                   LintSource = "Unknown"
+	RFC5280                             LintSource = "RFC5280"
+	RFC5480                             LintSource = "RFC5480"
+	RFC5891                             LintSource = "RFC5891"
+	CABFBaselineRequirementsOverbearing LintSource = "CABF_BR_Overbearing"
```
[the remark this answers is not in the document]
RufusJWB: I did comment on this somewhere when I introduced, but I can't find the comment anymore. This PR gets confusing. The background of this additional set of LintSources is, that we do have rules, which come from the CAB BRGs that do apply for every certificate which is under an TLS capable CA, even if this certificate itself is not a TLS certificate. Now we had two options to represent this in the LINTS: option 1 was to update all other lints to reflect this (what I did at the beginning but got negative feedback here) and option 2 is to introduce this new lint source. At least I didn't have any more ideas how to handle the situation. If you have something else in your mind, please feel free to propose a different implementation and I'll update the sources accordingly.

v2/integration/config.json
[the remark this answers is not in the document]
cpu: Thank you! Sorry I missed that.

v2/lint/source.go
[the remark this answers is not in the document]
cpu: > This PR gets confusing.

Definitely, sorry for the thread maze we made here 😆 

> Now we had two options to represent this in the LINTS: option 1 was to update all other lints to reflect this (what I did at the beginning but got negative feedback here) and option 2 is to introduce this new lint source. At least I didn't have any more ideas how to handle the situation.

Hmm, I see. Given that `w_ocsp_id_pkix_ocsp_nocheck_ext_not_included_non_server_auth` is conservatively returning a `Warning` level finding I'm not sure I understand why we can't place these two new lints under the existing CABF source. What am I missing?

I'm inclined to stamp this even with the new source if others don't feel strongly. It would be possible to revisit in the future.

cpu (approved):
Stamping as-is but would appreciate if some of the folks that chimed in during the review process could weigh in as well. Particularly RE: the new lint source.

v2/lint/source.go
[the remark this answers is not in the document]
RufusJWB: @cpu If we list `w_ocsp_id_pkix_ocsp_nocheck_ext_not_included_non_server_auth` in `CABFBaselineRequirements` then we have to change line 86 of base.go to allow checks for non-server-auth-certificates for lints of the source `CABFBaselineRequirements`. But if we do this, we get plenty of new warnings and errors which are false positives.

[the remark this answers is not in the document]
RufusJWB: @cpu Do you think I would get this PR faster through the door if I would split it into two PRs? One for errors and one for warnings?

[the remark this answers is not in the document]
sleevi: @RufusJWB 

I think one of the challenges I've had is understanding the design constraints for your decisions, and I think that makes it a bit harder to try to reverse engineer to figure out what we can do better here. I think this is where, in thinking about the future, we might want to iterate and explore options on a GH issue first, just to make sure we capture the constraints (like you've done in https://github.com/zmap/zlint/pull/472/files#r508462206 )

I mention this, because
> But if we do this, we get plenty of new warnings and errors which are false positives.

I haven't really seen the breakdown here to understand what is meant here. I thought we'd found a viable path in previous iterations, which was including OCSP for the scope of the BRGs. I apologies that both personal and professional life have gotten busy here, but could you remind me why this approach doesn't work? I believe it was because you thought some OCSP responders would end up violating certain requirements, right? Could you remind me what those are?

I think we're at a place where there's fundamentally not any 'great' answer here, and the question we've been examining is whether to over-warn or under-warn. There was some discussion with how CAs are integrating, and I thought we'd accepted that over-warning was acceptable here. It sounds like you don't believe that's the case, so I'm trying to understand more.

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_non_server_auth_test.go
```diff
@@ -0,0 +1,185 @@
...
+		Filename       string
+		ExpectedResult lint.LintStatus
+	}{
+
+		// Legend for the nameing:
+		// o1 --> EKU OCSPSigning set; o0 not set
+		// s1 --> EKU serverAuth set, s0 not set
+		// ep1 --> EKU emailProtection set, ep0 not set
+		// a1 --> EKU anyExtendedKeyUsage set, a0 not set
+		// nc1 --> noCheck set, nc0 not set
```
sleevi, line 36: I don't know that this combinatorial approach makes sense anymore, now that you've fundamentally redesigned the thing.

As a reviewer, these fundamental redesigns make it difficult to pick up, because it means I have to start from scratch all over again, and that's why I was hoping we might take a step back here, make sure we're in agreement and understand the challenges, and figure out forward. Reviewing lots of code, only for it to be thrown away, I'm sure is not a good use of your time, but it also doesn't feel good as a reviewer.

Does it make sense to take a step back, discuss on the issue the possible approaches and pro/cons you see, and make sure we're in agreement? Given that this is now the... I believe fourth... different approach, it seems like you've thought a lot about this, and it might help to summarize that clearly and their merits, and then we can easily review the implementation as aligned with whatever tradeoffs we accept.

For example, my original suggestion was only to have 5 tests, and so I'm not sure why you've covered this extensive permutation here of all combinations. It's certainly interesting, but I also worry it's not really scalable or manageable for all these certificates, and may not be buying us anything for the conditions we logically care about. I'm happy, of course, to better understand and learn more, because I'm sure there's a good reason you thought this would be better, and it's just not obvious to me.

sleevi (changes requested):

cardonator:
I have to be honest, the approach here has changed so many times that I'm having a hard time keeping up. Is there any way we can close this PR and start over again from scratch so it's easier to follow what's being changed and why certain decisions were made?

This is a real struggle as it relates to a lint that I'd really like to see added to zlint to add better protection from problems that could arise from OCSP and thinking I can just juggle the few places that @sleevi pointed out without re-reviewing the entire thing over again. I'd rather just know I need to start over from the beginning!

RufusJWB:
@cardonator thank you for your proposal of starting over from scratch. This PR is really a mess. Basically, I was able to split this PR into three distinct PRs: #489 , #490 , #491 . Each of the PRs can be reviewed independently . I really hope that makes things easier.

#489 --> only improves the readability of a single line of code
#490 --> introduces the OCSP EKU check for TLS certificates
#491 --> introduces the OCSP EKU check for non TLS certificates

The split between #490 and #491 is based on the remark in https://github.com/zmap/zlint/issues/476#issuecomment-698376809 .

If you like this new approach I'll close this PR and we focus on the three new ones. Okay?

cardonator:
That is much better, thanks!

v2/lint/source.go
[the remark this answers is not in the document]
RufusJWB: @sleevi I'm really sorry, that this PR became such a big mess. I'm for sure not happy neither since I invested also a lot of my time to over and over update this PR. I think at the core the problem was, that I did try to cover both "warning" and "error" in one PR. This made things complex, especially because the "warning" lint is covering certificates which usually don't fall under the BRGs. Furthermore, I did try to follow all the proposals given by the various contributors and some advice I found in the discussion of other bugs. Wherever possible I did try to document this either in the commit message or in the discussion here in this PR. But anyhow, the situation has room for improvement and I have learned a lot about working in GitHub. Based on the proposal of @cardonator I closed this PR and created two new PRs, one for TLS certificates (#490 ) and one for non-TLS certificates (#491 ). Would you never the less be willing to review also the two new PRs? I think with two new PRs we are very close to getting this issue solved.

Issue #454, opened 2020-07-13
Request for classification of the key usage "keyEncipherment" in TLS certificates using elliptic curves as "error"


ENEN-DTR:
Dear ZLint-Team,

Recently we wrongly issued one certificate using elliptic curves with TLS attributes which contained the key usage “keyEncipherment” instead of “keyAgreement.“ The incorrect key usage was inadvertently assigned. This was not previously detected in the dual control procedure.

Linting is part of our pre- and post-production process. We use your linting tool. If the outcome of the linting process is an error, no certificate will be produced. If the outcome is a warning, an operator or product manager of our company will investigate the reason but the certificate will be produced anyway.

In our case no error was confirmed and the certificate was produced. But we got a warning and the certificate was subjected to an additional and more detailed manual check after being issued. The incorrectly used key usage was detected and the certificate was revoked.

As one of the actions to prevent the production of certificates with the same error, we classified this error in our implementation as an “Error” instead of an “Warning.” I want to ask you to reclassify this case from “Warning” to “Error” in the source code of ZLint. This might help prevent other CAs from running into the same situation that we had. 

Thanks,
Enrico 
D-TRUST

zakird:
Thanks for the report, @ENEN-DTR. Which lint produced a `WARN` instead of an `ERROR`? Can you point to the BR or RFC that states that this is mandatory?

sleevi:
The RFC text is currently being processed in IETF: https://tools.ietf.org/html/draft-ietf-lamps-5480-ku-clarifications

This was addressed through Mozilla Policy previously.

ENEN-DTR:
We use the latest version 2.1.0

sleevi:
https://www.rfc-editor.org/info/rfc8813 is now published

PR #479, opened 2020-09-25
Update lint_ecdsa_ee_invalid_ku.go

ENEN-DTR:
Changed Error handling for invalidKUs

### DIFF
--- a/v2/lints/rfc/lint_ecdsa_ee_invalid_ku.go
+++ b/v2/lints/rfc/lint_ecdsa_ee_invalid_ku.go
@@ -136,7 +136,7 @@ func (l *ecdsaInvalidKU) Execute(c *x509.Certificate) *lint.LintResult {
 
 func init() {
 	lint.RegisterLint(&lint.Lint{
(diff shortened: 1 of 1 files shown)


zakird:
Can you provide some context around the change? Based on the RFC text, the original code looks to be correct. Nothing says that the certificate `MUST NOT` have other codes.

ENEN-DTR:
Yes, of course. This should address issue #454 .

zakird:
The text of this RFC seems to be rather different from what's being checked in the proposed change:

```
   If the keyUsage extension is present in a certificate that indicates
   id-ecPublicKey in SubjectPublicKeyInfo, then following values MUST
   NOT be present:

     keyEncipherment; and
     dataEncipherment.

   If the keyUsage extension is present in a certificate that indicates
   id-ecDH or id-ecMQV in SubjectPublicKeyInfo, then the following
   values also MUST NOT be present:

     keyEncipherment; and
     dataEncipherment.
```

The PR will need to check _exactly_ what's laid out in this RFC and also update the description to reference these statements.

ENEN-DTR:
PR closed to #599

PR #490, opened 2020-10-30
Ocsp eku check for tls certificates

RufusJWB:
This PR is split from https://github.com/zmap/zlint/pull/472 to simplify the review process.

I did some sample testing on the findings on the test corpus:

- https://censys.io/certificates/01a070613aaeb102c44468a231559e3b66d1982ec6435891909a9e6b366423d2
- https://censys.io/certificates/01f8971121f4103d30be4235cd7dc0eee6c6ae12fca7750848ea0e2e13fc2428
- https://censys.io/certificates/01fd614879394b3ae149252fc15698674159ce5550669ff912892e19835a1e4c
- https://censys.io/certificates/03b5ffe7db4d85717cb2ed07119dd63a8d592c2fd04f2988d1c074f0fffaaa2e
- https://censys.io/certificates/04ecba8f92bff7458c4a5e7c69261fc7e2ef52d5af54fbdd92b17141bbe0651f

--> All the samples do look reasonable

Issue #476, opened 2020-09-24
[Discussion] Lints that cover both SHOULD and MUST conditions in the same lint


cardonator:
Hi all, I wanted to discuss the precedent set by some recent lints in terms of combining SHOULD and MUST conditions into one lint. 

In earlier zlint times, every lint was clearly discernible by its name as to whether it would return an Error, Warn, or Notice condition (by the `e_` prefix, for example). That meant you could use the lint name to distinguish zlint related conditions in your logs pretty easily as often the name of the lint also described what condition you were running into.

Recently we've added some lints that cover multiple SHOULD and MUST conditions and could return Notice, Warning, and Error conditions inside the same lint. A recent example is `e_tls_server_cert_valid_time_longer_than_398_days` which return a warning for SHOULD between 397-398 day validity. Due to this, the lint name became less readable as to what condition it was failing to meet and getting a warning on.

So I thought I would raise this conversation about what the project _should_ be doing in these cases. Should we have multiple lints for covering the various possible conditions in the industry requirements? Does it make sense for an `e_` lint to also return other types of conditions? If so, should we adjust how we look at Lint naming to ensure it aligns more appropriately with the conditions the lint is checking? (As a poor man's example, maybe something like `m_tls_server_cert_validity` perhaps `m_` for multiple potential response statuses?)

robstradling:
Thanks for starting this discussion, @cardonator.  I'm interested in the outcome because https://crt.sh/?zlint=issues currently shows a misleading "TLS server certificates issued on or after September 1, 2020 00:00 GMT/UTC must not have a validity period greater than 398 days" warning for certs with validity periods greater than 397 days but not greater than 398 days.

I would prefer to see SHOULD and MUST conditions in separate lints.  e.g., in this particular case, move the "SHOULD...397 days" condition into a separate lint named `w_tls_server_cert_valid_time_longer_than_397_days`.

zakird:
This sounds like a regression and we need to fix it. Lints should only return a warning or error based on their type.

zakird:
This is not explicitly laid out in https://github.com/zmap/zlint/blob/master/CONTRIBUTING.md, but we should update to match.

Issue #428, opened 2020-05-07
Discussion: Issuer certificate as (optional) input to lint


mtgag:
Hello,

this came up while working on 
https://github.com/zmap/zlint/pull/378

The current implementation makes some assumptions because the certificate issuer is not present. An assumption-free implementation would need the issuer certificate as input to the lint. 

The proposal is to allow the issuer certificate as (optional) input to lints. This has the advantage that several other lints would then be implementable (e.g. signature verification, AKIE consistency, SubjectDN-issuerDN encoding consistency) also the lint who brought this issue up could be implemented in another way.

What is your opinion on that?

zakird:
One issue is that this makes testing a specific certificate non-deterministic. We could incidentally mark a certificate as having an `error` because we passed in the wrong parent in during the linting process. We should probably have some way of denoting which parent was used for the test in the lint output.

dadrian:
As @zakird noted, the non-determinism makes this a bit complicated. Assuming the use-cases @mtgag described above (AKID checking, Issuer-Subject DN comparisons), I don’t see us automatically finding the issuer, since to decide _which_ certificate is the issuer, we would be using the aspects of the certificate these lints are specifically checking for errors. Necessarily, we would need to know the “right” answer for the SKID of the parent certificate, the encoding of the Subject DN of the parent, and so on.

Given this, I’d assume anyone using these lints would have a priori decided which certificates to lint against which issuers. One use case might be a CA linting only their own certificates from a specific issuer at time of issuance.

I suggest we take any lints of this form, put them in their own “group”, and assume that ZLint is instantiated with a single issuer to check against, leaving it up to the user to select which certificates to lint, and which issuer to use. We can somehow indicate in the output if we do not think the issuer certificate actually issued an input certificate. We can probably also support one issuer per key type, e.g. one RSA cert, one ECDSA certificate, etc.

If someone wishes to send all certificates ever issued through the lints requiring an issuer, it would be “on them” to figure out which issuer to use with which certificate.

zakird:
I agree with @dadrian points out here, but also note that CAs add the parent to CT, so in many ways we  know the issuing certificate that CAs intended to use when performing external validation.

mtgag:
I thought about a simple approach.  Something like extending the LintInterface to

CheckApplies(c *x509.Certificate, issuer *x509.Certificate) bool and 

Execute(c *x509.Certificate, issuer *x509.Certificate) *LintResult

where the issuer is optional.

[...]

PR #491, opened 2020-10-30
Ocsp eku check for non tls certificates

RufusJWB:
This PR is a split from #472 to simplify the review process.

I did some sample testing on the findings on the test corpus:

- https://censys.io/certificates/0051c98924f239bd49ab8c8f6802f8b9cf4ad41eb3272f1a34c0cd6d61dcb3d4
- https://censys.io/certificates/007182136c03682d9b05d699ed6ed4fb78208b98f416fde2aa03f59421998ef5
- https://censys.io/certificates/01e87af0c5b740c8b1ccca07192a4284923b3f27672f6e1a64d7a61dac503627
- https://censys.io/certificates/05bfb6605d48516a571baf9a7ff75376130470da5ee7ff684c2672eaa0c0c8ad
- https://censys.io/certificates/0664f1eeeed1a7355700201d2f4431dc43b0bceab129a8cabfd4ed87d04341ef

--> All the samples do look reasonable


RufusJWB:
@RufusJWB @cardonator any thoughts on this newly simplified Lint?

RufusJWB:
@cardonator @zakird since the OCSP responder topic came up on the mdsp mailing list again, do you think we could finalize this PR?

RufusJWB:
Is there someone willing to review?

v2/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_non_server_auth.go
```diff
@@ -0,0 +1,60 @@
...
+	if util.IsExtInCert(c, util.OscpNoCheckOID) {
+		return &lint.LintResult{Status: lint.Pass}
+	}
+
+	// If the certificate is not unambiguously a TLS certificate, then whether or not the OCSP responder is
+	// in scope of the Baseline Requirements depends on whether the issuer of this certificate is capable of
+	// being used for TLS. At present, this requires manual attention, so only return a warning.
+	return &lint.LintResult{
+		Status: lint.Warn,
+		Details: "OCSP signing Certificate without id-pkix-ocsp-nocheck. If the associated CA certificate is" +
```
christopher-henderson, line 45: It would be nice if we could properly convert this to an error in the case that the issuer has `serverAauth`

@zakird There wouldn't happen to already be any sort of framework in place for taking an entire chain into consideration (or otherwise having the presence of a cert pool for a particular run) would there?

v2/lint/source.go
```diff
@@ -25,17 +25,18 @@ import (
...
-	MozillaRootStorePolicy   LintSource = "Mozilla"
-	AppleCTPolicy            LintSource = "Apple"
-	ZLint                    LintSource = "ZLint"
-	AWSLabs                  LintSource = "AWSLabs"
-	EtsiEsi                  LintSource = "ETSI_ESI"
+	UnknownLintSource                   LintSource = "Unknown"
+	RFC5280                             LintSource = "RFC5280"
+	RFC5480                             LintSource = "RFC5480"
+	RFC5891                             LintSource = "RFC5891"
+	CABFBaselineRequirementsOverbearing LintSource = "CABF_BR_Overbearing"
```
christopher-henderson, line 32: Howdy @RufusJWB, I'll be taking over a good deal of the day-to-day maintenance of the repo so I'll be getting us over the finish line with these PRs. 

I have briefly reviewed the pull request that preceded this one, and it seems that I'm about to continue the thread on this particular issue with more windy roads :smile:

It seems as though this category of linter was added because a certificate could be an OCSP Delegated Responder and _not_ have a `serverAuth` EKU. Normally then, the CAB BR would not have an opinion on it. However, the CAB BR _does_ have an opinion on it if the issuing certificate (or any certificate further up the chain?) _does_ have the `serverAuth` EKU. Is that a correct restatment?

However, since we do not have a copy of the certificate chain, we cannot verify whether or not the CAB BR applies, so we simply call it a warning and put it in the category of "Overbearing" (which I reckon is hinting at the notion that the category is the CAB BR possibly "overstepping" its bounds).

But what if we _did_ have a copy of the certificate chain? Would that allow us to disambiguate this?

```go
// Nothing in the chain is a TLS cert, so warn on CAB.
func (w* WarningLint) CheckApplies(c *x509.Certificate) bool {
	return util.IsDelegatedOCSPResponderCert(c) && !util.Any(c.Chain, util.IsServerAuthCert)
}
// Something in the chain is a TLS cert, so error on CAB.
func (w* ErrorLint) CheckApplies(c *x509.Certificate) bool {
	return util.IsDelegatedOCSPResponderCert(c) && util.Any(c.Chain, util.IsServerAuthCert)
}
```
Note that I'm not 100% on the semantics of that example, but it was just an example.

@zakird The easiest way I can currently think to accomplish something like this is to update zcrypto's `Certificate` type to actually retain the chain so that linters can access it if present (with minimal changes to ZLint aside from the obvious CLI update that would have to occur). I reckon that these sorts of hacks are the exact reason why the fork exists, right? I'd be happy to open a PR if we like this idea.

sleevi: Well, one of the big issues is that there are many chains, and the violation might depend on the context, or the result might change over time (as new chains become available).

Today, things are hermetic: the results for a cert are consistent over time (at least, for the same version). Adding the chain makes it a bit complex.

It seems that there’s room for offering an API here, but it would require fleshing out a bit of a proposal. There’s a big question on whether it’s better to:
  1. Supply a single chain, and perform multiple invocations
  2. Provide multiple chains in one call.

There are good arguments for both, such as performance and usability, so it’s probably worth considering on a new issue and noodling about the design space here.

zakird: This is discussed here: https://github.com/zmap/zlint/issues/428. As @sleevi mentions, it's a bit unfortunate because lint results are now dependent on a (leaf, intermediate) instead of just the (leaf,). But, if it allows us to tests additional functionality, I'm of the opinion: so be it, particularly given that CT already provides the chain in the API, so both monitors and CAs have a pretty good idea what they "should" test. We need to change the API though to accept parent certificates optionally, and likely a new output status.

RufusJWB: Hi Christopher! Thank you for taking over here and I'm very interested to get this PR through the door. It was a long trip so far. I would love to implement your proposal, but one problem I'm seeing with it is, that it collides with https://github.com/zmap/zlint/issues/476 where it was stated, that one lint should have only one result, either `WARN`or `ERROR`. So basically we would need to move the `util.Any(c.Chain, util.IsServerAuthCert)` into the `isApplicabable` method of both lints. What do you think?

christopher-henderson: So, I still like the idea of extending the CLI to accept zero-or-more certificate chains as I reckon that this is not the last time that a linter is going to want to have the chain around in order to do its business. However, that is a much larger architectural change that Id' rather do on my own time instead of blocking this PR on it (as Zakir and Sleevi point out there are some interesting implications and considerations to this).

Besides, I believe that the core issue would still be there even if we did implement such a system. That is, if an operator runs this lint without providing at least one certificate chain, then what should we output here? A `Warning` that says in the comments that this might actually be an `Error`? People are used to warnings complaining about `MAY` clauses whereas this particular warning is actually saying that they _might_ be in violation of a `SHALL`. That sounds like a sneaky pitfall.

How about a new `lint.LintStatus` ,`Unproven`? This can stand in places where we want to say, "listen, this lint _might_ apply or it _might_ pass but we don't have enough information to know for sure". It's a notch down from a failure, but is more eyebrow raising than a warning.

cardonator: It seems to me that there are just two lint conditions, one is only relevant if the parent certificate chain was provided. In our systems, we have had constant questions about how to perform lints where the certificate chain is relevant and, since that currently isn't within the purview of zlint, we have had to manually implement those lints. I'm not sure we can simply ignore the benefits of adding the chain to the set of lints that could then be created.

That said, I still agree with the concern I raised on #428 that it seems particularly heavy to modify the Lint interface for every Lint when only a subset of Lints requires an issuing chain to be passed in. It will also cause incompatibilities with external library integrations, so the cost of modifying the Lint interface should be weighed heavily before making such a change. I don't have a good solution to present for this, unfortunately.

Historically, we have preferred to have the lints following the appropriate MAY, SHOULD, and MUST terminology that is present in the guidelines, however in #492 I indicated another area that presents a parallel, but similar, flaw in that translation. The issue is that zlint can't possibly always know enough information to determine if the lint currently being run violates the Status level that the lint was created for. I'm not sure that adding another Status accurately reflects the amount of concern that is warranted via failure in those cases.

v2/lint/source.go
```diff
@@ -71,6 +72,8 @@ func (s *LintSource) FromString(src string) {
 		*s = RFC5480
 	case RFC5891:
 		*s = RFC5891
+	case CABFBaselineRequirementsOverbearing:
```
cardonator, line 75: I know we kind of already had this conversation but I'm curious on @sleevi thoughts on this. We did discuss that this is an actual CABF requirement, right? Are we considering it Overbearing because we can't check that it's actually violating the requirement currently?

RufusJWB:
@christopher-henderson @cardonator Do you know, what I need to change to make the `test` working again? The integration is working but the pipeline complains :-(

v2/lint/source.go
christopher-henderson: @cardonator could it be an issue of granularity? That is, completely flesh out the enum to have...

```go
UnprovenWarning = 8
UnprovenFatal = 9
...
```

...so that we can accurately convey that we might have found a `MUST` violation but we couldn't quite prove it. A question remaining there is, "should an `UnprovenFatal` result in a non-zero exit code?".

cpu:
> Do you know, what I need to change to make the test working again? The integration is working but the pipeline complains :-(

@RufusJWB The integration test stage did pass in CI. There are two other failures:

* https://github.com/zmap/zlint/actions/runs/40565548 - this is a true-positive golangci-lint finding. You need to `gofmt -s` the `v3/lints/cabf_br/lint_ocsp_id_pkix_ocsp_nocheck_ext_not_included_non_server_auth_test.g` file.
* https://github.com/zmap/zlint/pull/491/checks?check_run_id=1510210079 - this is two unit tests in `TestRegister` failing. I thought I fixed this ordering in https://github.com/zmap/zlint/commit/0d48ea15507cb16a08f694b8d9b64f91b601d3e7 but it might not be a stable ordering. Does the test pass locally for you? I can look deeper myself later today.

v2/lint/source.go
cardonator: @christopher-henderson hmm, maybe. I think the challenge with a concept like "Unproven" is that it's unclear how concerned you should be (which you correctly pointed out). Currently, zlint failure levels are pretty straightforward, i.e. if you get a Fail it's a big concern, if you get a Warning it's a medium concern and if you get a Notice it's just something interesting to take note of or verify your compliance with.

In the context of this suggestion, what would you change about the ultimate solution to #492? I think that's a good sample for how something like this would be handled, although this particular case is likely more perilous than that one was if research determined a CA was out of compliance.

sleevi: Should we move this discussion to #428 just to capture it all? It sounds like there are elements of the design space being explored, and it might make sense to make sure it's kept in one place (or some of the proposals here written up). Not a big ask, just keeps drawing my attention to this issue, which is a bit different :)

v2/lint/source.go
sleevi: That was my understanding: That the current design was accounting for a lack of the full chain (see #428 ), and so was using this category as a way of letting CAs distinguish so they can possibly suppress this case. I'm not entirely sold (e.g. they could suppress this lint for certificates issued from their S/MIME CAs, without needing the extra source), and from the spec side, it's still a responder.

RufusJWB:
@cpu I was able to fix the first of the two errors. I have no idea, but for whatever reasons, Visual Studio Code didn't run `gofmt -s -w` on this very file. I did it manually and now it passes the source code lint.

Regarding the second test: How can I execute it locally? I did try `make test` but this ends with:

```
GORACE=halt_on_error=1 go test -race ./...
Der Befehl "GORACE" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
make: *** [makefile:31: test] Error 1
```

v2/lint/source.go
RufusJWB: My main reason to introduce this new category was, to execute this one test for all certificates, regardless if it is a server auth certificate or not. The first option I explored was to keep it in the group `CABFBaselineRequiremens` and remove the filter, that lints from this group only apply to server auth certificates. But to do so, I had to adapt 50+ lints. So I introduced this new group of lints.

cpu:
> Regarding the second test: How can I execute it locally? I did try make test but this ends with:

@RufusJWB If I'm reading the machine translation of your err correctly it looks like your shell is interpreting the `GORACE` env var as a command. Are you using BASH on Linux or something more esoteric? You might prefer to use the makefile with `make test`. It will run the equiv. command without needing shell var management.

In this case since the race detector likely doesn't matter and we're interested in a single file's tests I would try a more targeted execution like:

```
go test -v --test.run "TestRegister*" ./v3/lint/`
```

RufusJWB:
@cpu I'm using something very esoteric: PowerShell on Windows ;-) But luckily I had the Windows Subsystem for Linux installed and with that, I was able to execute the command. Sadly I got the same error message as in the git actions. And I have no idea, how to fix them.

cpu:
> I'm using something very esoteric: PowerShell on Windows ;-) But luckily I had the Windows Subsystem for Linux installed and with that, I was able to execute the command. 

Hah ok, you're on your own with that setup :-)

> I was able to execute the command. Sadly I got the same error message as in the git actions.

I'm able to reproduce this locally too and it _does_ seem like a regression in this branch.

<details>
  <summary>Checking out master I see tests pass: </summary>
  
zlint on  HEAD [$?] took 12s
> git checkout zmap/master
Previous HEAD position was 24ef2127 merge to master
HEAD is now at 1eb11ce3 Ocsp eku check for tls certificates (#490)

zlint on  HEAD [$?]
> git rev-parse HEAD
1eb11ce3552704904404d56a65f4be3511828dbf

zlint on  HEAD [$?]
> go test -v --test.run "TestRegister*" ./v3/lint/
=== RUN   TestRegister
=== RUN   TestRegister/nil_lint
=== RUN   TestRegister/nil_lint_ptr
=== RUN   TestRegister/empty_name
=== RUN   TestRegister/duplicate_name
=== RUN   TestRegister/bad_init_with_initialize
=== RUN   TestRegister/bad_init_with_no_initialize
=== RUN   TestRegister/good_lint_register
--- PASS: TestRegister (0.00s)
    --- PASS: TestRegister/nil_lint (0.00s)
    --- PASS: TestRegister/nil_lint_ptr (0.00s)
    --- PASS: TestRegister/empty_name (0.00s)
    --- PASS: TestRegister/duplicate_name (0.00s)
    --- PASS: TestRegister/bad_init_with_initialize (0.00s)
    --- PASS: TestRegister/bad_init_with_no_initialize (0.00s)
    --- PASS: TestRegister/good_lint_register (0.00s)
PASS
ok  	github.com/zmap/zlint/v3/lint	(cached)
</details>

vs:

<details>
  <summary>Checking out your branch I can reproduce the failure:</summary>
  zlint on  HEAD [$?] took 2s
> git checkout zmap/pr/491
Previous HEAD position was 1eb11ce3 Ocsp eku check for tls certificates (#490)
HEAD is now at 24ef2127 merge to master

zlint on  HEAD [$?]
> git rev-parse HEAD
24ef2127f1f9f22a1034436eabf1757abe050449

zlint on  HEAD [$?]
> go test -v --test.run "TestRegister*" ./v3/lint/
=== RUN   TestRegister
=== RUN   TestRegister/nil_lint
=== RUN   TestRegister/nil_lint_ptr
=== RUN   TestRegister/empty_name
=== RUN   TestRegister/duplicate_name
=== RUN   TestRegister/bad_init_with_initialize
=== RUN   TestRegister/bad_init_with_no_initialize
=== RUN   TestRegister/good_lint_register
    TestRegister/good_lint_register: registration_test.go:173: expected sources [ZLint Mozilla], got [Mozilla ZLint]
--- FAIL: TestRegister (0.00s)
    --- PASS: TestRegister/nil_lint (0.00s)
    --- PASS: TestRegister/nil_lint_ptr (0.00s)
    --- PASS: TestRegister/empty_name (0.00s)
    --- PASS: TestRegister/duplicate_name (0.00s)
    --- PASS: TestRegister/bad_init_with_initialize (0.00s)
    --- PASS: TestRegister/bad_init_with_no_initialize (0.00s)
    --- FAIL: TestRegister/good_lint_register (0.00s)
FAIL
FAIL	github.com/zmap/zlint/v3/lint	0.006s
FAIL
</details>

>  And I have no idea, how to fix them.

@RufusJWB Give me a little bit, I think I can figure this out now that I have a local failure to work with. If so I'll suggest a diff.

cpu:
> Give me a little bit, I think I can figure this out now that I have a local failure to work with. If so I'll suggest a diff.

@RufusJWB The root cause of the problem is 6a277113bf5c53cccde8db5e390e855e135bdd64 When you merged the master branch back into your PR in that commit you made the **lint.Community** source from https://github.com/zmap/zlint/commit/0d48ea15507cb16a08f694b8d9b64f91b601d3e7 have the value `"Zlint"` instead of `"Community"`. That changed the sort order of things like `[]SourceList{egLint.Source, MozillaRootStorePolicy}` so that the registry tests began to fail.

With this diff applied to your branch all of the unit tests pass locally again:
```diff
diff --git a/v3/lint/source.go b/v3/lint/source.go
index a7c0f2aa..37fae73b 100644
--- a/v3/lint/source.go
+++ b/v3/lint/source.go
@@ -34,7 +34,7 @@ const (
        CABFEVGuidelines                    LintSource = "CABF_EV"
        MozillaRootStorePolicy              LintSource = "Mozilla"
        AppleRootStorePolicy                LintSource = "Apple"
-       Community                           LintSource = "ZLint"
+       Community                           LintSource = "Community"
        EtsiEsi                             LintSource = "ETSI_ESI"
 )
```

RufusJWB:
@cpu your diff did the trick! Thank you very much!

v2/lint/source.go
RufusJWB: How do we get this lint through the door now? Currently, I see the following options:
1. We wait until #428 is implemented and finalize this lint again
2. We create a new category for TLS BRG lints that are 'overbearing' (the current approach)
3. We remove the filter for TLS certificates in the base.go file and update most of the over lints as I did here: https://github.com/zmap/zlint/pull/472/commits/5efa6d1e2f7cdc1db39b998baf92e77c522330b4
4. We do some special case handling for this lint in the base.go file as I did here: https://github.com/zmap/zlint/pull/472#pullrequestreview-489707702

Which version do you prefer @cpu @cardonator and @christopher-henderson ?

zakird: I don't think I'm comfortable with (2) or (4) from a codebase perspective. Creating a new category (2) will cause confusion --- it's not obvious what folks should do about the results that come from it, and (4) inappropriately mixes lint code into the framework. It would be good to get @sleevi's take on this as well since he knows the bug at hand better than I do.

RufusJWB: For me it is okay to go with (3) but I had the feeling in the discussion of the original PR, that the maintainers weren't happy with adjusting practically all BRG lints - which would be the consequence of choosing (3).

v2/lint/source.go
christopher-henderson: Ported this over to #428 and fired up #522 to discuss the issue of ambiguous lint results.

@RufusJWB seeing as we got the other review in for certs that are definitively for TLS, would it be okay for this one to sit on the back-burner until we figure out the above (it just seems less pressing than explicitly TLS certs)?

RufusJWB: @christopher-henderson I'm sorry, that I come back to this topic only now. They the holiday season is supposed to be the quietest time in the year but in reality, I always have the feeling it is the most stressful one at work... Anyhow, yes I'm fine with waiting for the results of those two discussions.

dadrian:
I'm not sure why this stalled out a few years ago, but now it merge conflicts. I'm trying to get ZLint back into shape and so will close this out.

If this lint is still relevant, please let me know and/or reopen rebased against main.

Issue #418, opened 2020-03-11
breaking change: rename exported lint.AppleCTPolicy constant


cpu:
The `lint.AppleCTPolicy` constant has a generic value ("Apple") but the constant name itself is too specific "CT Policy": https://github.com/zmap/zlint/blob/b28794baf37819442e98e28655a8d4536681b095/v2/lint/source.go#L35

That was fine for the v2.0.0 release because there was only one lint in `lints/apple/` using the source, and it is CT specific. Now that we're adding more Apple policy lints (e.g. #417) we need a better solution.

Since this is an exported var we can't rename it (or change the value) without it being a breaking change. We could introduce a second "lint.ApplePolicy" constant, but if it had the same generic value there will be trouble unmarshaling.

Likely the best solution is to rip the bandaid off and just rename the constant from `lint.AppleCTPolicy` to `lint.ApplePolicy` and bump the major version again. Since that's a fair amount of work I'm filing this as an issue to save for when we have more breaking changes to make.

zakird:
Agree, we should "just do it" and do it sooner rather than later. _sigh_.

cpu:
> Agree, we should "just do it" and do it sooner rather than later. sigh.

Wellllllllll.... sooner turned into later anyway :-) BUT we have [a better reason](https://github.com/zmap/zlint/pull/493) to cut a major release so we might as well knock this out at the same time.

I assigned this to myself and will try to land a PR shortly.

cpu:
PR up: #501

PR #493, opened 2020-11-04
make two lints notice instead of warn, 

### LINKED ISSUE #492: dnsname_wildcard_left_of_public_suffix: Currently Warn but should be Notice?
I was just noticing some warning results we have seen recently in our implementation of zlint and looking into them a bit deeper. I noticed this lint show up a couple of times and looking at the Source/Description I'm not sure that it qualifies as a Warning.

The description says:
```
the CA MUST establish and follow a documented procedure[^pubsuffix] that determines if the wildcard character occurs in the first label position to the left of a “registry‐controlled” label or “public suffix”
```

This is complicated with the currently established meaning of Warn and Error somewhat correlating to SHOULD and MUST requirements. There is no way for zlint to ensure that the CA has established and follows this procedure. Does it damage the value of Warning level lints to have Warnings that zlint can't actually verify?

A similar lint that almost always pairs with this one is `san_iana_pub_suffix_empty` which is a community lint from cablint. As this is a community lint and doesn't correlate to a BR does it make sense to also be a notice? Interested in other contributor's thoughts.

cardonator:
As discussed in #492 these two lints don't make sense to be Warning level. The BR they cover is not possible to be verified by a linter (BR 3.2.2.6).

closes #492

### DIFF
--- a/v2/benchmarks_test.go
+++ b/v2/benchmarks_test.go
@@ -119,7 +119,7 @@ func BenchmarkZlint(b *testing.B) {
 			for _, key := range names {
 				switch key {
 				case "w_dnsname_underscore_in_trd", "e_dnsname_underscore_in_sld", "e_dnsname_hyphen_in_sld",
-					"w_dnsname_wildcard_left_of_public_suffix", "w_san_iana_pub_suffix_empty":
+					"n_dnsname_wildcard_left_of_public_suffix", "n_san_iana_pub_suffix_empty":
 					continue
 				}
 				value := lint.GlobalRegistry().ByName(key)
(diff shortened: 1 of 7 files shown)


zakird (approved):

zakird:
Looks good to me. Will give other folks some time to chime in. Otherwise, we can get merged next day or so.

cpu (approved):
This seems fine to me. Thanks @cardonator 

One note for folks to remember, this will require bumping the [major version](https://github.com/zmap/zlint#versioning-and-releases) (taking care to do it in the go modules compatible way):
> Since downstream projects depend on lint results and names for policy decisions changes of this nature will result in MAJOR version revision.

cardonator:
Thanks for the reminder @cpu. Should we try to get a fix in for #476 for that update as well, then? Are there other lints that could use a result change we should try to get in?

cpu:
@cardonator Sorry I missed your follow-up comment. It looks like #476 was resolved by a docs update, was there something else you were thinking needed to be done? 

>  Are there other lints that could use a result change we should try to get in?

I can't think of any such lints off-hand. It would be nice to fix https://github.com/zmap/zlint/issues/418 as part of this major release but otherwise I'm probably in favour of cutting the major release ASAP. It will make it easier to not have to juggle multiple branches in the event we need to do a point release for a bugfix or something in the meantime. Better to release smaller releases more often IMO.

I'll try to grab #418 this weekend and will start making plans for a v3 release soon after unless anything else comes up.

cardonator:
No worries @cpu the only other one I can think of is fixing the 398 day lint which has a warning and error on the same lint.

cpu:
> the only other one I can think of is fixing the 398 day lint which has a warning and error on the same lint.

oh, good call! Thanks - https://github.com/zmap/zlint/issues/496

PR #495, opened 2020-11-10
Update Contributing Guidelines

### LINKED ISSUE #412: Project Decision: What is a Warning?
When we initially started ZLint, all warnings and errors were expected to have a citation to an RFC or BRs statement. All other checks were expected to be Informational. There was a single exception to this rule: presence of subject CN since _every_ certificate has this. IIRC, some browsers would reject certificates without. Instead, all certificates today have an informational for this.

Since then, a few things have changed:

 * We've found that many things we want lints for are not explicitly forbidden by a standards document despite agreement among all contributors that no one should issue certificates in that manner. Some of these have snuck in as warnings (which bother me slightly since it violates our documentation). 

 * Many CAs block now issuance on a `WARNING` or `ERROR` but not an `INFO`. I'd guess everyone ignores `INFO`s. 

 * We have extended lint sources as new standards have emerged

With that in mind, I think we should make a project-level decision about what to do with citation-less-but-important lints. I think there are a few options:

 * Add a new lint source `COMMUNITY` and use the existing INFO/WARNING/ERROR levels there. If CAs want to exclude the ZLint community's best practices, that's their choice.

 * Change our definition of `WARNING` to not require a citation

 * Remove `CN` lint so that informational lints can be used in a blocking manner

I'm also of course open to others' ideas. What do folks think? 

/cc @cardonator, @sleevi, @cpu

### LINKED ISSUE #476: [Discussion] Lints that cover both SHOULD and MUST conditions in the same lint

zakird:
This PR codifies decisions made in #476 and #412. 

Fixes #476 
Fixes #412

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ fields filled out.
    lints should not be used for non-deterministic errors.
 
  * **Notice:** `Notice` should be used for more general "FYI" statements that
-   indicate there may be a problem.
+   indicate there may be a problem. Non-deterministic lints are OK. 
 
 Lints only return one non-success or non-fatal status, which must also match
 its name prefix. For example, `e_ian_wildcard_not_first` can only return a
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ part of the certificate processing.
 
 **Lint Source:** Typically Lint Source is straightfoward since every lint needs
 a citation. However, sometimes the community has lints that aren't codified in
-a formal document. In these situations, do not create an `NOTICE` lint under a
+a formal document. In these situations, do not create a `NOTICE` lint under a
 common source (e.g,. RFC or Baseline Requirements). Instead, create a lint
 using the `ZLint` source. Lints in this source are included at the maintainers'
 discretion, though we typically shy away from lints with significant
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,16 +22,35 @@ Struct names following Go conventions, e.g., `subjectCommonNameNotFromSAN`. Exam
 This will generate a new lint in the `lints/rfc` directory with the necessary
 fields filled out.
 
-**Choosing a Lint Result Level.** When choosing what `lints.LintStatus` your new
-lint should return (e.g. `Notice`,`Warn`, `Error`, or `Fatal`) the following
-general guidance may help. `Error` should be used for clear violations of RFC/BR
-`MUST` or `MUST NOT` requirements and include strong citations. `Warn` should be
-used for violations of RFC/BR `SHOULD` or `SHOULD NOT` requirements and again
-should include strong citations. `Notice` should be used for more general "FYI"
-statements that violate non-codified community standards or for cases where
-citations are unclear. Lastly `Fatal` should be used when there is an
-unresolvable error in `zlint`, `zcrypto` or some other part of the certificate
-processing.
+**Choosing Result Level.**  Lints return a single type of status:
+
+ * **Error:** `Error` can only be used for clear violations of `MUST` or `MUST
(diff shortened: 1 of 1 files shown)


sleevi (approved):

CONTRIBUTING.md
```diff
@@ -22,16 +22,35 @@ Struct names following Go conventions, e.g., `subjectCommonNameNotFromSAN`. Exam
...
-processing.
+**Choosing Result Level.**  Lints return a single type of status:
+
+ * **Error:** `Error` can only be used for clear violations of `MUST` or `MUST
+   NOT` requirements and must include a specific citation.
+
+ * **Warning:** `Warn` can only be used for violations of `SHOULD` or `SHOULD
+   NOT` requirements and again should include strong citations. Many
+   certificate authorities block on both Error and Warning lints, and Warning
+   lints should not be used for non-deterministic errors.
```
cpu, line 33: tiny nit: what's "non-deterministic" in this context? Without more background I might assume it's a lint that returns a warning finding some of the time but not others for the same input certificate. Do you mean more that this shouldn't be used for cases where the linter can't make a conclusive determination (e.g. due to missing information)? If so it might be better to be more explicit about that vs saying non-deterministic. WDYT?

CONTRIBUTING.md
```diff
@@ -22,16 +22,35 @@ Struct names following Go conventions, e.g., `subjectCommonNameNotFromSAN`. Exam
...
+Lints only return one non-success or non-fatal status, which must also match
+their name prefix. For example, `e_ian_wildcard_not_first` can only return a
+`SUCCESS`, `ERROR`, or `FATAL` status.  It cannot return a `NOTICE` or
+`WARNING` status. Any lint can return a `FATAL` error, but `FATAL` should only
+be used when there is an unresolvable error in `zlint`, `zcrypto` or some other
+part of the certificate processing.
+
+**Lint Source:** Typically Lint Source is straightfoward since every lint needs
+a citation. However, sometimes the community has lints that aren't codified in
+a formal document. In these situations, do not create an `NOTICE` lint under a
```
cpu, line 47: minor typo:
```suggestion
a formal document. In these situations, do not create a `NOTICE` lint under a
```

cpu (approved):
Thanks @zakird !

CONTRIBUTING.md
zakird: e.g., SN entropy

zakird: I've added the example.

cpu: Thanks, I think the example is enough to clarify :thumbsup:

cpu (approved):

cardonator (commented):
This looks great, thanks!

Issue #473, opened 2020-09-18
LINT for the correlation between EKU and KU


RufusJWB:
I couldn't find a LINT for the correlation between EKU and KU as described in [RFC5280 4.2.1.12](https://tools.ietf.org/html/rfc5280#section-4.2.1.12) . Is there one? If not, we should implement one according this spec:

>  If a certificate contains both a key usage extension and an extended
>    key usage extension, then both extensions MUST be processed
>    independently and the certificate MUST only be used for a purpose
>    consistent with both extensions.  If there is no purpose consistent
>    with both extensions, then the certificate MUST NOT be used for any
>    purpose.
> 
>    The following key usage purposes are defined:
> 
>    anyExtendedKeyUsage OBJECT IDENTIFIER ::= { id-ce-extKeyUsage 0 }
> 
>    id-kp OBJECT IDENTIFIER ::= { id-pkix 3 }
> 
>    id-kp-serverAuth             OBJECT IDENTIFIER ::= { id-kp 1 }
>    -- TLS WWW server authentication
>    -- Key usage bits that may be consistent: digitalSignature,
>    -- keyEncipherment or keyAgreement
> 
>    id-kp-clientAuth             OBJECT IDENTIFIER ::= { id-kp 2 }
>    -- TLS WWW client authentication
>    -- Key usage bits that may be consistent: digitalSignature
>    -- and/or keyAgreement
> 
>    id-kp-codeSigning             OBJECT IDENTIFIER ::= { id-kp 3 }
>    -- Signing of downloadable executable code
>    -- Key usage bits that may be consistent: digitalSignature
> 
>    id-kp-emailProtection         OBJECT IDENTIFIER ::= { id-kp 4 }
>    -- Email protection
>    -- Key usage bits that may be consistent: digitalSignature,
>    -- nonRepudiation, and/or (keyEncipherment or keyAgreement)
> 
>    id-kp-timeStamping            OBJECT IDENTIFIER ::= { id-kp 8 }
>    -- Binding the hash of an object to a time
>    -- Key usage bits that may be consistent: digitalSignature
>    -- and/or nonRepudiation
> 
>    id-kp-OCSPSigning            OBJECT IDENTIFIER ::= { id-kp 9 }
>    -- Signing OCSP responses
>    -- Key usage bits that may be consistent: digitalSignature
>    -- and/or nonRepudiation
>

cpu:
[...]

PR #497, opened 2020-11-18
Added a new lint and tests for correlation between KU&EKU

ZsofiaTomicsko:
Hi!

I have created a new lint for issue #473 . Could someone please review it for me? :)


ZsofiaTomicsko:
@christopher-henderson Could you maybe review this pr, when you have some time for it? Thank you!

v3/lints/rfc/lint_key_usage_and_extended_key_usage_inconsistent.go
```diff
@@ -0,0 +1,195 @@
...
+
+//CheckConsistencyWithEKUEmailProtection checks if KU bits are consistent with Email Protection EKU bit
+func CheckConsistencyWithEKUEmailProtection(c *x509.Certificate) bool {
+	// 	RFC 5280 4.2.1.12 on KU consistency with Email Protection EKU:
+	// 	  -- Email protection
+	//    -- Key usage bits that may be consistent: digitalSignature,
+	//    -- nonRepudiation, and/or (keyEncipherment or keyAgreement)
+	//  Note: Recent editions of X.509 have renamed nonRepudiation bit to contentCommitment
+
+	allowedCombinations := []x509.KeyUsage{
```
christopher-henderson, line 134: Of course, Go doesn't have sets natively (which is a bit absurd imho) but the idiomatic alternative tends to be a `map[<T>]bool` which I think would work well in these cases as it obviates the need for helper function.

```suggestion
	return map[x509.KeyUsage]bool{
		x509.KeyUsageDigitalSignature:                                                                 true,
		x509.KeyUsageContentCommitment:                                                                true,
		x509.KeyUsageKeyEncipherment:                                                                  true,
		x509.KeyUsageKeyAgreement:                                                                     true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageContentCommitment:                                true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment:                                  true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageKeyAgreement:                                     true,
		x509.KeyUsageContentCommitment | x509.KeyUsageKeyEncipherment:                                 true,
		x509.KeyUsageContentCommitment | x509.KeyUsageKeyAgreement:                                    true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageContentCommitment | x509.KeyUsageKeyEncipherment: true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageContentCommitment | x509.KeyUsageKeyAgreement:    true,
	}[c.KeyUsage]
```

christopher-henderson (commented):
So I'm working through this and by far the most difficult part is parsing out the RFC author's usage of plain English boolean expressions. The trouble here is that the plain English usage of "or" tends to actually be xor. Take the following:

"What would you like, cake or pie?"

To the brain of a computer, this means that you can have either cake, pie, or both. However, what our baker actually meant was "pick exactly one of either cake or pie - you can't have both".

Additionally, I'm findind thier usage of commas with trailing boolean expressions a bit tricky. It SEEMS to me that each comma is to be replaced with the trailing boolean expresion. Take for example...

`A, B, and/or (C or D)`

To put this functionally, it would be...

`andor(A, andor(B, or(C, D)))`

And what, exactly, is and/or? This seems like the traditional boolen definition of "or" (that is, `this-or-that-or-both`). So should the above actually be:

`or(A, or(B, xor(C, D)))`

Let's take the most complex one - email protection.

`or(DigitalSignature, or(NonRepudiation, xor(KeyEncipherment, KeyAgreement))`

...I'm almost tempted write some util functions just to help automate building these truth tables because
this seems very easy to screw up.






#

Additionally I'm a bit hesitant to mark this as an error (although I'm leaning more toward error as I type this) because of the word "used" in...

> the certificate MUST only be used for a purpose consistent with both extensions.

To me, this wording implies that this is a MUST on the implementation that is accepting the certificate and not a MUST on the certificate itself (after all, how can this lint know how the certificate is being used as that is a runtime characterstic). Also, the often repeated weasel word of _**lower case**_ "may":

> Key usage bits that may be consistent...

This wording is telling me, "by the way, here are some _example_ pairings, but you may have more".

Restated, I'm reading this as saying

> The KUs and the EKUs MUST be consistent, but what "consistent" is is left as a policy of the accepting application.

...but then the following clause uses the all important word - "defined"...

> The following key usage purposes are defined

...this seems to say, "...but at minimum, you must use these". So, as long as the ones defined here are bare minimum examples then I suppose this should be an error.

ZsofiaTomicsko:
Thank you for your review!

### 1. Map[< T >]bool solution:

I have changed the code accordingly and I will push the changes soon.

### 2. Parsing the RFC:

Yes, that's how i understood the RFC too. I can't think of another way to interpret "and/or" other than the traditional "or" in computer language. And because of this I have to assume that by "or" the author meant a different "or", probably the one used in plain English: ```this-or-that``` as you have explained.

I thought of writing util functions too but since the only truly complex case seemed to be the Email Protection case I figured that maybe the code would be more readable this way. I can definitely write something like that if you think it would be a good idea. I'm just a bit unsure about what the input should be. The RFC text maybe? For example a string of ```digitalSignature, nonRepudiation, and/or (keyEncipherment or keyAgreement)```

### 3. Error or Warning:

I agree with what you said, but I don't understand why you reached the conclusion that it should be marked as an error in the end if you consider the list of the EKUs a bare minimum example. Now i'm leaning towards changing the lint to a warning instead. From your reply I got the feeling that you think that the list states what must be considered consistent for a few EKUs, but there could be other combinations that the accepting application could see as consistent. In my opinion marking this as an error would only make sense if we interpreted that "may" word in 

> Key usage bits that may be consistent: digitalSignature, keyEncipherment or keyAgreement

 as these are the KU bits that are recommended to view as consistent, but the application could decide to accept **less** . For example the accepting application could decide to not accept keyAgreement, but only digitalSignature, and key Encipherment.

> "by the way, here are some example pairings, but you may have more"

If I interpret the list as some example pairings, but there could be other combinations, then I think this should be marked as a warning, since the certificate isn't necessarily wrong if the two extensions aren't considered consistent according to the RFC since some applications could still accept the certificate.

christopher-henderson:
(Thank you for your patience. Even moderately long boolean expressions can be difficult for us lowly humans to keep straight in our head. Doubly so here since `nonRepudiation` in the RFC is renamed `contentCommitment` in the x509 library which is even more cognitive load.)

So I actually did end up writing a small bit of code to help confirm these truth tables (honestly just the email one was a bit difficult to sit down and stare at). If you would please, I would like to have the suite of `CheckConsistencyWithEKU<WHATEVER>` functions changed to the simple map definitions that you have graciously built and then write some tests against those hard coded truth tables.

That is, can we remove the function definition of...

```go
func CheckConsistencyWithEKUEmailProtection(x *x509.Certificate) bool {} 
```

...and replace it with a simple private mapping...

```go
// Checks if KU bits are consistent with Email Protection EKU bit
// 	RFC 5280 4.2.1.12 on KU consistency with Email Protection EKU:
// 	  -- Email protection
//    -- Key usage bits that may be consistent: digitalSignature,
//    -- nonRepudiation, and/or (keyEncipherment or keyAgreement)	
//  Note: Recent editions of X.509 have renamed nonRepudiation bit to contentCommitment
var emailProtection = map[x509.KeyUsage]bool{
		x509.KeyUsageDigitalSignature:                                                                 true,
		x509.KeyUsageContentCommitment:                                                                true,
		x509.KeyUsageKeyEncipherment:                                                                  true,
		x509.KeyUsageKeyAgreement:                                                                     true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageContentCommitment:                                true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageKeyEncipherment:                                  true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageKeyAgreement:                                     true,
		x509.KeyUsageContentCommitment | x509.KeyUsageKeyEncipherment:                                 true,
		x509.KeyUsageContentCommitment | x509.KeyUsageKeyAgreement:                                    true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageContentCommitment | x509.KeyUsageKeyEncipherment: true,
		x509.KeyUsageDigitalSignature | x509.KeyUsageContentCommitment | x509.KeyUsageKeyAgreement:    true,
}
```

...and update it's usage to be...

```go
switch {
...
case x509.ExtKeyUsageEmailProtection:
	if !emailProtection[c.KeyUsage] {
	}
...
}
```

And from that, we can write tests to make sure that our hardoded truth tables are easier to verify:

```go
func TestEKUEmailProtection(t *testing.T) {
	got := KeyUsage(x509.KeyUsageDigitalSignature).
		Or(KeyUsage(x509.KeyUsageContentCommitment).
			Or(KeyUsage(x509.KeyUsageKeyEncipherment).Xor(KeyUsage(x509.KeyUsageKeyAgreement))))
	for w := range emailProtection {
		if !got[KeyUsage(w)] {
			t.Errorf("expected %d to be present in the computed map", w)
		}
	}
	for g := range got {
		if !emailProtection[x509.KeyUsage(g)] {
			t.Errorf("expected %d to be present in the pre-computed map", g)
		}
	}
}

type Truther interface {
	Or(a Truther) Accepted
	Xor(a Truther) Accepted
}

type KeyUsage x509.KeyUsage
type Accepted map[KeyUsage]bool

func (usage KeyUsage) Or(truther Truther) Accepted {
	if truther == nil {
		return Accepted{usage: true}
	}
	switch t := truther.(type) {
	case KeyUsage:
		return map[KeyUsage]bool{
			usage:     true,
			t:         true,
			usage | t: true,
		}
	case Accepted:
		t.Or(usage)
		return t
	default:
		panic("")
	}
}

func (usage KeyUsage) Xor(truther Truther) Accepted {
	if truther == nil {
		return Accepted{usage: true}
	}
	switch t := truther.(type) {
	case KeyUsage:
		return Accepted{
			usage: true,
			t:     true,
		}
	case Accepted:
		t.Xor(usage)
		return t
	default:
		panic("")
	}
}

func (accepted Accepted) Or(truther Truther) Accepted {
	if truther == nil {
		return accepted
	}
	switch other := truther.(type) {
	case KeyUsage:
		accepted[other] = true
		for current := range accepted {
			accepted[current|other] = true
		}
	case Accepted:
		for key := range other {
			accepted[key] = true
			for inner := range accepted {
				accepted[key|inner] = true
			}
		}
	}
	return accepted
}

func (accepted Accepted) Xor(truther Truther) Accepted {
	if truther == nil {
		return accepted
	}
	switch other := truther.(type) {
	case KeyUsage:
		for key := range accepted {
			accepted[key|other] = true
		}
		accepted[other] = true
	case Accepted:
		for key := range other {
			accepted.Xor(key)
		}
	}
	return accepted
}
```

christopher-henderson:
>  are recommended to view as consistent, but the application could decide to accept **less**

The reason why I backed off from making it a warning is because I started read it as:
 
> ...you _**must**_ honor at least these but you **_may_** have more.

sleevi:
Does RFC 8813 help address some of the this-or-that confusion? The intent
is *only* the usages in the enumerated set; any other usage is not
consistent.

christopher-henderson:
I don't think RFC 8813 has quite what we're looking for, but the insight that these usages are the _only_ acceptable combinations is quite useful - thank you @sleevi!

christopher-henderson (approved):
Thank you very much!

christopher-henderson:
@zakird this one looks good to me (still not think I'm allowed to click the merge button - maybe I'm missing something).

ZsofiaTomicsko:
Sorry I had to leave for an urgent last-minute Christmas shopping before i could finish :) I will add the tests in a bit too. And I wanted to change up the comments. I will push 2 more commits soon. And thank you for the help!

christopher-henderson:
> Sorry I had to leave for an urgent last-minute Christmas shopping before i could finish

😄 Christmas trumps code every time (at least to me, anyways).

I need to figure what would be a good workflow for opening PRs against PRs that are coming from a fork. Normally I would just write what I had in mind myself and contribute rather than just asking, but having the branch come from a fork is complicating that.

PR #506, opened 2020-11-28
lints: split Apple cert lifetime lint per-result.

### LINKED ISSUE #496: Split e_tls_server_cert_valid_time_longer_than_398_days into a `e_` and `w_` lint.
Per [recent updates](https://github.com/zmap/zlint/commit/29b3fa90013c35c61f635f58f0d306a23a169a86) to the contrib guidelines  the [`e_tls_server_cert_valid_time_longer_than_398_days` lint](https://github.com/zmap/zlint/blob/2b994a741bfde638a648a759c16644724d07e708/v2/lints/apple/lint_e_server_cert_valid_time_longer_than_398_days.go) should be split into two lints: one that returns an error and one that returns a warning.

We should try to do this ahead of the next major release since modifying the existing lint has compat. implications.

cpu:
Previously the `e_tls_server_cert_valid_time_longer_than_398_days` lint could return either an error result or a warning result.  We prefer having lints return one status level only.

This commit breaks the lint up so that `e_tls_server_cert_valid_time_longer_than_398_days` only handles the error case and a new `w_tls_server_cert_valid_time_longer_than_397_days` lint handles the warning case.

Resolves #496

### DIFF
--- a/v2/lints/apple/lint_w_server_cert_valid_time_longer_than_397_days.go
+++ b/v2/lints/apple/lint_w_server_cert_valid_time_longer_than_397_days.go
@@ -22,27 +22,33 @@ import (
 	"github.com/zmap/zlint/v2/util"
 )
 
-type serverCertValidityTooLong struct{}
+type serverCertValidityAlmostTooLong struct{}
 
-func (l *serverCertValidityTooLong) Initialize() error {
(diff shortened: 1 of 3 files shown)


zakird (approved):

cpu:
Thanks for the :mag:

Issue #507, opened 2020-11-28
Implement zlint-autopull with Github Actions


cpu:
Presently the automation that invokes the `zlint-gtld-update` binary periodically and creates PRs with the resulting diff lives in a separate `zlint-autopull` repo: https://github.com/cpu/zlint-autopull It's also based on Travis, where cron builds were limited to at most once a day.

This feels like something that could be converted to a Github Action. Potentially even removing the `tld-update-bot`'s access to this repository.

Issue #512, opened 2020-11-30
TLD removal auto-pr missing for "shriram"


ezekiel:
"shriram" has a removal date of `2020-11-24`. Likely that @tld-update-bot just needs a nudge?

cpu:
@ezekiel hey there :-) :wave:

Thanks for flagging. I took a peek at the [automation Travis](https://travis-ci.com/github/cpu/zlint-autopull) and I think the last build was ~7d ago which would explain a missing update from the 24th of Nov. There's a few problems with the Travis automation:

* The Travis integration somehow got disabled (???) and I had to click a button to re-enable. That should be fixed now.
* The build I kicked off was broken because it uses HEAD of master w/ a `v2` path and https://github.com/zmap/zlint/commit/fe65bae28cbc112082e21449a373f7ccf059422f moved the `zlint-tld-update` script. I [fixed this](https://github.com/cpu/zlint-autopull/commit/3836de0c0cb0bbf2f4988cdb983312e287cbbf12) in master.
* Travis is lagging, or not picking up the new commit. I can't seem to trigger a build with the fix for the `v3` tool path.

Bitrot comes for us all.

I already have a Github Action based version of this automation [written up](https://github.com/cpu/zlint/commit/0bc791a414b8f2d059bc0a912ed458797c696c5a) for https://github.com/zmap/zlint/issues/507 that worked successfully last night in my [test repo](https://github.com/cpu/zlint/pull/5).

After work today I'll open a PR to get ZLint switched to using that integration instead of the Travis one. I think that's the quickest way to get this fixed if Travis doesn't queue a fixed build soon.

cpu:
>  I think that's the quickest way to get this fixed if Travis doesn't queue a fixed build soon.

(Or I could stamp a one-off manually created PR for this fix if it's important to land in <12-24hr)

ezekiel:
:wave: hello!

Please take your time, no urgency from me for merging. Happy to wait and to see things ironed out. Thank you!

cpu:
> After work today I'll open a PR to get ZLint switched to using that integration instead of the Travis one.

https://github.com/zmap/zlint/pull/513 - once this is merged we should see a PR with the `shriram` removal update in <1hr.

cpu:
Resolved by https://github.com/zmap/zlint/pull/514

We have a couple little things to sort out with the new automation but it should be reliable enough now to get regular updates again. Will monitor.

PR #513, opened 2020-11-30
CI: Add cron workflow for gTLD update PRs, deprecate tld-update-bot.

cpu:
### Description

Previously we used a separate repo with [a bash script](https://github.com/cpu/zlint-autopull/blob/master/autopull) hooked up to a [bot github user](https://github.com/tld-update-bot) in a Travis CI cron build to automatically create PRs updating ZLint TLD data on a periodic basis using the `zlint-gtld-update` tool from this repo. Recently it's been having reliability issues that made us miss a TLD update: https://github.com/zmap/zlint/issues/512 

Now that we're using Github Actions we can make things much simpler and self-contained. This commit adds a `tld-update.yml` workflow that uses a [create-pull-request Github action](https://github.com/peter-evans/create-pull-request) to replace the separate repo/bash script/bot user approach.

Resolves https://github.com/zmap/zlint/issues/507

### Notable Improvements

* Reduced dependencies - we can drop the Travis integration, the @tld-update-bot account, and the bot's access to this repo.
* Consolidation - all of the configuration and logic live in the ZLint repo now as opposed to being spread across the ZLint repo, an automation repo (in a separate Github account!), and Travis CI env vars.
* More flexible scheduling - the Github Actions cron schedule is more flexible so we can run the new action once every hour instead of just once a day like the Travis version.
* Smarter integration - unlike the bash script this version shouldn't recreate the same PR over and over if it isn't merged in a timely manner.

### Testing

To test the integration I pushed a commit to a fork of ZLint that [removed some of the gTLD map data](https://github.com/cpu/zlint/commit/195e0fcc184d85495da26de7672e1f28844384f6). Afterwards I updated the action to run on push of a specific test branch in addition to the cron schedule. Pushing that test branch resulted in a correct automation PR being opened to add back the removed data (and the missed update from #512). You can see [the PR that was opened here](https://github.com/cpu/zlint/pull/5). No bot account or bash scripts required :tada:

### DIFF
--- a/.github/workflows/tld-update.yml
+++ b/.github/workflows/tld-update.yml
@@ -1,0 +1,46 @@
+name: tld-update
+on:
+  schedule:
+    # Run every hour, at the 15 minute mark. E.g.
+    # 2020-11-29 00:15:00 UTC, 2020-11-29 01:15:00 UTC, 2020-11-29 02:15:00 UTC
+    - cron:  '15 * * * *'
+jobs:
+  zlint-gtld-update:
+    name: Check for TLD data updates
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+
+      - name: Set current date as env variable
+        run: echo "NOW=$(date +'%Y-%m-%dT%H:%M:%S %Z')" >> $GITHUB_ENV
+
+      - name: Install zlint-gtld-update
+        run: go install ./cmd/zlint-gtld-update/...
+        working-directory: v3
+
+      - name: Run go-generate
+        run: go generate ./...
+        working-directory: v3
+
+      - name: Create pull-request
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: "util: gtld_map autopull updates for ${{ env.NOW }}"
+          title: "util: gtld_map autopull updates for ${{ env.NOW }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/tld-update.yml
```diff
@@ -0,0 +1,46 @@
...
+
+      - name: Create pull-request
+        id: cpr
+        uses: peter-evans/create-pull-request@v3
+        with:
+          commit-message: "util: gtld_map autopull updates for ${{ env.NOW }}"
+          title: "util: gtld_map autopull updates for ${{ env.NOW }}"
+          body: "ZLint gTLD data updates from `go generate ./...` for ${{ env.NOW }}."
+          labels: tld-update
+          branch: zlint-gtld-update
```
zakird, line 40: What happens if we have two unmerged PRs? Do we want to use the same branch name every time?

cpu: I haven't tested this part yet (it's trickier to arrange) but my understanding is that the bot will continue to append commits to the open PR until it's merged.

The action author has a longer description of how this works here: https://github.com/peter-evans/create-pull-request#action-behaviour

cpu: The "Alternative strategy - Always create a new pull request branch" is closer to what we had before and it's not very good for this use-case.

zakird: That sounds reasonable to me -- mostly just wanted to make sure wouldn't crash.

zakird (approved):

.github/workflows/tld-update.yml
cpu: Happy to iterate further based on our experience if bugs show up.

cpu:
I removed @tld-update-bot's access to the ZLint repo.

PR #527, opened 2020-12-26
Fail certificates whose SN would DER encode past 20 octets

### LINKED ISSUE #502: lint_serial_number_longer_than_20_octets does not check the encoded value length of the serial number
The serial number length check in 
https://github.com/zmap/zlint/blob/71e2966bbbbd8a3ae6e7ff5f07502f6f59926757/v2/lints/rfc/lint_serial_number_longer_than_20_octets.go#L51 permits overly long serial numbers. Namely, serial numbers consisting of 160 bits with the MSB set to 1 will require a leading 0x00 octet in the encoded value to denote the sign, which yields an encoded length of 21 octets; the current logic does not account for this leading octet. While RFC 5280 is not clear on whether this is a violation, discussion on MDSP (specifically this comment: https://groups.google.com/g/mozilla.dev.security.policy/c/b33_4CyJbWI/m/flGvK3DeBgAJ) indicates that several client implementations will not accept certificates with a serial number whose encoded value exceeds 20 octets.

In order for certificates that pose interoperability concerns to be appropriately flagged, the check should cause the lint to error out if the BitLen of the serial number value exceeds 159 bits.

christopher-henderson:
This resolves #502.


v3/lints/rfc/lint_serial_number_longer_than_20_octets_test.go
```diff
@@ -38,3 +38,12 @@ func TestSnNotTooLarge(t *testing.T) {
 		t.Errorf("%s: expected %s, got %s", inputPath, expected, out.Status)
 	}
 }
+
+func TestSnTooLargeDueToSignedMSB(t *testing.T) {
```
cpu, line 42: Not strictly necessary but it would be nice if you could rework the three separate test functions in a table-driven style. We generally prefer those for new tests.

cpu (changes requested):
LGTM modulo the integration test work. Thanks!

cpu:
> This addresses issue #502.

Editorial note: If you rephrase this slightly as "Resolves #502" Github will auto-close the issue when this is merged. Presently it's not picking up the linkage based on your current desc and won't.

zakird:
```
// The maximum number of octets is 20 (160 bits), however there is a
// a complication when the serial number is exactly 160 bits long,
// and the MSB is 1, wherein implementations often prefix the serial
// with 0x00 in order to clearly signify the sign, thus putting the
// encoding past the octet limit.
```

Does this have the potential for false positives? Not quite sure what to take of the "often prefix". Would your certificate be compliant if you're in one of the limitations that doesn't prefix, and we'd accidentally trigger?

cpu:
> Does this have the potential for false positives? Not quite sure what to take of the "often prefix". Would your certificate be compliant if you're in one of the limitations that doesn't prefix, and we'd accidentally trigger?

@zakird I found the wording of this comment a little hard to parse too. Could probably use some rephrasing.

If I'm understanding correctly if an implementation didn't prefix a 160 bit serial number with a positive MSB with a leading 0 then it will end up being parsed by the stdlib `asn1` decoder as a negative `big.Int` number in `parseBigInt` (which will trigger a separate linter).

christopher-henderson:
> Does this have the potential for false positives? Not quite sure what to take of the "often prefix".

Well, a DER integer is a signed integer, so I suppose that this shouldn't be "often prefix" but rather "must prefix". It could be a false positive if the serial number is negative. Even though that is also a failing condition, this lint should be aware of it.

v3/lints/rfc/lint_serial_number_longer_than_20_octets.go
```diff
@@ -48,7 +48,18 @@ func (l *serialNumberTooLong) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *serialNumberTooLong) Execute(c *x509.Certificate) *lint.LintResult {
-	if c.SerialNumber.BitLen() > 160 { // 20 octets
+	if c.SerialNumber.BitLen() > 159 {
```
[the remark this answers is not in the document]
christopher-henderson: Chunking through a handful of these and unfortunately they are making sense so far (SNs that are encoding to 21 octets).

I'll keep thumbing through these to find inconsistencies, but at a glance this issue does appear to be somewhat endemic.

zakird:
Got it, thanks for the clarification. Might make sense to update the comment to that affect.

v3/lints/rfc/lint_serial_number_longer_than_20_octets_test.go
christopher-henderson: Improvements are always welcome!

christopher-henderson:
@zakird and @cpu I realized that since what we're really talking about here is the _encoded_ length of the SN that it might be better to simply encode the value and check its length rather than trying to predict its encoded length. Note that this does put a bit more reliance upon the stdlib implementation, although it does show the behavior of prefixing with a `00` byte when necessary. Lemme know what you think of switching this strategy up.

Here as an example test output string:

> The DER encoded certificate serial number is 21 octets long. If this is surprising to you, note that DER integers are signed and that SNs that are 20 octets long with an MSB of 1 will be automatically prefixed with 0x00, thus bumping it up to 21 octets long. SN: 0088D27CBF0003174B1C7EE0208EDB668BE7A06199

v3/lints/rfc/lint_serial_number_longer_than_20_octets.go
```diff
@@ -48,8 +51,19 @@ func (l *serialNumberTooLong) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *serialNumberTooLong) Execute(c *x509.Certificate) *lint.LintResult {
-	if c.SerialNumber.BitLen() > 160 { // 20 octets
-		return &lint.LintResult{Status: lint.Error}
+	sn, err := asn1.Marshal(c.SerialNumber)
+	if err != nil {
+		return &lint.LintResult{Status: lint.Fatal, Details: fmt.Sprint(err)}
+	}
+	// ASN1 integers are [tag (2), length, octets...]
```
christopher-henderson, line 58: ...ish, need to take into account long form length encodings.

v3/lints/rfc/lint_serial_number_longer_than_20_octets.go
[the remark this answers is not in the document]
christopher-henderson: I fed all 190 through a spot checker and sure enough all of them had exactly the maximum bitlength that overflows after encoding. Updated the integration test expectations.

cpu:
@christopher-henderson Thanks, will take a look at your new approach shortly.

zakird (approved):
Looks good to me. Will let @cpu merge since he has previous changes requested on the PR.

cpu:
> Will let @cpu merge since he has previous changes requested on the PR.

Thanks, haven't forgotten about this one (& saw the other @). Will make sure both get reviews this week.

cpu (approved):
Thanks @christopher-henderson, this looks good to me.

PR #528, opened 2020-12-29
KU and EKU Inconsistent lint correction

ZsofiaTomicsko:
Hi!

#497 was closed before I could add the last 2 commits.

1. I have removed the comments explaining the previously removed functions and added new ones
2. I added the tests for the truth tables

@christopher-henderson I have changed a small detail in your code that I think was a typo: 

```
func (accepted Accepted) Xor(truther Truther) Accepted {
	if truther == nil {
		return accepted
	}
	switch other := truther.(type) {
	case KeyUsage:
		for key := range accepted {
			accepted[key|other] = true
		}
		accepted[other] = true
	case Accepted:
		for key := range other {
			accepted.Xor(key)
		}
	}
	return accepted
}
```

In the case of the input being a KeyUsage I removed:
``` 
for key := range accepted {
			accepted[key|other] = true
		}
```

I didn't know where to put the xor and or functions so I just copied them inside the test file too. Should I maybe move them elsewhere?


christopher-henderson:
Thank you for revisiting this!

> I didn't know where to put the xor and or functions so I just copied them inside the test file too. Should I maybe move them elsewhere?

You're right about the bug in my code and honestly the test is more complex itself, so I think keeping it private to this file is for the best as it's really not battle hardened for general use.

christopher-henderson (approved):

christopher-henderson:
@zakird ready to merge these truth table tests (say that ten time fast).

zakird (approved):

Issue #531, opened 2021-01-03
CABF EV Guidelines 1.7.0 Section 9.8.2 Registration Scheme


christopher-henderson:
[CA-Browser-Forum-EV-Guidelines-v1.7.0](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-EV-Guidelines-v1.7.0.pdf) Section 9.8.2

> The Registration Scheme MUST be encoded as described by the following ASN.1 grammar...

PR #532, opened 2021-01-03
CABF EV Guidelines 1.7.0 Section 9.8.2 Organization ID Present

### LINKED ISSUE #530: CABF EV Guidelines 1.7.0 Section 9.8.2 Organization ID Present
[CA-Browser-Forum-EV-Guidelines-v1.7.0](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-EV-Guidelines-v1.7.0.pdf) Section 9.8.2

> Effective January 31, 2020, if the subject:organizationIdentifier field is present, this field MUST be present.

christopher-henderson:
Resolves #530 

This PR obviates #399 which was attempting to do far more than the simple contains relationship asked of by the CABF clause. This change does _not_ address #531.


zakird (approved):

zakird:
This looks good. Do the documents also specify that the values need to be the same? I suspect we'll also want to add parsing for the extension into ZCrypto, though we don't need to block this PR on that enhancement.

sleevi:
> This looks good. Do the documents also specify that the values need to be the same?

Yes. [EVG 9.8.2](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-EV-Guidelines-v1.7.4.pdf#page=23) states

> The Registration Scheme MUST be encoded as described by the following ASN.1 grammar:
> <snip>
> where the subfields have the same values, meanings, and restrictions described in Section 9.2.8. The CA SHALL validate the contents using the requirements in Section 9.2.8.

zakird:
If that's the case, it would seem that a lint that checks that these are the same would be better since that would fully supersede the check here. @christopher-henderson what would you think about implementing and checking that instead of this check?

v3/util/time.go
```diff
@@ -42,6 +42,7 @@ var (
 	CABV116Date                 = time.Date(2013, time.July, 29, 0, 0, 0, 0, time.UTC)
 	CABV130Date                 = time.Date(2015, time.April, 16, 0, 0, 0, 0, time.UTC)
 	CABV131Date                 = time.Date(2015, time.September, 28, 0, 0, 0, 0, time.UTC)
+	CABV170Date                 = time.Date(2020, time.January, 31, 0, 0, 0, 0, time.UTC)
```
sleevi, line 45: You can see https://github.com/zmap/zlint/pull/449#discussion_r436397811 for some discussion, but I think this might make more sense with a descriptive name about the requirement effective-date rather than the document effective date. WDYT?

v3/lints/cabf_ev/lint_ev_organization_id_missing_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/test"
+)
+
+var tests = map[string]lint.LintStatus{
+	"EvAltRegNumCert52NoOrgId.pem":                       lint.NA,
+	"EvAltRegNumCert55OrgIdExtMissing.pem":               lint.NE,
+	"EvAltRegNumCert67ValidNtrWithOrgIdExt.pem":          lint.NE,
+	"EvAltRegNumCert55OrgIdExtMissingNewerDate.pem":      lint.Error,
+	"EvAltRegNumCert67ValidNtrWithOrgIdExtNewerDate.pem": lint.Pass,
```
sleevi, line 29: Do the numbers in the filenames have meaning here? Should they?

I don't have strong feelings, but it seems that `EvAltregNumCertXX` is... less than descriptive. And the non-sequentially of these suggests that any implied semantics (e.g. 52 < 55) would be defeated if 53/54 are introduced.

christopher-henderson: > Do the numbers in the filenames have meaning here? Should they?

I admit that I wondered the exact same thing as I simply took the test data from #399 and spot checked them, but otherwise forgot about the filenames. I'll try to figure out if there are any semantics attached there, otherwise I'll just rename these.

v3/util/time.go
christopher-henderson: Okay so I wasn't the _only_ one confused about this (neither the version number nor the ballots were perfectly lining up on this)!

Perhaps a date for the document effective date and aliases for specific requirements?

```go
const (
    // https://cabforum.org/link/to/doc.pdf
    CABV123 = ...
    MustBeTotallyCool = CABV123
    MustBeMostRadical = CABV123
)
```

That is, it signals that these specific requirements were all introduced in this parent document with the breadcrumb to the PDF.

sleevi: Yeah, although it’s possible for one version to have different effective dates for different requirements (like the browser alignment ballot, IIRC), but that SGTM.

v3/lints/cabf_ev/lint_ev_organization_id_missing.go
```diff
@@ -0,0 +1,49 @@
+/*
+ * ZLint Copyright 2020 Regents of the University of Michigan
```
cpu, line 2: :joy: I think we'll need to do this across the codebase/template.

```suggestion
 * ZLint Copyright 2021 Regents of the University of Michigan
```

v3/lints/cabf_ev/lint_ev_organization_id_missing.go
```diff
@@ -0,0 +1,49 @@
...
+	return nil
+}
+
+func (l *evOrgIdExtMissing) CheckApplies(c *x509.Certificate) bool {
+	return util.IsEV(c.PolicyIdentifiers) && len(c.Subject.OrganizationIDs) > 0
+}
+
+func (l *evOrgIdExtMissing) Execute(c *x509.Certificate) *lint.LintResult {
+	if !util.IsExtInCert(c, util.CabfExtensionOrganizationIdentifier) {
+		return &lint.LintResult{Status: lint.Error, Details: "subject:organizationIdentifier field is present in an EV certificate but the CA/Browser Forum Organization Identifier Field Extension is missing"}
```
cpu, line 35: small nit: can you break this up across a few lines?

v3/lints/cabf_ev/lint_ev_organization_id_missing_test.go
```diff
@@ -0,0 +1,39 @@
+package cabf_ev
+
+/*
+ * ZLint Copyright 2020 Regents of the University of Michigan
```
cpu, line 4:
```suggestion
 * ZLint Copyright 2021 Regents of the University of Michigan
```

v3/lints/cabf_ev/lint_ev_organization_id_missing_test.go
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import (
+	"testing"
+
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/test"
+)
+
+var tests = map[string]lint.LintStatus{
```
cpu, line 24: tiny nit: is there a reason to make this a global instead of a var scoped to `TestOrganizationIDMissing`?

v3/lints/cabf_ev/lint_ev_organization_id_missing_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"EvAltRegNumCert52NoOrgId.pem":                       lint.NA,
+	"EvAltRegNumCert55OrgIdExtMissing.pem":               lint.NE,
+	"EvAltRegNumCert67ValidNtrWithOrgIdExt.pem":          lint.NE,
+	"EvAltRegNumCert55OrgIdExtMissingNewerDate.pem":      lint.Error,
+	"EvAltRegNumCert67ValidNtrWithOrgIdExtNewerDate.pem": lint.Pass,
+}
+
+func TestOrganizationIDMissing(t *testing.T) {
+	for file, want := range tests {
```
cpu, line 33: Can you implement this using `t.Run()` with each `tests` instance as [a subtest](https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks)? It makes the output much nicer, especially in the presence of failures.

cpu (changes requested):
Overall looks good to me, just a couple small things to request.

v3/lints/cabf_ev/lint_ev_organization_id_missing.go
christopher-henderson: Proof that TV shows like Silicon Valley has no idea what our day-to-day lives are actually like :joy: 

I'll give this a go in another PR.

> And engineer has a problem. He decides to use a regular expression. He now has two problems.

v3/lints/cabf_ev/lint_ev_organization_id_missing_test.go
christopher-henderson: None whatsoever, making in locally scoped now.

v3/lints/cabf_ev/lint_ev_organization_id_missing_test.go
christopher-henderson: Sure! That's a good idea.

v3/lints/cabf_ev/lint_ev_organization_id_missing.go
christopher-henderson: Can and will.

christopher-henderson:
@zakird 
 
 > what would you think about implementing and checking that instead of this check?
 
 I actually think that maybe that should be a separate check? Keeping to the notion of one lint per clause, I suppose. I think there might actually be at least three lints here:
 
 1. If subjectOrgID exists then this CABF extension must exist (this PR).
 2. If this CABF extension exists then it must be of the given the ASN1 grammar (although I suppose this handled naturally by parsing it).
 3. If this CABF extension exists then it must have the same value as the subjectoOrgID.
 
 That all gets a wrench thrown into it, however, if this CABF extension can exist without the subjectOrgID being present and be still be legal.
 
 That is, would it be legal for the CABF extension to exist without the presense of the subjectOrgID. If so, then asserting the same value would be a false positive in this case. Although this particular question may be getting into the weeds of a different lint altogether...

cpu (approved):
Thanks @christopher-henderson

cpu:
I will squash merge once CI completes.

timfromdigicert:
It should be legal for the cabfOrganizationIdentifier to exist without the corresponding OrganizationIdentifier.  It's clearly marked "optional" in the EV Guidelines and there are no requirements in the EV Guidelines or elsewhere prohibiting it from being there. 
 The whole reason it exists was because the technical ASN.1 encoding was viewed as superior to the string-based encoding standardized by ETSI.  It would be odd to prohibit the preferred form on the grounds that an inferior encoding is not present, especially since the hope was that people would migrate to the preferred form over time.

sleevi:
@timfromdigicert Could you perhaps clarify your comment. Do you believe this lint is causing a false positive?

Or were you replying to https://github.com/zmap/zlint/pull/532#issuecomment-757361826 , which noted (emphasis added):
> That is, would it be legal for the CABF extension to exist without the presense of the subjectOrgID. If so, then asserting the same value would be a false positive in this case. **Although this particular question may be getting into the weeds of a different lint altogether...**

and just trying to highlight this permissiveness for that future lint?

timfromdigicert:
Sorry, I was commenting on "That is, would it be legal for the CABF extension to exist without the presense of the subjectOrgID".

PR #533, opened 2021-01-05
New lint for checking if the name constraints for URIs are FQDN

ZsofiaTomicsko:
Hi!

I wrote a small code to check if the name constraints are FQDNs. Here is the text from RFC5280:

> For URIs, the constraint applies to the host part of the name.  The
   constraint MUST be specified as a fully qualified domain name and MAY
   specify a host or a domain.  Examples would be "host.example.com" and
   ".example.com".  When the constraint begins with a period, it MAY be
   expanded with one or more labels.  That is, the constraint
   ".example.com" is satisfied by both host.example.com and
   my.host.example.com.  However, the constraint ".example.com" is not
   satisfied by "example.com".  When the constraint does not begin with
   a period, it specifies a host.


v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (l *nameConstraintNotFQDN) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, subtreeString := range c.PermittedURIs {
+		var host = subtreeString.Data
+		if string(host[0]) == string(".") {
+			host = "l" + host
+		}
+		if !util.IsFQDN(host) {
+			return &lint.LintResult{Status: lint.Error}
```
zakird, line 40: It would be helpful to return the FQDN that flagged the error in the description.

zakird (approved):
Overall this looks good to me. It would be helpful to change the description to tell us what name(s) are wrong as is done in other lints.

v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (l *nameConstraintNotFQDN) CheckApplies(c *x509.Certificate) bool {
+	return util.IsExtInCert(c, util.NameConstOID)
+}
+
+func (l *nameConstraintNotFQDN) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, subtreeString := range c.PermittedURIs {
+		var host = subtreeString.Data
+		if string(host[0]) == string(".") {
+			host = "l" + host
```
cpu, line 39: Why prefix the host with "l" here? Would it be better to strip the leading "." instead? I think that might be less confusing to the reader. 

It also matches the approach `util.IsFQDN` takes for wildcards/question marks.

v3/lints/rfc/lint_name_constraint_not_fqdn_test.go
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import (
+	"testing"
+
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/test"
+)
+
+func TestBeginsWithPeridFQDN(t *testing.T) {
```
cpu, line 24: I won't consider it a blocker for merging but if you're up to it I'd prefer to see these unit test functions implemented as one test function that uses [a table-driven design](https://github.com/golang/go/wiki/TableDrivenTests).

cpu (changes requested):
Thanks @ZsofiaTomicsko ! 

Here's some review feedback from a first pass.

ZsofiaTomicsko:
Thank you very much for the feedback! :) I will correct the code.

v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,118 @@
...
+	incorrectPermittedHosts := make([]string, 0)
+	incorrectExcludedHosts := make([]string, 0)
+	errString := ""
+
+	for _, subtreeString := range c.PermittedURIs {
+		isHostFQDN(subtreeString.Data, &incorrectPermittedHosts)
+	}
+	for _, subtreeString := range c.ExcludedURIs {
+		isHostFQDN(subtreeString.Data, &incorrectExcludedHosts)
+	}
```
cpu, line 46: This looks a lot better! Thanks for the refactoring.

v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,118 @@
...
+		return &lint.LintResult{
+			Status:  lint.Error,
+			Details: errString,
+		}
+	}
+
+	return &lint.LintResult{Status: lint.Pass}
+}
+
+func isHostFQDN(host string, incorrectHosts *[]string) {
```
cpu, line 68: In the effort of keeping functions doing one job only I think it would be a little bit more natural to write this as a function that returns a `bool` for the `host` and doesn't do the job of collecting up the `incorrectHosts` (the caller would do that). That has a few advantages:

1. When folks see a function named "isXXX" the natural expectation is that it returns true/false.
1. It avoids the need to pass the `incorrectHosts` slice in by reference so that this function can mutate it on L73. It's harder to reason about code that shares a var like this.

If you wanted to keep things mostly the way you have them I'd change `isHostFQDN` to be something like:

```go
func isHostFQDN(host string) bool {
   host = strings.TrimPrefix(host, ".")
   return util.IsFQDN(host)
}
```

Alternatively, what I was originally thinking (sorry if I didn't make this very clear) was you could write a function that takes a `[]string` as argument and returns a `[]string` of all of the entries that are **not** an FQDN. Then the caller only has to check if the len of the returned `[]string` is non-zero and if it is, process into a failure result & message.

v3/lints/rfc/lint_name_constraint_not_fqdn_test.go
```diff
@@ -0,0 +1,65 @@
...
+		},
+		{
+			Name:           "TestMultipleWrongContstraintsNotFQDN",
+			Filename:       "multipleWrongConstraints.pem",
+			ExpectedResult: lint.Error,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
```
cpu, line 58: Thanks for making this table-driven. Looks great :tada:

v3/lints/rfc/lint_name_constraint_not_fqdn_test.go
```diff
@@ -0,0 +1,65 @@
...
+			result := test.TestLint("e_name_constraint_not_fqdn", tc.Filename)
+			if result.Status != tc.ExpectedResult {
```
cpu, line 60: It would be nice to also get test coverage of the logic you wrote in `returnErrorString`. What do you think about extending the `testCases` to also include the expected details message and verifying that as well as the `result.Status`. I think that would help us get confidence that the string manipulation in `returnErrorString` is correct.

v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,118 @@
...
+
+	host = strings.TrimPrefix(host, ".")
+
+	if !util.IsFQDN(host) {
+		*incorrectHosts = append(*incorrectHosts, host)
+	}
+
+}
+
+func returnErrorString(incorrectHosts []string, isInclusion bool) string {
```
cpu, line 78: bikeshed nit: I would probably name this function something like `buildErrorString`, I think the "return" part is implicit in the function definition saying it returns a string. Just a personal preference!

cpu (changes requested):
Thanks @ZsofiaTomicsko, I appreciate your attention to the review feedback. I think we're moving in the right direction but I had a few remaining changes to request.

v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (l *nameConstraintNotFQDN) CheckApplies(c *x509.Certificate) bool {
+	return util.IsExtInCert(c, util.NameConstOID)
+}
+
+func (l *nameConstraintNotFQDN) Execute(c *x509.Certificate) *lint.LintResult {
+
+	incorrectPermittedHosts := make([]string, 0)
+	incorrectExcludedHosts := make([]string, 0)
+	errString := ""
```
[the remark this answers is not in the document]
ZsofiaTomicsko: Oh alright! This was a misunderstanding from my part. I will correct it in a sec

v3/lints/rfc/lint_name_constraint_not_fqdn_test.go
ZsofiaTomicsko: Okay! I will include a test for every case then. I actually already tested most of them I'm not sure why I decided against including those in my last commit. :'D

v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,120 @@
...
+
+		host = strings.TrimPrefix(host, ".")
+
+		if !util.IsFQDN(host) {
+			incorrectHosts = append(incorrectHosts, host)
+		}
+	}
+
+	return incorrectHosts
+}
```
ZsofiaTomicsko, line 78: I'm sorry, but I'm still not 100% sure if i fully understood your original idea. This function does more or less what you described in your previous comment, but instead of the []string it takes a []x509.GeneralSubtreeString. I could write another function that collects the data from this list and return a list of strings too, but I didn't think it was necessary. Does this look good to you now?

v3/lints/rfc/lint_name_constraint_not_fqdn_test.go
ZsofiaTomicsko: I have already had test files for:
1. 1 inclusion constraint
2.  multiple inclusion constraints
I added the following cases yesterday:
1. 1 exclusion constraint
2. 1 inclusion constraint and multiple exclusion constraints
3.  more than 1 constraint of both types
4. 1 constraint of both types

And now I'm realizing that I misread your comment yesterday. :'D I will include the error messages in my next commit. Sorry for this.

v3/lints/rfc/lint_name_constraint_not_fqdn.go
```diff
@@ -0,0 +1,120 @@
...
+
+func buildErrorString(incorrectHosts []string, isInclusion bool) string {
```
christopher-henderson, line 80: Grammatically correct string building is not easy :joy: but it is very much appreciated by users - thank you!

christopher-henderson (approved):

cpu:
I will give this another :mag: over the next few days. Apologies - it's a busy week.

v3/lints/rfc/lint_name_constraint_not_fqdn_test.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+func TestUriNameConstraintsFqdn(t *testing.T) {
+	testCases := []struct {
+		Name           string
+		Filename       string
+		ExpectedResult lint.LintStatus
+	}{
+		{
+			Name:           "TestBeginsWithPeridFQDN",
```
kruton, line 31: s/Perid/Period

v3/lints/rfc/lint_name_constraint_not_fqdn_test.go
cpu: Hi @ZsofiaTomicsko,

Did you end up pushing another commit for this? I see that L83 is logging the details message but I don't see the expected details message reflected in the `testCases`, or an assertion that the details message in `result` matches the expected one from the test case

cpu (approved):
Thanks for the iteration (and patience) @ZsofiaTomicsko, this looks great.

There's still one comment from my last review related to asserting on the expected vs actual details message in the error result test cases if you're interested in a small follow-up PR but I'm happy with the state of things as-is. I spot checked the logged details messages from `go test -v --test.run TestUriNameConstraintsFqdn  ./lints/rfc/...` and they looked good to me :+1:

PR #534, opened 2021-01-07
created lint for checking common name length limit

titmuscody:
lint to check if the common name is too long


v2/lints/rfc/lint_common_name_length.go
```diff
@@ -0,0 +1,81 @@
...
+	if c.Subject.CommonName != "" {
+		return true
+	}
+	return false
+}
+
+func (l *commonNameLengthLimit) Execute(c *x509.Certificate) *lint.LintResult {
+	// Add actual lint here
+	if c.Issuer.CommonName != "" {
+		if len(c.Issuer.CommonName) > 64 {
```
cpu: I think this implementation has a subtle problem. Remember that Go strings are sequences of UTF-8 characters and that isn't a 1:1 mapping with bytes. Your lint is checking that there are no more than 64 characters, but not that the encoding of the string is less than 64 bytes.

zakird:
Hi @titmus, thanks for the contribution. I believe that this lint is already covered here: https://github.com/zmap/zlint/blob/master/v3/lints/rfc/lint_subject_common_name_max_length.go.

cpu (changes requested):
Hi @titmuscody, thanks for the PR.

This looks like a duplicate of the [`e_subject_common_name_max_length` lint](https://github.com/zmap/zlint/blob/master/v3/lints/rfc/lint_subject_common_name_max_length.go).

If you're thinking of contributing other lints in the future we would also love to see [unit tests](https://github.com/zmap/zlint/blob/master/CONTRIBUTING.md#testing-lints) included. It would be a tough sell to merge a new lint without that coverage.

Thanks!

v2/lints/rfc/lint_common_name_length.go
cpu: For contrast the existing `e_subject_common_name_max_length` lint uses [`utf8.RuneCountInString`](https://golang.org/pkg/unicode/utf8/#RuneCountInString):

>	if utf8.RuneCountInString(c.Subject.CommonName) > 64 { ....

cpu:
I think we'll decline this PR for now. Let us know if we missed something!

PR #547, opened 2021-01-20
Follow-up PR for lint_name_constraint_not_fqdn_test

ZsofiaTomicsko:
Sorry for the delay, but I added the ErrorDetails to the testCases as you have requested @cpu . This is a follow-up PR to #533 .


cpu (approved):
Thanks @ZsofiaTomicsko - I appreciate you following up with this last bit :-)

Issue #549, opened 2021-01-24
Update e_ext_duplicate_extension error result with Details message.


cpu:
The [`e_ext_duplicate_extension`](https://github.com/zmap/zlint/blob/6d643b9bddd500d5585876859efa09403a1f0f42/v3/lints/rfc/lint_ext_duplicate_extension.go) lint presently returns a plain `&lint.LintResult{Status: lint.Error}` when it encounters a single duplicated extension.

It would be a nicer user experience if we returned a `LintResult` with a `Details` message that indicated what duplicate extension OIDs were found. Relates to https://github.com/zmap/zcertificate/issues/23.

cpu:
Fixed in master. We'll pull this into the next tagged release.

PR #550, opened 2021-01-24
lints: return detail for e_ext_duplicate_extension.

cpu:
Previously the `e_ext_duplicate_extension` lint from the `lint.RFC5280` source only returned a `lint.Error` result as soon as one duplicate extension was found in a certificate. It did not indicate which extension OID was duplicated, or if there was more than one duplicated extensions.

This PR reworks the lint to do both of these things. The detail string now indicates all of the extension OIDs that were present more than once.

Resolves https://github.com/zmap/zlint/issues/549

### DIFF
--- a/v3/testdata/multDupeExts.pem
+++ b/v3/testdata/multDupeExts.pem
@@ -1,0 +1,133 @@
+Certificate:
+    Data:
+        Version: 3 (0x2)
+        Serial Number: 2 (0x2)
+        Signature Algorithm: sha256WithRSAEncryption
+        Issuer: C = UN, ST = NYS, O = UNGA, OU = UNSC, CN = DT
+        Validity
+            Not Before: Jun 25 19:55:19 2019 GMT
+            Not After : Jan 21 03:29:38 2030 GMT
+        Subject: C = UN, ST = NYS, O = UNGA, OU = UNSC-peace, CN = DT-peace
+        Subject Public Key Info:
+            Public Key Algorithm: rsaEncryption
+                RSA Public-Key: (4096 bit)
+                Modulus:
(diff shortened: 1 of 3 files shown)


v3/testdata/multDupeExts.pem
```diff
@@ -0,0 +1,133 @@
...
+                keyid:EA:D3:9D:F2:FA:12:15:1D:6B:90:01:1F:1D:DB:27:7F:AF:D1:65:D7
+
+            X509v3 Subject Key Identifier: 
+                BC:93:A7:C1:4D:51:A1:B1:1E:5D:C9:C1:91:EA:DB:5B:53:D5:BC:58
+            X509v3 Key Usage: critical
+                Certificate Sign
+            X509v3 Basic Constraints: critical
+                CA:FALSE
+            X509v3 Subject Key Identifier: 
+                BC:93:A7:C1:4D:51:A1:B1:1E:5D:C9:C1:91:EA:DB:5B:53:D5:BC:58
```
cpu, line 62: I took the certificate provided in https://github.com/zmap/zcertificate/issues/23 which already had a duplicated x509v3 Authority Key Identifier extension and then used [der-ascii](https://github.com/google/der-ascii/) to additionally duplicate the x509v3 Subject Key Identifier extension to make this test file.

zakird (approved):

zakird:
This looks good to me. Thanks for the clearer message. Small nit, we may want to match the error text to be closer to the name of the lint, e.g., something like "Found extension OIDs present more than one time" -> "The following extensions are duplicated".

cpu:
>  Small nit, we may want to match the error text to be closer to the name of the lint

@zakird Good idea, fixed in f4223a9

PR #554, opened 2021-01-28
lints: fix description of e_ext_ian_uri_not_ia5

mathewhodson:


cpu (approved):
Good catch, thanks @mathewhodson

PR #556, opened 2021-01-29
lints: revert e_key_usage_and_extended_key_usage_inconsistent.

cpu:
The `e_key_usage_and_extended_key_usage_inconsistent` lint's interpretation of RFC 5280 is under question (see zmap#553).

We also had an integration test bug that resulted in massively under-estimating its impact on our integration test corpus (see zmap#555).

Let's remove this lint while we sort out the correct logic.


christopher-henderson:
I was going to suggest exactly this this evening, thank you for jumping on it!

(I've been sharing these threads with friends and colleagues with the spicy title `Yeah, but what does "or" mean?` and it's turning into "gold-blue dress" like debacle)

christopher-henderson (approved):

cpu:
Thanks for the quick :mag:

cpu:
Thanks for the quick :mag:

Issue #555, opened 2021-01-28
Integration test resultCounts overflow if >255 certs lint to the same status from the same lint


cpu:
#### Background

While investigating https://github.com/zmap/zlint/issues/553 I noticed that there was an inconsistency between the expected error results by lint name & the number of certificate fingerprints that failed linting:

> Running:
>
> make integration SHELL=(which bash) INT_FLAGS="-lintSummary -fingerprintSummary -lintFilter='e_key_usage_and_extended_key_usage_inconsistent'"
> 
>I get the summary:
>
> summary of result type by lint name:
>   e_key_usage_and_extended_key_usage_inconsistent                  	fatals: 0    errs: 140  warns: 0    infos: 0
>
> But there are ~566,923 certificate fingerprints with "errs: 1" in the output from -fingerprintSummary. 

#### Root cause
The root cause is that the `integration.resultCount` type used to collate findings uses a collection of `uint8` counters:
https://github.com/zmap/zlint/blob/c1c6681339ef2247dca5ff81162ab80c4811b154/v3/integration/result.go#L11-L16

This **overflows** when more than 255 certificates return a result of a given level. :facepalm:  :facepalm:  :facepalm: 

I can't remember how or why I made the decision to use `uint8` for these fields. It seems very very silly in retrospect!

cpu:
> I can't remember how or why I made the decision to use uint8 for these fields. It seems very very silly in retrospect!

I think I remember the rough shape of the "why" now: initially I implemented the integration test ground truth per-test certificate (e.g. ~600,000 rows of JSON, each with an err, info, etc count). That made for a big JSON file (~100mb) and I was noodling with using a smaller type since we didn't have 255 lints. Later we restructured the tests to group source of truth by lint instead of by cert but didn't update this type. Oh well :cry:

PR #557, opened 2021-01-29
integration: fix resultCount types to not overflow, update expected vals.

cpu:
#### integration: fix resultCount types to not overflow

Using `uint8` as the type for the count fields in the `resultCount` type produces overflows if more than 255 certificates with a given result level are linted.

Our integration test corpus is just shy of 600,000 certificates so `uint32` should be more than sufficient.

####  integration: update expected values to correct overflows.

Any lints that had more than 255 results at a given level will have overflowed, meaning expected counts were not correct. Also updates the expected values for the small config (which had drifted out of sync).

Resolves https://github.com/zmap/zlint/issues/555


v3/integration/result.go
```diff
@@ -9,10 +9,10 @@ import (
 )
 
 type resultCount struct {
-	FatalCount  uint8 `json:",omitempty"`
-	ErrCount    uint8 `json:",omitempty"`
-	WarnCount   uint8 `json:",omitempty"`
-	NoticeCount uint8 `json:",omitempty"`
+	FatalCount  uint32 `json:",omitempty"`
```
christopher-henderson, line 12: image

I'm gonna pull this branch and give it a quick smoke check myself as well as double check the sanity if the new values.

cpu: Thank you! Take your time :pray:

christopher-henderson (approved):
A quick spot check that the new values make sense in that they would overflow to the old values.

```go
package main

func main() {
	var overflows = [][2]uint32{
		{322, 66},
		{431, 175},
		{738, 226},
		{3463, 135},
		{1677, 141},
		{607, 95},
		{607, 95},
		{365, 109},
		{325, 69},
		{724, 212},
		{459, 203},
		{705, 193},
		{678, 166},
		{593196, 44},
		{4680, 72},
		{339, 83},
		{284, 28},
		{7191, 23},
		{9872, 144},
		{16755, 115},
		{59194, 58},
		{816, 48},
		{574, 62},
		{33636, 100},
		{2025, 233},
		{19776, 64},
		{329, 73},
		{552, 40},
		{1944, 152},
		{1122, 98},
	}
	for _, pair := range overflows {
		is := pair[0]
		was := pair[1]
		if uint32(uint8(is)) != was {
			panic("💣")
		}
	}
}
```

v3/integration/config.json
```diff
@@ -638,47 +647,47 @@
...
     "n_sub_ca_eku_missing": {
-      "NoticeCount": 166
+      "NoticeCount": 678
     },
     "n_sub_ca_eku_not_technically_constrained": {
       "NoticeCount": 10
     },
     "n_subject_common_name_included": {
-      "NoticeCount": 44
+      "NoticeCount": 593196
```
christopher-henderson, line 662: This...is especially terrifying.

zakird:
Well, everything about this ticket makes me sad, but I'm glad we're getting it cleaned up.

zakird (approved):

v3/integration/config.json
cpu: but expected! That's why it's a Notice level. The deprecation of subject CNs is... long and ongoing. Honestly seeing only 44 notices in our corpus before for this lint should have been a red flag :sweat:

PR #560, opened 2021-01-31
Extend documentation on how to generate a certificate in go

RufusJWB:
Since I struggled to find documentation on how to generate a certificate in `go` I added an example to save other potential contributors from googling around.


CONTRIBUTING.md
```diff
@@ -112,13 +112,15 @@ Testing Lints
 **Creating Unit Tests.** Every lint should also have corresponding unit
 tests (generally at least one for a success and one for a failure condition). We
 have typically generated test certificates using Go (see
-[documentation][CreateCertificates] for details), but OpenSSL
-could also be used. Test certificates should be placed in `testdata/` and called
-from the test file created by `newLint.sh`. You may want to prepend the PEM with
-the output of `openssl x509 -text`. You can run your lint against a test
-certificate from a unit test using the `test.TestLint` helper function.
+[documentation][CreateCertificates] for details and [example][certGenerator] as
```
cpu, line 115: couple nit:

* it's `CreateCertificate` singular,
* also I think sensible to reference the namespace, `x509.CreateCertificate`
* lastly, a grammatical tweak (and example as reference -> and this example code as reference).


```suggestion
[documentation][x509.CreateCertificate] for details and [this example code][certGenerator] as
```

CONTRIBUTING.md
```diff
@@ -112,13 +112,15 @@ Testing Lints
...
+reference), but OpenSSL could also be used. Test certificates should be placed in
+`testdata/` and called from the test file created by `newLint.sh`. You may want to
+prepend the PEM with the output of `openssl x509 -text`. You can run your lint
+against a test certificate from a unit test using the `test.TestLint` helper
```
cpu, line 119: @RufusJWB While you're editing this part of the docs would you mind updating this part:

> You may want to the output of `openssl x509 -text`. You can run your lint against a test	prepend the PEM with the output of `openssl x509 -text`. 

This is actually mandatory now (CI will fail otherwise) and the best way to do it is by running `test/prepend_testcerts_openssl.sh`. Maybe we can say something like:

> All test certificates **must** have the textual description from `openssl x509 -text` added before the PEM header or CI will flag them as a build error. You can add the text decoding to all of the test certs missing it by running `test/prepend_testcerts_openssl.sh`.

CONTRIBUTING.md
```diff
@@ -112,13 +112,15 @@ Testing Lints
...
+function.
 
 [CreateCertificates]: https://golang.org/pkg/crypto/x509/#CreateCertificate
+[certGenerator]: ./certGenerator.go
```
cpu, line 123: I'd prefer this doesn't live in the root of the repository. It should probably be under `v3/test`.

certGenerator.go
```diff
@@ -0,0 +1,46 @@
+// based on https://stackoverflow.com/q/26441547/1426535
```
cpu, line 1: This file needs the ZLint licensing information

certGenerator.go
```diff
@@ -0,0 +1,46 @@
...
+
```
cpu, line 2: This might be overkill, but I'd like to see this gated behind a [build contraint tag](https://golang.org/cmd/go/#hdr-Build_constraints), similar to the integration tests.

I don't think we want to distribute this CLI program (e.g. in a ZLint release).

certGenerator.go
```diff
@@ -0,0 +1,46 @@
...
+	privatekey, err := rsa.GenerateKey(rand.Reader, 2048)
+	if err != nil {
+		panic(err)
+	}
+
+	crt, err := x509.CreateCertificate(rand.Reader, &template, &template, &privatekey.PublicKey, privatekey)
+	if err != nil {
+		panic(err)
+	}
```
cpu, line 39: I have to say, I'm a little bit torn about whether this is really a useful piece of ZLint documentation.

As one example, in my experience writing other small programs to generate ZLint test certificates (e.g.[sctTestCerts.go](https://gist.github.com/cpu/6d26b2718f29e184ff88a90f02d7cbcb/raw/2a5f8a0c673f197de83d176667c76f5f8e498147/sctTestCerts.go)) I've needed to do more contortions to make sure the test certs will be linted. Anything following this example closely will never pass the (very commonly used) [`util.IsSubscriberCert`](https://github.com/zmap/zlint/blob/848c50b07013a773f2d98620400a6be900c88d2c/v3/util/ca.go#L41-L45) check because the example is a self-signed certificate. You could fix the example to also generate an issuer certificate but then it will be even more code in-tree.

certGenerator.go
```diff
@@ -0,0 +1,46 @@
...
+
+	var certOut, keyOut bytes.Buffer
+	pem.Encode(&certOut, &pem.Block{Type: "CERTIFICATE", Bytes: crt})
+	pem.Encode(&keyOut, &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(privatekey)})
+	ioutil.WriteFile("c:\\temp\\private_"+filename+".key", keyOut.Bytes(), 0644)
+	ioutil.WriteFile("c:\\temp\\certificate_"+filename+".pem", certOut.Bytes(), 0644)
```
cpu, line 45: I think these should be `flag.String` arguments to the program instead of hardcoded paths. I'd also prefer to see the defaults assume a UNIX system since the rest of ZLint's codebase is fairly UNIX-centric.

It should also handle the `error` return from `ioutil.WriteFile` or make explicit that you aren't (e.g. with `_`).

certGenerator.go
```diff
@@ -0,0 +1,46 @@
...
+
+	crt, err := x509.CreateCertificate(rand.Reader, &template, &template, &privatekey.PublicKey, privatekey)
+	if err != nil {
+		panic(err)
+	}
+
+	var certOut, keyOut bytes.Buffer
+	pem.Encode(&certOut, &pem.Block{Type: "CERTIFICATE", Bytes: crt})
+	pem.Encode(&keyOut, &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(privatekey)})
+	ioutil.WriteFile("c:\\temp\\private_"+filename+".key", keyOut.Bytes(), 0644)
```
cpu, line 44: My instincts are to twitch when I see a private key being written with mode `0644`, even if it's a test certificate! A stricter mode would make me feel better....

Also is there a reason to output the key at all? Only the cert is checked into the repo. I typically throw away the private keys.

cpu (changes requested):
Thanks for the PR @RufusJWB 

I've left a handful of pieces of feedback but overall I'm a little bit torn on whether this is a good fit for the repo or not.

In general I feel like it's reasonable to expect ZLint contributors to be able to navigate the `x509` docs to generate a certificate. I admit that I'm likely making things harder for folks new to Go and x509 but I also have to admit that our reviewing bandwidth is limited and we can't necessarily prioritize helping folks learn standard libraries.

Would this be better as an external resource we could link to? A blog post or a gist or something of that nature? I think there's a need for something with more guidance than "Here's the x509 package docs, good luck!" but a well written external resource could fill that void without needing to be something we maintain in-tree.

What do you think?

CONTRIBUTING.md
cpu: Also I believe where you've placed it now it's not being built in CI or linted with golangci-lint. We'd definitely want the latter at least, and likely should consider the former to prevent bitrot over time.

certGenerator.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	crt, err := x509.CreateCertificate(rand.Reader, &template, &template, &privatekey.PublicKey, privatekey)
+	if err != nil {
+		panic(err)
+	}
+
+	var certOut, keyOut bytes.Buffer
+	pem.Encode(&certOut, &pem.Block{Type: "CERTIFICATE", Bytes: crt})
+	pem.Encode(&keyOut, &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(privatekey)})
```
cpu, line 43: I think you're also missing a potential `error` return here too (the linter would catch this if you move it into the `v3` module)

cardonator:
I agree that this probably isn't right to include in this repo. Is there an existing open source project that could be linked to in order to accomplish a same end?

RufusJWB:
Based on the feedback here, which I appreciate, I would update my code sample, add it to a GIST and we could link to that GIST? Is this okay for everybody?

cpu:
@RufusJWB That sounds good to me.

christopher-henderson:
> I agree that this probably isn't right to include in this repo. Is there an existing open source project that could be linked to in order to accomplish a same end?

I've been directed to [minica](https://github.com/jsha/minica) before for generating certificates from a Go program. It's from Josh over at Let's Encrypt so even if it's just a toy it's likely in the ballpark at least. There is also a pretty long list of [repos](https://github.com/topics/certificate-authority) (many likely hobbyist) that are likely mostly frontends to OpenSSL.

Perhaps we could advise to generate a certificate via something like minica, deserialize it, whack it around in the exact right way that you want, and reserialize it for submissions as test data?

```go
func main() {
    src := os.Args[1]
    dst := os.Args[2]
    cert := parse(src)
    whack(cert)
    serialize(dst, cert)
}

func whack(cert *x509.Certificate) error {
	// Do whatever you need to do to the cert.
	cert.Issuer.CommonName = "Honest Achmed's Used Cars and Certificates"
	return nil
}
```

The difficulty I hit with the above is actually serializing the certificate again. A `*x509.Certificate` is really just a bunch of fields and when you mutate `cert.Issuer.CommonName` that does not reflect in `cert.Raw`. So then you have to really get this struct back into asn1. However, when you feed the cert to `asn1.Marshal` you quickly realize that `x509.Certificate` doesn't have any of the asn1 tagging necessary for serialization - all of those are in a private type named `certificate`.

sleevi:
At the risk of piling on, I use https://github.com/google/der-ascii/blob/master/samples/certificates.md for anything related to linting/testing, precisely because it lets me create as many ASN.1 mutants as I can possibly imagine :)

cpu:
>  It's from Josh over at Let's Encrypt so even if it's just a toy it's likely in the ballpark at least

_I think you mean Jsha, not Josh Aas :-)_

+1 for linking to miniCA as a reference if one is required.

> At the risk of piling on, I use https://github.com/google/der-ascii/blob/master/samples/certificates.md for anything related to linting/testing, precisely because it lets me create as many ASN.1 mutants as I can possibly imagine :)

Yes! der-ascii rules.

RufusJWB:
> > At the risk of piling on, I use https://github.com/google/der-ascii/blob/master/samples/certificates.md for anything related to linting/testing, precisely because it lets me create as many ASN.1 mutants as I can possibly imagine :)
> 
> Yes! der-ascii rules.

I was playing around with `der-ascii` when I started with my first lint here. I must have missed the explanation, how to sign a modified certificate as described in https://github.com/google/der-ascii/blob/master/samples/certificates.md . That's why I skipped using it.

cpu:
@RufusJWB is this branch ready for a second look? Thanks

RufusJWB:
> @RufusJWB is this branch ready for a second look? Thanks

@cpu I did three things:

1. Removed the `.go`-file
2. Updated the paragraph about the prepended plain text
3. Added different options how to generate test certificate

--> I think this PR is ready for a second look.

christopher-henderson (changes requested):
Just some points on conjugations.

The larger topic of users then in turn being confused by the standard library I think is out of scope of this change.

CONTRIBUTING.md
RufusJWB: done

christopher-henderson (approved):

cpu (approved):
Thanks!

Issue #564, opened 2021-02-13
Lint template produces a file with an `init` function that is not at the top of the new lint


christopher-henderson:

cpu:
Do as I say, not as I generate

christopher-henderson:
@cpu it reminds me of one time when I made a small contribution to a project. The entire project was not following the "standard" style guidelines, but it was at least very consistent in the style that it had chosen. So, since I was in someone else's home, my small change emulated their style to the T.

So of course the very author of those many thousands of lines of customly stylized code chastised me for not following the official style guide.

![image](https://user-images.githubusercontent.com/8518514/107861897-c4275e80-6dfd-11eb-9ac4-860ea0ab3f42.png)

robplee:
One other thing about the lint template is that it currently has the import wrong - "github.com/zmap/zlint/v2/lint" should be v3 these days :)

christopher-henderson:
@robplee precisely, and we did catch and bundle that in the change [on the sly](https://github.com/zmap/zlint/commit/1f157ab6c8ab2a591daefeaad74bd061b13b4d3f).

But I certainly did forget to close this issue, so thanks for the reminder. :joy:

PR #565, opened 2021-02-13
Lint template produces a file with an `init` function that is not at the top of the new lint

christopher-henderson:
This address issues #564 

(also caught the `v2` import, but I don't reckon that's worth another PR)


cpu (approved):
:sweat_smile: Nice catch.

Issue #568, opened 2021-02-14
Include a playground script for generating one off certificates and certificate chains


christopher-henderson:
It was brought up by @RufusJWB that generating test data certificates can be opaque and slow for contributors. I actually found myself writing a small Go script the other day in order to rapid fire generate (relatively speaking) test data. I'll open up a PR to see how palatable it is to have a playground in the repo.

PR #569, opened 2021-02-14
Include a playground script for generating one off certificates and certificate chains

christopher-henderson:
@RufusJWB I think this might go some ways towards what you were looking for with regard to getting contributors an easier way to generate test data.

The intent of this script is that it behave more like a Python script than a Go CLI program. That is, I don't think that we should be attempting some sort of generic CLI tool or service here; that already exists in OpenSSL and our version would only be even more confusing. Rather, I'm thinking that this can just be something that contributors can edit, `go run`, and get their strangely specific certs generated as one-offs (at least, this is how I was using it).

That is:

> Here is a working sample of the Go standard library. Here is the x509.Certificate type with all of its fields laid bare to you. Here are some functions for building a signed certificate chain. And here is a function that gets you the pretty string output that you can just copy into `testdata/myTestCert.pem`.

The idea is that contributors can hop to the fully expanding constructor list for each x509.Certificate and immediately see what is available to them. The minutiae of how to get a properly signed chained is largely handled for them, and even if they have to do something different I believe this serves as a workable sample that someone new to the API can easily figure out and bend to their will (after all, just seeing a working sample goes a long ways).

Once we merge the custom code linter then I think it would be a good idea to write a lint that hashes this script and makes sure that people don't accidentally commit their local changes.

This addresses issue #568


sleevi (commented):
I'm probably a bad one to review :)

I've found most playgrounds inevitably end up being more hassle than they're worth (I suppose I'm old and hate learning new things now... also, please watch the grass while you move to the sidewalk). That said, I'm also poorly calibrated for what is good to help someone bootstrap into this space, since I'm... off in the deep-end of weird things :) So I'm not opposed to this, but also don't feel qualified to offer much about how users could/should/will use this :)

christopher-henderson:
@sleevi however your wisdom is always appreciated :)

I suppose that this would be more of a user acceptance test from contributors outside of the zmap org, especially those who do not typically write Go. Unfortunately those are exactly the people who cannot leave green/red checkmarks on a review, so I kinda grasped at straws for names on this one.

zakird (commented):
I think this is fine to merge. I have a somewhat similar view as Sleevi, but there's really no harm and it might indeed be useful to someone, so I don't see much reason not to. I'll leave final decision to @cpu since I think he might have the the strongest opinions. My one comment would be maybe to move that content out of README to elsewhere.

v3/integration/README.md
```diff
@@ -332,3 +332,32 @@ Once you're confident the observed results match expectations you can add the
 new lint results to the expected data by running the integration tests with
 `-overwriteExpected` and committing the updated config file along with your new
 lint. Nice work!
+
+Generating Test Certificates
+-----------------
```
cpu, line 337: This feels like something better suited to CONTRIBUTING.md than v3/integration/README.md.

It feels confusing to have this here because the integration test data corpus is entirely a static set of real world certificates from Censys, no special snowflakes.

v3/integration/README.md
```diff
@@ -332,3 +332,32 @@ Once you're confident the observed results match expectations you can add the
...
+your goals. If you think that you have improved upon the contents of the script itself, then please do
+open a pull against the script itself (however, please refrain from bundling it with anything else such as 
+a new lint).
+
+This script has a facility for generating a self signed trust anchor to act as a CA, a facility
+for generating intermediate certificates, and a facility for generating a leaf certificate.
+The certificates generated by each are NOT healthy nor acceptable to any reasonable PKI system.
+However, being a complete and usable certificate is not necessarily required when you are writing
+a lint for, say, checking that a certificate does not expire on Valentine's Day (because no certificate
+should be alone on Valentine's Day).
```
cpu, line 359: :revolving_hearts: :lock:

v3/integration/genTestCerts/genTestCerts.go
```diff
@@ -0,0 +1,314 @@
...
+	"time"
+)
+
+// Generates a CA, an intermediate, and a leaf certificate and prints their
+// OpenSSL textual output to stdout.
+func main() {
+	ca, err := NewTrustAnchor()
+	if err != nil {
+		panic(err)
+	}
```
cpu, line 24: small nit: it might be useful to define a small function that panics on a non-nil err and then use it throughout to reduce the duplication in `main` somewhat. 

```go
panicOnErr := func(err error) {
  if err != nil {
    panic(err)
  }
}
ca, err := NewTrustAnchor()
panicOnErr(err)
fmted, err := OpenSSLFormatCertificate(ca)
panicOnErr(err)
...
```

v3/integration/genTestCerts/genTestCerts.go
```diff
@@ -0,0 +1,314 @@
...
+	fmted, err := OpenSSLFormatCertificate(ca)
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println("-------------Trust Anchor-------------")
```
cpu, line 29: Another ultra tiny nit: I would be tempted to make a `headerDelim :=  "-------------"` var and then use a style like `fmt.Printf("%sTrust Anchor%s", headerDelim, headerDelim)` throughout. Mostly so we don't ever have to worry about mismatching the number of `-'s` in a header.

v3/integration/genTestCerts/genTestCerts.go
```diff
@@ -0,0 +1,314 @@
...
+		return nil, err
+	}
+	return x509.ParseCertificate(cert)
+}
+
+// This is NOT a healthy example of a leaf certificate, this is nothing
+// more than a self signed certificate with IsCA set to false. Not even any
+// basic constraints are defined. Please do not think that this will be
+// acceptable to any system, let alone lint particularly well.
+func NewLeaf(trustAnchor *x509.Certificate, intermediates []*x509.Certificate) (*x509.Certificate, error) {
```
cpu, line 183: It seems like in all cases this function only looks at the first entry in `intermediates`. Would it be clearer to make this argument `intermediate *x509.Certificate` instead of `intermediates []*x509.Certificate`?

v3/integration/genTestCerts/genTestCerts.go
```diff
@@ -0,0 +1,314 @@
...
+	var parent *x509.Certificate
+	if len(intermediates) == 0 {
```
cpu, line 185: if you take my suggestion from above this will have to change to `if intermediate == nil`

v3/integration/genTestCerts/genTestCerts.go
```diff
@@ -0,0 +1,314 @@
...
+		parent = trustAnchor
+	} else {
+		parent = intermediates[len(intermediates)-1]
+	}
+	// Edit this template to look like whatever lead cert you need.
```
cpu, line 190: :warning: Lead certificates are known to the Government Of California to be a carcinogen :warning: 

```suggestion
	// Edit this template to look like whatever leaf cert you need.
```

v3/integration/genTestCerts/genTestCerts.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	fmt.Println("-------------Leaf-------------")
+	fmt.Println(fmted)
+}
+
+// This is NOT a healthy example of a CA certificate, this is nothing
+// more than a self signed certificate with IsCA set to true. Not even any
+// basic constraints are defined. Please do not think that this will be
+// acceptable to any system, let alone lint particularly well.
+func NewTrustAnchor() (*x509.Certificate, error) {
```
cpu, line 57: Across the board I'd encourage not exporting any funcs since they're part of the internal API for the script.

v3/integration/genTestCerts/genTestCerts.go
```diff
@@ -0,0 +1,314 @@
...
+	cmd.Stdin = strings.NewReader(string(block))
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", err
+	}
+	return string(output), nil
+}
+
+// Just a simple, sequential, serial number generator.
+var NextSerial = func() func() *big.Int {
```
cpu, line 305: This is neat but a random serial number probably would have worked OK too :laughing: If the target audience of this script is folks not very familiar with Go a generator function that uses a sync primitive is probably going to seem somewhat magical. It's more robust, but this is a test script and so I think simplicity might be a better tradeoff.

cpu:
> I think this is fine to merge. I have a somewhat similar view as Sleevi, but there's really no harm and it might indeed be useful to someone, so I don't see much reason not to. I'll leave final decision to @cpu since I think he might have the the strongest opinions.

I'm still mildly unconvinced that there's an audience that is ready to contribute to ZLint but not ready to figure out x509 certificate generation _but_ I also recognize that this could be a nice time-saver for folks and seems like a reasonable template to start from.

It might be nice if we could add a default subject prefix or something to the certificate template that might let us see down the road how many unit tests are added to the tree that use certs generated by this script. If it doesn't see any use by contributors after a year or two we could revisit.

v3/integration/README.md
cpu: Coming back around, ditto for the script itself: can it live somewhere other than `v3/integration`? It isn't used by integration tests or run in CI.

RufusJWB:
> > I think this is fine to merge. I have a somewhat similar view as Sleevi, but there's really no harm and it might indeed be useful to someone, so I don't see much reason not to. I'll leave final decision to @cpu since I think he might have the the strongest opinions.
> 
> I'm still mildly unconvinced that there's an audience that is ready to contribute to ZLint but not ready to figure out x509 certificate generation _but_ I also recognize that this could be a nice time-saver for folks and seems like a reasonable template to start from.

Well, when I started to write the Lint for https://github.com/zmap/zlint/issues/471 it took me a day or so to find out, how exactly to configure OpenSSL to create exactly the certificates I needed for that Lint. Of course, there are plenty of examples of how to use OpenSSL to generate certificates but to tweak these examples into the exact configuration needed was not really straightforward. Therefore I'm convinced there are people out there with sufficient know-how about PKI to help writing Lints but who never needed to use OpenSSL to create certificates.

cpu:
@RufusJWB Thanks for weighing in, I appreciate your input here as an external contributor. I'm convinced we should merge this PR (I'll defer doing that to @christopher-henderson so they can sort through my review feedback if interested).

RE your experience, I think my core takeaway is that we shouldn't be encouraging using OpenSSL, not that we need too much in the way of extra resources. IME OpenSSL is a particularly bad tool for most everything, certificate generation included. 

Either way I hope the Go based example will make for an easier experience in the future!

RufusJWB:
> RE your experience, I think my core takeaway is that we shouldn't be encouraging using OpenSSL, not that we need too much in the way of extra resources. IME OpenSSL is a particularly bad tool for most everything, certificate generation included.

OpenSSL seems to be PKI equivalent to editing a text in `vi`. If you are able to master it, you are the cool hero of the neighborhood but for everybody else, it is just scaring

v3/integration/genTestCerts/genTestCerts.go
christopher-henderson: Agreed, although this is a `package main` and I don't think others could import these names even they wanted to (I don't think you can import `example.com/project/main`).

But sure. Perhaps I just don't like camel case and that's why I do this for self contained scripts =p

cpu: > (I don't think you can import example.com/project/main).

Fair point!

v3/integration/genTestCerts/genTestCerts.go
christopher-henderson: Not quite. A certificate chain is a trust anchor, followed by any number of intermediates, followed by a leaf. What the body of the function is saying is:

```
If there are no intermediates:
    then the parent of the leaf is the trust anchor itself
else:
    then the parent of the leaf is the last intermediate in the chain
```

I thought of writing the function as:

```go
func NewLeaf(parent *x509.Certificate) (*x509.Certificate, error) {}
```

...but then I realized that then the caller would have to figure out who the appropriate parent is.

My goal was to have the inputs to this function be the outputs of the other two functions so that if you whacked around, say, `NewIntermediate`, to always return an empty slice then the downstream logic would pick up on that without any changes.

v3/integration/genTestCerts/genTestCerts.go
christopher-henderson: I think I'll have both - one sequential and one random (because lord knows at someone point we're going to need to lint something to do with serial numbers with various properties).

christopher-henderson:
@cpu mostly I moved the documentation to `CONTRIBUTING.md` and put the command itself under `v3/cmd/genTestCerts` because...well...it's command. It seems reasonable to me, at least.

Perhaps more interestingly I added a new code lint to `v3/integration/lints/lints/not_commiting_genTestCerts.go` which asserts that SHA256 of this file. That is, if we are going to encourage others and ourselves to edit this file to generate arbitrary test data, then it would be nice if we didn't have to hunt down accidental commits to this file on every PR.

v3/integration/lints/lints/not_commiting_genTestCerts.go
```diff
@@ -0,0 +1,58 @@
...
+	"go/ast"
+	"io/ioutil"
+	"strings"
+
+	"github.com/zmap/zlint/v3/integration/lints/lint"
+)
+
+const want = `a132b2f256dd0b68bff043568f0e9bae69089082805da64720168294f835e197`
+
+type NotCommitingGenTestCerts struct{}
```
robplee, line 29: Shouldn't it be "NotCommittingGenTestCerts" with the extra 't' in committing?

v3/integration/lints/main.go
```diff
@@ -26,6 +26,7 @@ import (
 
 var linters = []lint.Lint{
 	&lints.InitFirst{},
+	&lints.NotCommitingGenTestCerts{},
```
robplee, line 29: Same comment re: spelling as on the file defining lint in question.

robplee (commented):
Only a minor spelling issue from me, otherwise I think this is a great addition that very much should make the lives of anyone trying to add lints much easier!

v3/integration/lints/lints/not_commiting_genTestCerts.go
christopher-henderson: I am a native English speaker who finds great joy in our bizarre little Frankenstein of a language. But I swear that "committing" gets me every time.

christopher-henderson:
Gentle ping to @cpu and @robplee (I think the most pressing issue was location, for which I moved the help documentation to `CONTRIBUTING.md` and the actual command to `v3/cmd/genTestCerts/genTestCerts.go`)

cpu:
@christopher-henderson Thanks for the ping - will take a look shortly.

cpu (approved):
Thanks @christopher-henderson

robplee:
Hi @christopher-henderson, I am sorry to make such a small comment when you're so close to merging but this filename should also be sorted to contain the regulation number of 't's in 'committing' v3/integration/lints/lints/not_commiting_genTestCerts.go

christopher-henderson:
lol no worries @robplee cleanliness is next to godliness ;)

robplee (approved):
Looks good to me.

Issue #537, opened 2021-01-08
Lint that DSA is not used 


sleevi:
DSA was part of the BRs up to [version 1.7.1](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.1.pdf) (2020-08-20), where it was removed as part of the [browser alignment ballot](https://cabforum.org/2020/07/16/ballot-sc31-browser-alignment/).

The source of the "lack of DSA" goes back to [Mozilla Policy 2.0](https://wiki.mozilla.org/CA:CertMaintenancePolicyV2.0), which did not list DSA as acceptable/supported (effective 2011-02-02). However, the question of "absence or violation" was not resolved until [Policy 2.4.1](https://github.com/mozilla/pkipolicy/blob/2.4.1/rootstore/policy.md), effective date 2017-03-31, in which it was restructured to clarify that it's an exclusive set of algorithms.

Zlint has `cabf_br/lint_dh_params_missing.go`, to ensure DH params are present for DSA certificates, but that lint is obviated by BRs 1.7.1, and may itself be supplanted by an additional Mozilla-specific lint on DSA prohibition set for the 2.4.1 effective date.

sleevi:
Marking it as "question" so people can offer their interpretation of the effective dates here, should a lint be contributed.

sleevi:
This also requires updating `cabf_br/lint_public_key_type_not_allowed`

robplee:
Question: a number of lints are going to be deprecated/rendered worthless by a new "lint_no_more_dsa" check.  Should we also create an issue to remove them now or do we want to wait until this has been completed?  I guess the other question I have, as a new contributor, is to ask if removing said lints lints will necessitate a major version number increase?

Apologies if this is the wrong place to ask.

sleevi:
@robplee It’s a good question! I mentioned the effective dates, precisely because it is example of something that once was allowed, but no longer. In other cases, such as certificate lifetimes, lints have added new restrictions (e.g. 825 days to 398), but we’ve kept the old lint around to cover the period in which it was allowed. So introducing a “no DSA allowed” would have an effective date `Y`, as mentioned above, while the other DSA tests would cover `X...Y`.

That’s at least how things have been done, but as a result, it does mean things are additive over time, which might make things less sustainable, which is also why I tagged it discussion.

Removing lints would generally be a new major version, yes.

zakird:
[...]

PR #572, opened 2021-03-01
Add a new lint to prohibit using DSA

robplee:
This is my attempt at addressing #537 although I expect there may be a number of changes that will be needed as this is my first attempt at adding a lint to zlint.  I created a new certificate to cover an error case but I reused some of the certs already in testdata to cover Pass and Not Effective cases.

Please let me know if and when anything needs changing :)


cpu (approved):
Thanks @robplee, really nice work! 

I'll hold on merging this for a little while to give others a chance to review but it looks good to me.

cpu:
> @cpu cpu closed this now
> @cpu cpu reopened this now

CI statuses looked jammed up. Gave it the ol' close/open to get things flowing again.

christopher-henderson (approved):
Looks great to me, thank you @robplee !

zakird (approved):

sleevi:
Was the intent to only add this as a Mozilla test, or to also reflect that the BRs also now forbid this? I probably worded the issue wrong, but was trying to capture from (moz 2.4.1 effective date to BRs 1.7.1 effective date), Moz forbade it, and then onward, the BRs forbade it.

I suspect that means two separate lints (this was the question part of #537 ), and I’m fine with only tackling one of those in this :)

zakird:
Two lints :)

sleevi:
Yeah, sorry for confusing things yet again :) We resolved on the issue two lints, and so I just wanted to make sure this wasn’t seen as closing out the issue, since this is only one lint. But I’m definitely good with landing this lint as is, or also duplicating it (and changing the effective date) as a BR lint, which would then close #537

christopher-henderson:
Howdy @sleevi. I'm gonna merge this and I went ahead and fired up #574 to track the similar work for the BR effective date.

PR #576, opened 2021-03-07
Introduce an upper bounds to effective dates 

### LINKED ISSUE #575: Introduce an upper bounds to effective dates
In the issue for [`Lint that DSA is not used `](https://github.com/zmap/zlint/issues/537#issuecomment-757876869) we opened up the idea of a policy that may eventually be deprecated, and the notion of a sort of `notAfter` for lints was introduced.

One would think this change would be as simple as

```go
// Lints automatically returns NE for all certificates where CheckApplies() is
// true but with NotBefore < EffectiveDate. This check is bypassed if
// EffectiveDate is zero.
EffectiveDate time.Time `json:"-"`
// Lints automatically returns NE for all certificates where CheckApplies() is
// true but with NotBefore > IneffectiveDate. This check is bypassed if
// IneffectiveDate is zero.
IneffectiveDate time.Time `json:"-"`
```

...with the consequent updates to the framework and documentation. I don't think I see any reason why any extant lint would need a code change to accommodate something like this.

@zakird @cpu @sleevi and @robplee seemed to be the interested parties here.

christopher-henderson:
We would like to add a mechanism for a lint to declare that it is no longer effective for certificates issued past a certain date. This involves:

1. Updating the `Lint` struct appropriately.
2. Updating both the `CheckEffective` body and documentation. Note that I reorganized the boolean logic to be a bit clearer (that is, I find it much easier to read these expressions when their plain English translation aligns more closely to the documentation).
3. Rewrote the unit tests since the addition of the upper bounds date greatly expanded the truth table under consideration.

I'm not particularly married to the name `IneffectiveDate`. It's just, ya know, the antonym to `EffectiveDate` so :man_shrugging: Maybe refactor the entire project to use `NotBefore` and `NotAfter`? Contributors would clearly recognize the concept at play there, but then I worry that that may cause confusion between the Linter's `NotBefore/After` and those found on certificates.

This PR resolves #575

### DIFF
--- a/v3/lint/base.go
+++ b/v3/lint/base.go
@@ -58,20 +58,35 @@ type Lint struct {
 
 	// Lints automatically returns NE for all certificates where CheckApplies() is
 	// true but with NotBefore < EffectiveDate. This check is bypassed if
-	// EffectiveDate is zero.
+	// EffectiveDate is zero. Please see CheckEffective for more information.
 	EffectiveDate time.Time `json:"-"`
+
+	// Lints automatically returns NE for all certificates where CheckApplies() is
+	// true but with NotBefore >= IneffectiveDate. This check is bypassed if
+	// IneffectiveDate is zero. Please see CheckEffective for more information.
+	IneffectiveDate time.Time `json:"-"`
 
 	// The implementation of the lint logic.
 	Lint LintInterface `json:"-"`
 }
 
-// CheckEffective returns true if c was issued on or after the EffectiveDate. If
-// EffectiveDate is zero, CheckEffective always returns true.
+// CheckEffective returns true if c was issued on or after the EffectiveDate
+// AND before (but not on) the Ineffective date. That is, CheckEffective
+// returns true if...
+//
+// 	c.NotBefore in [EffectiveDate, IneffectiveDate)
+//
+// If EffectiveDate is zero, then only IneffectiveDate is checked. Conversely,
+// if IneffectiveDate is zero then only EffectiveDate is checked. If both EffectiveDate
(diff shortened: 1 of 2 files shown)


zakird (approved):
LGTM

v3/lint/base.go
```diff
@@ -58,20 +58,34 @@ type Lint struct {
...
+// if IneffectiveDate is zero then only EffectiveDate is checked. If both EffectiveDate
+// and IneffectiveDate are zero then CheckEffective always returns true.
 func (l *Lint) CheckEffective(c *x509.Certificate) bool {
-	if l.EffectiveDate.IsZero() || !l.EffectiveDate.After(c.NotBefore) {
-		return true
-	}
-	return false
+	afterOrOnEffective := l.EffectiveDate.IsZero() ||
+		c.NotBefore.After(l.EffectiveDate) ||
+		c.NotBefore.Equal(l.EffectiveDate)
```
dadrian, line 85: This allows equality where it previously did not. Is that intentional?

dadrian (approved):

v3/lint/base.go
christopher-henderson: So equality actually was previously allowed.

```go
func TestEquals(t *testing.T) {
	effective := time.Unix(5, 0)
	notBefre := time.Unix(5, 0)
	t.Log(!effective.After(notBefre))
}
```

It's just that when I was cooking up this patch that fact was really hard to see. In fact, I had to whip up some small tests like the above to figure it out, which is why I rearranged these expressions to be a bit more obvious.

dadrian: Ah, my mistake. Good catch!

PR #578, opened 2021-03-14
Update zcrypto for vendored crypto/dsa package

christopher-henderson:
This check has been failing master silently for several days.

https://github.com/zmap/zlint/actions/runs/651345041

Essentially, it doesn't seem to like that we import `crypto/dsa` at any point.


zakird:
> SA1019: package crypto/dsa is deprecated: DSA is a legacy algorithm, and modern alternatives such as Ed25519 (implemented by package crypto/ed25519) should be used instead. Keys with 1024-bit moduli (L1024N160 parameters) are cryptographically weak, while bigger keys are not widely supported. Note that FIPS 186-5 no longer approves DSA for signature generation. (staticcheck)

This might actually be something that we want to resolve rather than commenting out. If the DSA package is deprecated, it raises the question, "When will the package be removed?" and our code break. Does it make more sense to instead move this to ZCrypto package?

christopher-henderson:
@zakird I reckon that the question of when we can remove our dependence on this package is largely predicated on the question of when we can remove any lints that might ask any questions about DSA (including lints that insist that it _not_ be algorithm of choice of a particular cert).

But really, when this error says that `package crypto/dsa is deprecated` I sort of reckon they don't mean "this package will get straight up deleted". I'm reading that as, "please _please_ don't use this, please use something better - I am begging you".

.golangci.yaml
```diff
@@ -33,6 +33,7 @@ linters:
     - goconst
     - golint
     - godot
+    - staticcheck
```
cpu, line 36: I don't think we should disable this linter globally. I would rather see files excluded for this lint in the config, or the lint disabled inline where required with `// nolint:staticcheck` comments. The [docs detail the available options](https://golangci-lint.run/usage/false-positives/).

cpu (changes requested):

cpu:
> This check has been failing master silently for several days.

@christopher-henderson Nice catch! It might make sense to pair the cron builds with an email notification for failures to make this more obvious in the future.

> Does it make more sense to instead move this to ZCrypto package?

@zakird agreed, this seems like the best longer term fix :thumbsup:

.golangci.yaml
christopher-henderson: Ya know, I tried that exact documentation when we were looking at custom linters awhile back it just does not appear to actually work. =/ I'm not sure if I was doing something wrong, but I remember there being a slew of issues and PR opens claiming that various "ignore this block of code" pragmas don't actually work.

christopher-henderson:
I'd be amenable to that. Would that essentially mean vendoring the `dsa` package under ZCrypto and changing all of these imports to `zcrypro/dsa`?

zakird:
Yes

.golangci.yaml
cpu: Weird, I haven't had trouble with it in the past. I've been using `//nolint`  comments in other projects 🤷‍♂️

christopher-henderson: Yeah, alternatively, this could be PEBKAC so who knows.

cpu (approved):
Thanks @christopher-henderson

Issue #538, opened 2021-01-08
e_dnsname_empty_label is a cabf_br lint, but should be an RFC lint


sleevi:
The lint currently checks that a DNSName does not contain an empty label, with reference to 7.1.4.2 of the BRs.

However, the source of this lint if RFC 5280, namely [Section 4.2.1.6](https://tools.ietf.org/html/rfc5280#section-4.2.1.6)

>   `When the subjectAltName extension contains a domain name system`
>   `label, the domain name MUST be stored in the dNSName (an IA5String).`
>   `The name MUST be in the "preferred name syntax", as specified by`
>   `Section 3.5 of [RFC1034] and as modified by Section 2.1 of`
>   `[RFC1123].  Note that while uppercase and lowercase letters are`
>   `allowed in domain names, no significance is attached to the case.  In`
>   `addition, while the string " " is a legal domain name, subjectAltName`
>   `extensions with a dNSName of " " MUST NOT be used. `

RFC 1034 gives the preferred name syntax with an ABNF of (abbreviated)
> `<domain> ::= <subdomain> | " "`
> `<subdomain> ::= <label> | <subdomain> "." <label>`
> `<label> ::= <letter> [ [ <ldh-str> ] <let-dig> ]`

As label always has at least a letter (in 1123, a let-dig), an empty label is not in the "preferred name syntax")

sleevi:
This also includes
`cabf_br/lint_dnsname_hyphen_in_sld`
`cabf_br/lint_dnsname_label_too_long`
`cabf_br/lint_dnsname_underscore_in_sld`
`cabf_br/lint_dnsname_underscore_in_trd`

sleevi:
For reference: The underscore ones are reiterated in 7.1.4.2.1, but they're merely capturing what 5280 states.

PR #586, opened 2021-04-11
Updating RFC surname and givenname character limits

christopher-henderson:
@sleevi I went ahead with an implementation of the idea I threw out yesterday. Lemme know if this indeed palatable.

```
if <= 64 {
    pass
} else if <= 32768 {
   warn
} else {
   error
}
```

Additionally, for the warnings, I've written the following details to try to help out operators understand what is going on here. I would appreciate a smoke check that my explanation is accurate.

> Although RFC 5280 prescribes a maximum length of 32768, the 1988 ITU-T for x.509 prescribes a maximum length of 64 characters. As such, your %d character long surname may encounter interoperability issues with systems that were built targeting  x.509's 64 character limit.

The following two certificates from the corpus no longer produce errors as their subject general names (`Johannes Elias Markus` and `SALVADOR CRISTOBAL` respectively) were over the previous limit.

```
Certificate:
    Data:
        Version: 3 (0x2)
        Serial Number: 3152400 (0x301a10)
        Signature Algorithm: sha256WithRSAEncryption
        Issuer: C = LU, O = LuxTrust S.A., CN = LuxTrust Global Qualified CA 3
        Validity
            Not Before: Nov 17 10:42:16 2017 GMT
            Not After : Nov 17 10:42:16 2020 GMT
        Subject: emailAddress = Johannes.Bihas@publications.europa.eu, C = FI, L = LU, O = PUBLICATIONS OFFICE, OU = NONE, CN = Johannes Elias Markus BIHAS, SN = BIHAS, GN = Johannes Elias Markus, serialNumber = 11103442740019010360, title = Professional Person
        Subject Public Key Info:
            Public Key Algorithm: rsaEncryption
                RSA Public-Key: (2048 bit)
                Modulus:
                    00:be:77:19:01:c8:e1:60:a7:16:0a:46:15:ad:bf:
                    2b:5f:b0:1d:22:d8:36:a5:4d:9d:ae:71:ff:9f:af:
                    34:53:a7:bd:d7:90:46:36:5d:4c:d3:c8:24:5d:62:
                    59:2c:96:d2:5e:01:94:4c:0c:a6:ff:ef:33:64:d1:
                    d1:ea:40:f7:4e:0a:fc:64:46:12:af:fd:ed:96:ea:
                    f4:4a:60:09:96:66:44:f6:c9:2f:aa:84:12:1b:4a:
                    72:43:d7:77:80:76:fd:9e:0e:e0:59:62:e0:7e:0d:
[...]

PR #587, opened 2021-04-16
Correct lint attribution for dnsname_etc lints

robplee:
Addressing issue #538 with this one.  I assume this fixes all the dnsname_* lints but I haven't checked them all - I just fixed the ones that were mentioned in the issue.

I did add a new lint source which I accept might not have been the best thing as I could have just set the lint source to 5280 and pointed the 'Citation' field in that direction also.  I am happy to redo that bit of the PR if it is preferred!


robplee:
I see that the integration test step is failing, my working theory for e_dnsname_empty_label and e_dnsname_label_too_long is that it's because I changed the effective date so it'll now check many more of the corpus certificates than before but as I didn't touch the effective date for e_dnsname_underscore_in_sld I'm a little confused as to why that would now be failing...  Thoughts would be welcome!

v3/lints/rfc/lint_dnsname_label_too_long.go
```diff
@@ -29,8 +29,8 @@ func init() {
 		Name:          "e_dnsname_label_too_long",
 		Description:   "DNSName labels MUST be less than or equal to 63 characters",
 		Citation:      "RFC 1035",
-		Source:        lint.CABFBaselineRequirements,
-		EffectiveDate: util.CABEffectiveDate,
+		Source:        lint.RFC1035,
```
christopher-henderson, line 32: So as you pointed out this might be a bit awkward, and I actually ran into this exact same issue myself in #586 .

My take on this is that we are citing 5280 which, in turns, "imports" and makes reference to 1035.

I suppose another way to say it is that the reason why this lint was written to begin with was not because of RFC1035, per se, but because RFC5280 invoked RFC1035. Only then did this character length become of interest to us.

@cpu and @zakird of course you folks have a stronger historical context than I do on the philosophy/semantics of `Source` so ya'lls two cents would carry significant weight on this.

christopher-henderson:
Hmmm your hypothesis on the failing integration tests is plausible, however the delta is high enough that we're probably going to have to write some scripts to verify. I don't have the time this exact moment, but I'll see about getting some shell|go scripts together that might help here.

christopher-henderson:
Failing Certs:
[smoke.tar.gz](https://github.com/zmap/zlint/files/6331985/smoke.tar.gz)

So @robplee I gathered up all of the failing certs and attached them above. As per your hypothesis, we would like to see 12 of these certs be effective for CABF and 292 to be effective for RFC. However, I whipped up the script below to smoke check this and found that the earliest `NotBefore` in the bunch is 2014, so all 292 are effective for both CABF and RFC. Of course, I could have fudged this script so I maybe I'm wrong myself.

Still, this should be a jumping off point to untangling this test.

```go
package main

import (
	"crypto/x509"
	"encoding/pem"
	"fmt"
	"io/ioutil"
	"os"
	"time"
)

const WHERE_THE_CERTS_BE_AT = "v3/smoke"

func main() {
	dir, err := ioutil.ReadDir(WHERE_THE_CERTS_BE_AT)
	if err != nil {
		panic(err)
	}
	cabf := 0
	rfc := 0
	earliest := 0
	for _, f := range dir {
		checkDate(toCert(f), &cabf, &rfc, &earliest)
	}
	fmt.Printf("CABF Effective: %v, RFC Effective: %v, Earliest: %v\n", cabf, rfc, earliest)
}

func toCert(f os.FileInfo) *x509.Certificate {
	b, err := ioutil.ReadFile(fmt.Sprintf("%s/%s", WHERE_THE_CERTS_BE_AT, f.Name()))
	if err != nil {
		panic(err)
	}
	p, _ := pem.Decode(b)
	cert, err := x509.ParseCertificate(p.Bytes)
	if err != nil {
		panic(err)
	}
	return cert
}

var CABEffectiveDate = time.Date(2012, time.July, 1, 0, 0, 0, 0, time.UTC)
var RFC5280Date = time.Date(2008, time.May, 1, 0, 0, 0, 0, time.UTC)

func checkDate(cert *x509.Certificate, cabf *int, rfc *int, earliest *int) {
	year := cert.NotBefore.Year()
	if year < *earliest || *earliest == 0 {
		*earliest = year
	}
	if cert.NotBefore.After(CABEffectiveDate) || cert.NotBefore.Equal(CABEffectiveDate) {
		*cabf += 1
	}
	if cert.NotBefore.After(RFC5280Date) || cert.NotBefore.Equal(RFC5280Date) {
		*rfc += 1
	}
}
```

v3/lints/rfc/lint_dnsname_label_too_long.go
cardonator: IMO, I prefer your interpretation. Perhaps additional context for why 5280 included it could be added to the description (such as `ctx. RFC1035`?

robplee:
An update for the group - @christopher-henderson and I have done a little investigation into this and the problem is that the certs weren't previously applied against the certificates they are now failing against.  We think the catch that was stopping them was this from lint/base.go

`func (l *Lint) Execute(cert *x509.Certificate) *LintResult {
}`

I've done some hacking around further to Chris' script and it very much seems that now that these lints are no longer cabf lints they may be being misapplied and that's the cause of the integration test failures.  I've not looked carefully at the three lints causing problems but, the e_dnsname_underscore_in_sld lint is throwing lint errors whenever it sees email addresses such as valid_email@nounderscoreshere.com which definitely makes it seem like something is going wrong somewhere!

The investigation continues :)

christopher-henderson:
So at least for the largest test different I believe that following is the real problem child here

https://github.com/zmap/zlint/blob/5b1042404f8ee953c125be6ed3cf351e3057e885/v3/lints/rfc/lint_dnsname_contains_empty_label.go#L56

```go
if c.Subject.CommonName != "" && !util.CommonNameIsIP(c) {
	if domainHasEmptyLabel(c.Subject.CommonName) {
	}
}
```

That is, as per 5280 4.2.1.6 we should be checking the SANs, however we are failing here because we are taking strings from the CN such as `My Super CA Ltd.`, splitting it on the `.` as though it were a DNS name, and getting an "empty" label on the right hand side of the `.`

This only manifested because we migrated this lint away from CABF which let hundreds of new certs make it past the following in base.go

https://github.com/zmap/zlint/blob/7e75dc35e04f682f0d0eb3de6d4af49ccd5db5af/v3/lint/base.go#L100

I believe this stems from #117 where we have @zakird petitioning for a check of the CommonName (https://github.com/zmap/zlint/pull/117#discussion_r135613521). I admit I'm not sure why it was desirable then, but at least as an 5280 4.2.1.6 I do not believe that the CN should be under consideration here. In fact, if we remove this check altogether then we get the integration tests back down to 10 errors (where the original was 11, so there might have been a false positive before that we are going to have to check).

christopher-henderson:
Sure enough, the 11th certificate that is failing `dnsname_contains_empty_label` in master is this [friendly fella](https://crt.sh/?q=c5e0cc80ea898c2a9e2d09b0960df895535017eac46050c02458d16c6f6841d5) that doesn't have any DNS names at all (it has an email and an `instance of other`). What is causing it to fail at all is the tell-tale trailing dot in its common name (`MONICOMP Kereskedelmi és Szolgáltató Zrt.`)

robplee:
Hi @christopher-henderson , I'm not sure if your plan with #609 was to completely supersede this PR, but I thought there might be some value in keeping this one for the lints that haven't caused any problems so your PR can focus on the real problem children ("e_dnsname_underscore_in_sld", "e_dnsname_label_too_long" and "e_dnsname_empty_label")

robplee:
I'm going to close this as the original issue has been closed with the merging of #609 so for the sake of keeping the repo cleaner, I'm going to close this

Issue #583, opened 2021-03-30
Discussion: Consider how proposed CA/B Forum SCWG profiles will be linted


sleevi:
I [recently shared](https://archive.cabforum.org/pipermail/validation/2021-March/001645.html) with the CA/B Forum the latest draft that attempts to overhaul how the Server Certificate Working Group expresses requirements on certificates, by moving from its three section approach ("root", "subordinate", "leaf") into category-specific profiles that fully express the intersections and permitted variations.

I wanted to draw attention to it for ZLint contributors, since while it's a draft, there's opportunity to flag any approaches that may make things easier or harder to lint. That is, I'd like to avoid a situation like ETSI (mentioned in https://github.com/zmap/zlint/issues/581#issuecomment-809529934 ) where things are _too_ context-dependent and subjective to be effectively linted.

To submit feedback to the CA/B Forum, questions@cabforum.org can be used (for those that aren't interested parties or employed by members), but I will also do my best to relay _problems_, although relaying _solutions_ may be a bit IPR-problematic. It's open as a PR at https://github.com/sleevi/cabforum-docs/pull/36 to provide easier diffing.

As a concrete example, the semantics of extensions like `extendedKeyUsage` or `certificatePolicies` can vary based on the type of CA certificate being issued, and whether the organization is an Affiliate or not of the issuing CA. The Affiliation status is not something we can know purely from the certificate alone. One option would be to signal some sort of "intended profile" (e.g. as a parameter to be passed through to lints as a context hint). Another option could just be as a warning/notice to require manual review.

I don't have strong opinions here, but did want to draw extra eyes to the work, in case folks read the profiles work and needed to say "Hold on, this will make linting much harder than it needs to be" :)

christopher-henderson:
First of all, this [tabular format](https://github.com/sleevi/cabforum-docs/blob/b0e0dd1699682b5e5cf9a166d00b1cd1f919f942/docs/BR.md#71211-root-ca-extensions) is _very_ helpful from a legibility perspective. It seems like so much time is spent converting prosaic English into lints that a format that is darned near machine readable is appreciated. So a huge thanks for that :smile: 

This is certainly not the first time that we have faced the Context Demon. For example, we had [#522 (What to do with ambiguous results?)](https://github.com/zmap/zlint/issues/522) which stemmed from certain lints only being deterministic if we were working within the context of the full certificate chain. And, of course, we have this week's #581 regarding ETSI and its policies that are out-of-band of the physical certificate in question.

[...]

Issue #594, opened 2021-05-01
Support for linting profiles


christopher-henderson:
Conversations in #583 led us down the path of the concept of a `profile` in ZLint.

A profile is a named grouping of lints that, taken together, refer to a published source that enumerates and names requirements into one, larger, requirement.

A profile _must_ come from a publicly published source and _should_ have a citation explaining why these requirements are grouped together.

#### Why not just use `excludeNames/includeNames`?
1. The intent of profiles is _not_ born of convenience.
    * That is, If individual operators wish to run arbitrary, privately published, groupings of lints then that is what `excludeNames/includeNames` are still for.
2. The intent is for profiles to be sourced from publically published requirements.
3. CAs will want to target these published profiles.
4. However, finding and grouping the correct lints together that match a given profile may not be trivial.
5. We would like for not every CA to have to replicate the same, magical, `excludeNames/includeNames` that correctly identifies the aforementioned profile.

cpu:
@christopher-henderson Can you write something in the issue body? I appreciate the desire to link PRs to issues but if the issues are blank it's maybe a net neutral act :-)

christopher-henderson:
Sorry @cpu in my rush last weekend I neglected to fill this ticket out with anything meaningufl :pray:

PR #595, opened 2021-05-01
Adding support for linting profiles

christopher-henderson:
# What is a Profile

A profile derives from the conversation in #583 wherein we would like to have curated lists of lints that, collectively, refer to a particular requirement from, say, the CABF.

The solution put forth in this PR is that a `profile` is essentially an alias for a comma separated list of lints submitted via `-includeNames`, however with some extra metadata attached (such as where did this `profile` come from).

# Profile Implementation

A profile does not appear to have any behavior of it's own. Rather, it is simply a declaration of lints that are logically grouped together by an external requirement. As such, we can borrow infrastructure from our lints that is similar, however much simpler.

I believe that something as simple as the following should suffice.

```go

package profiles

import "github.com/zmap/zlint/v3/lint"

func init() {
	lint.RegisterProfile(lint.Profile{
		Name:        "my_new_profile",
		Description: "In general, CAs should do all the things correctly and not do the wrong things",
		Citation:    "That one post I saw on Reddit said that this list was legit",
		Source:      lint.MyTrustWorthyUncle,
		LintNames: []string{
			"e_ca_common_name_missing",
			"e_ca_key_usage_missing",
		},
	})
}
```

I'm not sure if profiles should have a `CheckApplies`? One would think that is inherited from individual lints aggregated within the profile.

# Running a Profile

Let's say that `the_profile_in_question` is comprised of the lints `a`, `b`, `c`, `d`.

```bash
./zlint -profile=the_profile_in_question
```

Is the equivalent to running

```bash
./zlint -includeNames="a,b,c,d"
```

A new CLI flag has been added that lists each profile as line-delimited JSON (much in the same way one prints out available lints).

```bash
$ ./zlint list-profiles
{"name":"whatever","lints":["e_ca_common_name_missing","e_ca_key_usage_missing"]}
{"name":"yooodaw","description":"Fill this in...","citation":"Fill this in...","source":"Unknown","lints":[]}
{"name":"yooodaws","description":"Fill this in...","citation":"Fill this in...","source":"Unknown","lints":[]}
```

# Adding a New Profile

The convenience script `newProfile.sh <profile_name>` has been added that acts in much the same way as `newLint.sh` does.

Documentation has been added to [CONTRIBUTING.md](https://github.com/zmap/zlint/blob/4c0bbb772a3dacf1e9b379a113c8516c027c64c2/CONTRIBUTING.md#adding-new-profiles)

(resolves #595 )


README.md
```diff
@@ -97,6 +97,9 @@ Example ZLint CLI usage:
 	echo "Lint mycert.pem with all of the lints except for ETSI ESI sourced lints"
 	zlint -excludeSources=ETSI_ESI mycert.pem
 
+	echo "List available lint profiles. A profile is a pre-defined collection of lints."
```
christopher-henderson, line 100: I was apprehensive of adding an example of running a profile here because don't actually have any yet. That is, I didn't want to give an example that doesn't actually run.

v3/cmd/zlint/main.go
```diff
@@ -32,11 +32,14 @@ import (
 	"github.com/zmap/zlint/v3"
 	"github.com/zmap/zlint/v3/formattedoutput"
 	"github.com/zmap/zlint/v3/lint"
+
+	_ "github.com/zmap/zlint/v3/profiles"
```
christopher-henderson, line 36: We need to trigger the `init` functions in `/profiles` so that we can build our `FilterOptions` appropriately. If we try to add this import into `lint` then it causes an import cycle between `lint` and `profiles` :confused:

v3/cmd/zlint/main.go
```diff
@@ -192,10 +207,17 @@ func trimmedList(raw string) []string {
 // use.
 func setLints() (lint.Registry, error) {
 	// If there's no filter options set, use the global registry as-is
-	if nameFilter == "" && includeNames == "" && excludeNames == "" && includeSources == "" && excludeSources == "" {
```
christopher-henderson, line 195: Just a style change because this boolean condition was getting long.

v3/profiles/todo.go
```diff
@@ -0,0 +1,5 @@
+package profiles
```
christopher-henderson, line 1: Yeah, sorry, if you have a package with nothing in it but test files then `golangci-lint` vomits uncontrollably. We'll have to delete this once our first profile lands.

v3/cmd/zlint/main.go
cpu: Nice, looks good :-)

v3/lint/profile.go
```diff
@@ -0,0 +1,58 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package lint
+
+type Profile struct {
+	// Name is a lowercase underscore-separated string describing what a given
+	// lints a given profile aggregates.
```
cpu, line 19: Typo?

```suggestion
	// Name is a lowercase underscore-separated string describing what a given
	// profile aggregates.
```

v3/lint/profile.go
```diff
@@ -0,0 +1,58 @@
...
+	// A human-readable description of what the Profile checks. Usually copied
+	// directly from the CA/B Baseline Requirements, RFC 5280, or other published
+	// document.
+	Description string `json:"description,omitempty"`
+
+	// The source of the check, e.g. "BRs: 6.1.6" or "RFC 5280: 4.1.2.6".
+	Citation string `json:"citation,omitempty"`
+
+	// Programmatic source of the check, BRs, RFC5280, or ZLint
+	Source LintSource `json:"source,omitempty"`
```
cpu, line 31: I'm wondering whether it's acceptable to consider a `Profile`'s citations/sources as the aggregate of the citations/sources of the lints indicated in `LintNames`. If so maybe we can simplify the model a little bit and drop these two fields.

If you think there is a use-case for having profile-level citations/sources I think the comments need to be adjusted slightly. "source of the check" reads like there's a singular "check" which (IMO) conflicts with the multiple `LintNames`.

v3/lint/profile.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+// GetProfile returns the Profile for which the provided name matches Profile.Name.
+// If no such Profile exists then the `ok` returns false, else true.
+func GetProfile(name string) (profile Profile, ok bool) {
+	profile, ok = profiles[name]
+	return profile, ok
+}
+
+func AllProfiles() []Profile {
```
cpu, line 52: nit: can you add a godoc comment on this exported func too? :bow:

v3/newProfile.sh
```diff
@@ -0,0 +1,30 @@
+# Script to create new lint from template
```
cpu, line 1:
```suggestion
# Script to create new lint profile from template
```

cpu (changes requested):
Thanks @christopher-henderson. This seems like a nice extension. I left a few small comments.

Do you have thoughts on the first profile to include? There are a couple points in the PR where you work around not having a profile (e.g. w/ `todo.go`, and not including a README example). I also had a question about whether we needed a couple model fields. I think maybe we could address both of these things by including a commit in this branch that introduces a profile if there's a good first candidate. What do you think?

v3/lint/profile.go
christopher-henderson: (Sorry, this is likely an artifact that I hadn't bothered to aggregate the discussion in #583 into #594)

So these profiles aren't arbitrary groupings. The plan is that they _do_ come a from source requirement (the first impetus of which is https://github.com/sleevi/cabforum-docs/pull/36). So the idea is that a parent source lists a collection of requirements with an explanation as to _why_ they are grouped.

```
CABF/BR x.y.z

Section A:
    Due to whatever reason <requirement A>, <requirement B>, <requirement C> from RFRC5280 should...
```

```go
{
   Citation: "Due to whatever reason, these requirements from RFC5280 should...",
   Source: lint.CABFBR_X_Y_Y
}
```

> "source of the check" reads like there's a singular "check"

Effectively there is, in that if one sub-check fails then the whole check fails, and thus you have failed the profile from the cited source. A bit like how a boolean expression can comprised of multiple boolean expressions (and named).

```go
// In Chris' humble opinion the following is what is considered
// altogether a nice day.
func isANiceDay(sunny, cool, butNotTooCool, lightbreeze, bool) {
    return sunny && cool && butNotTooCool && lightbreeze
}
```
Each individual one may-or-may-not be desirable in their own context. However, grouped together they answer the single question, "By Chris' definition is this a nice day?"

v3/newProfile.sh
christopher-henderson:
```suggestion
# Script to create new profile from template
```

christopher-henderson:
@cpu 

> Do you have thoughts on the first profile to include?

The first candidates come from @sleevi's work at https://github.com/sleevi/cabforum-docs/pull/36.

1. Keep this branch around until concrete profiles are published.
    * It would be nice to have something ready on day zero, not entirely necessary from what I can see.
2. Merge empty and add profiles once they're published.
    * Awkward, like you have pointed out. Plus the framework may need tweeks once we have something concrete to work with

I'm somewhat partial to #1 as I think it makes sense to have code ready to go, but not necessarily have dead code in master.

v3/lint/profile.go
christopher-henderson: ....although that was my take on it. Since @sleevi is the first source of these, does it make sense for there to be a simple  inheritance type scheme where we can just say that a profile's citations is the list of its lints' citations?

PR #596, opened 2021-05-07
upgrade to ZCrypto with permissive asn1 parsing

dissoupov:


christopher-henderson (changes requested):
Thank you for the Zcrypto update! 

Is there any that chance we can give `cryptobyte` the same treatment and shovel that fork into ZCrypto?

dissoupov:
@christopher-henderson @codyprime @bwireman PR updated, now only with imports from `zcrypto`

zakird:
This looks good to me. I would like to give @christopher-henderson a chance to review since he originally commented on the PR.

PR #598, opened 2021-05-12
cmd: add `-version` to `zlint`, `zlint-gtld-update`.

### LINKED ISSUE #519: feature request: ZLint command line -version flag
I see that v3 was released recently. I just did an update from what I believe is v2 to v3. Is there an easy way to tell the update was successful and which version I am running?  I think a -version switch would be really useful.

cpu:
Our GoReleaser configuration already populates a `version` variable with `LDFLAGS` at build time. Prior to this commit we only included the dynamic version var in the usage output from `--help`. This made it easy to overlook. We also didn't set the dynamic var from the makefile, leaving all src builds as the static version "dev".

In this commit we add a `--version` flag to the `zlint` and `zlint-gtld-update` commands that prints the dynamic version and
exits. We also update the `makefile` so that both binaries get built with a version that includes the latest tag, and the SHA of the local git checkout, e.g. `v3.1.0-19-g0807bf95`. This should better match user expectation for CLI tools.

Resolves #519.

### DIFF
--- a/v3/cmd/zlint-gtld-update/main.go
+++ b/v3/cmd/zlint-gtld-update/main.go
@@ -50,6 +50,11 @@ const (
 )
 
 var (
+	// version is replaced by GoReleaser or `make` using an LDFlags option at
+	// build time. Here we supply a default value for folks that `go install` or
+	// `go build` directly from src.
+	version = "dev-unknown"
+
 	// httpClient is a http.Client instance configured with timeouts.
 	httpClient = &http.Client{
 		Transport: &http.Transport{
@@ -104,6 +109,8 @@ var tldMap = map[string]GTLDPeriod{
 	},
 }
 `))
+
+	printVersion bool = false
 )
 
 // getData fetches the response body bytes from an HTTP get to the provider url,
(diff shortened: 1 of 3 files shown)


cpu:
(_Also if you happen to be especially paranoid and notice that compared to normal I didn't sign these commits.... Yes..... Subkeys expired 2021-04-19 and the master is offline. Will fix shortly :sweat:_)

zakird (approved):

cpu:
Allow me these brief theatrics to justify digging out my offline master for another year:
```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

(Also if you happen to be especially paranoid and notice that compared to normal I didn't sign these commits.... Yes..... Subkeys expired 2021-04-19 and the master is offline. Will fix shortly sweat)
-----BEGIN PGP SIGNATURE-----

iQEzBAEBCgAdFiEEgCbSSqlmv5zTzcs8CPsr/EcOdbQFAmCjC2UACgkQCPsr/EcO
dbRP9Qf/SjUn4ham4atgfQfMNwNa3+JLvgRP1wbamjK5dW5ZV0nRalT4jLplXFF0
CcVxSkuwth2KKtLR5jkIj56ZiPh5G4N4cwwvPxhD7CbODBk7T5vpPbZtg96TGDyg
glI6e5/HRR5B2LmodDS9ah2a/n7ua21uedfB95TKMp+1t7LQcI5x1hi5bmq4rnmf
27UwFIwtDi0K4hI1PFUm+niupeaYwdhCbNfxBKIWk8UdolGF8gHJG0dfBOosIAWw
BNmo5kK4uaJgu+5e3/I2sw/8sll+F+BLhbNY2Pr9lOiCfFBuJqmIh53yDVfqMmrP
NTp9IpQDkD4jZlhvf3bcIUuMdId+hg==
=YPmt
-----END PGP SIGNATURE-----
```

christopher-henderson:
lol I enjoy a good easter egg

Issue #606, opened 2021-05-30
Swap out the Initialize method for proper constructors


christopher-henderson:
The `LintInterface` requires the method [`Initialize`](https://github.com/zmap/zlint/blob/7e75dc35e04f682f0d0eb3de6d4af49ccd5db5af/v3/lint/base.go#L27). It seems very reasonable to allow for struct initialization, however the `Initialize` method runs into several issues:

1. This is really just a constructor, and it's a bit awkward to have the constructor be called into post allocation in such a way.
2. `Initialize` returns an `error` during registration, [which immediately halts registration and exits the program with the raw error message returned](https://github.com/zmap/zlint/blob/7e75dc35e04f682f0d0eb3de6d4af49ccd5db5af/v3/cmd/zlint/main.go#L90). This is, in a way, even worse than simply panicking because all stack information is lost - we have no idea which lint produced the error.
3. ...except we sort can figure out which lint caused the error because there are currently a grand total of two lints that actually do anything in their `Initialize` (and these actions are really just normal constructor stuff) - [lint_dnsname_bad_character_in_label](https://github.com/zmap/zlint/blob/7e75dc35e04f682f0d0eb3de6d4af49ccd5db5af/v3/lints/cabf_br/lint_dnsname_bad_character_in_label.go#L40) and [lint_rsa_public_exponent_not_in_range](https://github.com/zmap/zlint/blob/7e75dc35e04f682f0d0eb3de6d4af49ccd5db5af/v3/lints/cabf_br/lint_rsa_public_exponent_not_in_range.go#L46).

### But the biggest problem is
4. Due to our use of a global registry all lints are singleton objects at runtime. This is fine at this exact moment in time, however it has caused extreme pain when attempting to implement #592. In #592 I like to do something like the following:

```go
type Configurable interface {
    Configure() interface{}
}
```

...so that a lint can receive further context as necessary. It simply gives us a struct to deserialize a configuration format into.

```go
struct MyLint {
    configuration struct {
        the_rum_is_gone bool
    }
}

func (m *MyLint) Configure() interface{} {
    return &m.configuration
}

func (m *MyLint) Execut(c *x509.Certificate) *lint.Result {
    if m.configuration.the_rum_is_gone {
        return &lint.Result{Status: lint.Err, Details: "Why is the rum gone!?"}
    }
    return &lint.Result{Status: lint.Pass}
}
```

[...]

PR #607, opened 2021-05-30
Swap out the Initialize method for proper constructors

christopher-henderson:
This PR addresses #606 (where I have much of the impetus and writeup for work in question).

Broadly speaking, to accomplish this, this change implements...

1. Swaps out the `Initialize` method from `LintInterface`
2. Changes `Lint.Lint` to be a function that returns a `LintInterface` rather than a `LintInterface` itself.
3. Removed all of the wiring and testing around `Initialize`
4. Updated the template script to auto generate a constructor.
5. Applied a find-and-replace on every lint to migrate them over to constructors.

For the sake of sharing the joy/pain, here's the regexes in question fed through Goland

```
Match: func\(\) lint.LintInterface \{ return &([a-zA-Z])(.*)\{\} \},\n(\s*)\}\)\n\}\n\n
Replace: New\u$1$2,\n$3\}\)\n\}\n\nfunc New\u$1$2() lint.LintInterface{\n\treturn &$1$2{}\n}\n\n
```

All of the lints appear to be fully functional, although I can imagine there be formatting that might be off-kilter after this find and replace (like a comment that is left orphaned or is out of order now).


v3/lint/registration.go
```diff
@@ -127,25 +127,12 @@ func (e errDuplicateName) Error() string {
 		e.lintName)
 }
 
-// errBadInit is returned from registry.Register if the provided lint's
```
christopher-henderson, line 130: No longer needed this aspect of the test.

v3/lints/apple/lint_ct_sct_policy_count_unsatisfied.go
```diff
@@ -33,15 +33,16 @@ func init() {
 		Citation:      "https://support.apple.com/en-us/HT205280",
```
christopher-henderson, line 33: From here on out it is (thankfully) pretty repetitive. It makes a new constructor for every type, deletes `Initialize`, and replaces `Lint: &structName{}` with `Lint: NewStructName`.

v3/newLint.sh
```diff
@@ -44,6 +44,7 @@ FILENAME=${LINTNAME:2}
 STRUCTNAME=$3
 
 sed -e "s/PACKAGE/${PATHNAME}/" \
+    -e "s/PASCAL_CASE_SUBST/${STRUCTNAME^}/g" \
```
christopher-henderson, line 47: Today I learned that `{$varname^}` will uppercase the first character in the string. Bash is endless.

dadrian (approved):
Apologies for the delay...went through this when you first posted it but didn't click approve.

Good work!

cpu (approved):
This makes sense to me! Thanks for all of the context in the PR desc/issue.

christopher-henderson:
Thank you folks! I was eyeing the merge button over the weekend, but then a pinch of food poisoning kicked me in the teeth. I appreciate ya'll driving it to completion.

Issue #541, opened 2021-01-08
e_sub_ca_aia_missing is no longer an error as of BRs 1.7.1


sleevi:
[BRs 1.7.1](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.7.1.pdf) adopted the [Browser Alignment Ballot SC31](https://cabforum.org/2020/07/16/ballot-sc31-browser-alignment/), which made AIA **optional** for subordinate CA certificates.

Previously, CAs were required to include OCSP responders for subordinate CAs. However, the new language of 7.1.2.2(c) reads:
> This extension SHOULD be present. It MUST NOT be marked critical.
> It SHOULD contain the HTTP URL of the Issuing CA’s certificate (accessMethod =
1.3.6.1.5.5.7.48.2). It MAY contain the HTTP URL of the Issuing CA’s OCSP responder
(accessMethod = 1.3.6.1.5.5.7.48.1)

PR #613, opened 2021-06-25
e_sub_ca_aia_missing is no longer an error as of BRs 1.7.1

robplee:
This is my fix to address the issue raised by @sleevi in #541.  From what was in the lint before it seems that the issue was that it was enforcing rules which have now been downgraded to a SHOULD from the previous MUST.

Therefore, I've duplicated the lint and changed the old one to be ineffective from CABF BR 1.7.1 date and to make the new one effective from then but return warnings instead of errors.

I also rewrote the tests for the original lint because it was easier to just copy over the tests I wrote for the warning version and change a few bits here and there than to work with the previous tests that were already there.

I noticed there were more changes to the config than just those concerned with the new lint in this PR but I've included all changes because certainly one of them looks like it should have been included in my #572 PR!  Happy to remove excess changes if people would like me to do so.


v3/integration/config.json
```diff
@@ -755,12 +755,8 @@
     "w_subject_dn_trailing_whitespace": {
       "WarnCount": 64
     },
-    "w_tls_server_cert_valid_time_longer_than_397_days": {},
-    "e_cert_sig_alg_not_match_tbs_sig_alg": {
-      "ErrCount": 10
```
christopher-henderson, line 760: ...and on this one I spent another good chunk of time trying to figure out why both having this and removing it still results in a pass in the corpus.

...

It's because it's a duplicate of the same entry at L296.

image

v3/integration/config.json
```diff
@@ -755,12 +755,8 @@
...
-    },
-    "e_ocsp_id_pkix_ocsp_nocheck_ext_not_included_server_auth": {
```
christopher-henderson, line 762: Same as above, duplicate of L450

christopher-henderson (approved):
Thank you @robplee, I'm glad to see the `IneffectiveDate` field seeing such effective use!

I wrote #615 for myself because the time I spent understanding why the changes to the integrations test config worked was a silly waste of time. I should probably also enforce that you can't have an entry for a lint that doesn't exist.

PR #614, opened 2021-06-26
Update README.md

defacto64:
Added Actalis to the bulleted list "Zlint Users/Integrations".


zakird (approved):
Thanks for the contribution! It's great to see ZLint in use.

PR #616, opened 2021-06-26
disallow duplicate entries in config.json

### LINKED ISSUE #615: Disallow duplicate entries in the test corpus config
Since the file gets parsed into a hashmap the later entry will override the former, which can be surprising. Additionally, it is extremely surprising to think that you have completely deleted an entry and yet certain conditions are still being asserted. See #613 for an example of a morning wasted trying to figure this out.

christopher-henderson:
#613 providing me with quite a painful experience of trying to figuring out why deleting and modifying those particular entries in `config.json` did not have the desired affect. That is, changing the entries would fail the test, but deleting them would not, which was quite surprising. It turns out it was because they were duplicates of entries already declared further up in the document.

This change disallows duplicate entries in the integration test JSON as I just don't see a usecase for it and it can cause some confusing results that waste your time.

This PR resolves #615


v3/integration/config.go
```diff
@@ -105,10 +106,44 @@ func loadConfig(file string) (*config, error) {
 	if err := json.Unmarshal(jsonBytes, &c); err != nil {
 		return nil, err
 	}
-
+	problems := findProblemsInTheConfig(jsonBytes, &c)
+	if len(problems) != 0 {
+		return nil, errors.New(strings.Join(problems, "\n"))
```
christopher-henderson, line 111: Example output is:

```
the lint 'e_cert_sig_alg_not_match_tbs_sig_alg' was declared 2 times and appeared on line numbers [296 759]
the lint 'e_ocsp_id_pkix_ocsp_nocheck_ext_not_included_server_auth' was declared 2 times and appeared on line numbers [450 762]
```

v3/integration/config.go
```diff
@@ -105,10 +106,44 @@ func loadConfig(file string) (*config, error) {
...
+	}
 	return &c, nil
 }
 
+// findProblemsInTheConfig tries keep the configuration honest with regard
+// to aspects such as duplicate entries with in the Expected field.
+func findProblemsInTheConfig(configBytes []byte, c *config) []string {
```
robplee, line 118: Do you have thoughts about other config problems you may try to find with this function in the future?  Otherwise, findDuplicateConfigEntries might be a more precise name for the function.

christopher-henderson: Yeah, I was thinking of other things like not being able to declare a result set for a lint that doesn't exist (which also happened in #613 )

robplee: Fair enough, the name seems reasonable then.

robplee (approved):

PR #618, opened 2021-06-30
Lint that the EKU values within a SubCA certificate comply with CABF_BR 7.1.2.2 letter g)

defacto64:
Adds a new lint, identified as e_sub_ca_eku_incmpatible_values,
to check that the EKU values within a Subordinate CA certificate meet the 
MUST NOT requirement of CABF Baseline Requirements 7.1.2.2, letter g).


defacto64:
I will re-do this pull after gofmt'ing the two .go files.

Issue #592, opened 2021-04-25
Support for lints with extended contexts


christopher-henderson:

cpu:
:bow: request for an issue description. I can infer from the link to #581 and the title but it would be better for everyone if we can capture more context here.

Issue #581, opened 2021-03-29
Bug: ESI lints are not consistent with ETSI EN 319 412-5 requirements regarding non-qualified certificates


attilarozgonyi:
When non-qualified certificates containing QC Statements are used, zlint returns an error for e_qcstatem_mandatory_etsi_statems. By the documentation's description, this lint "Checks that a QC Statement that contains at least one of the ETSI ESI statements, also features the set of mandatory ETSI ESI QC statements." and the description references "ETSI EN 319 412-5 V2.2.1 (2017 - 11) / Section 5". 
This assumption is however incorrect due to the mentioned clause features requirements for qualified certificates only, but there are ETSI-defined statements that can be used in other than qualified certificates, eg. esi4-qcStatement-2 or any other statements defined in subclause 4.3.

The same problem goes to w_qcstatem_qctype_web which "Checks that a QC Statement of the type Id-etsi-qcs-QcType features at least the type IdEtsiQcsQctWeb", and references "ETSI EN 319 412 - 5 V2.2.1 (2017 - 11) / Section 4.2.3", which is titled "QCStatement claiming that the certificate is a EU qualified certificate of a particular type", thus does not apply to non-qualified certificates (the new version, V2.3.1 adds non-qualified certificates to the subclause, but the usage is still not mandatory, only „when the certificate is issued in accordance with Annex III or Annex IV of Regulation (EU) No 910/2014”). Also, I understand that zlint is mainly focused on website authentication certificates but if other types (eg. electronic signature certificates) are in scope, this lint should be broadened to accept other identifiers than IdEtsiQcsQctWeb as well.

The solution to this would be either to fix the above-mentioned lints based on EN 319 412-5 requirements with differentiating between (EU) qualified and non-qualified certificates, or to omit these lints completely.

I have seen a similar bug (https://github.com/zmap/zlint/issues/424) that covers this one’s main idea, but only regarding e_qcstatem_mandatory_etsi_statems, and it contains the proposal of reducing the error to a warning which, following the fix, in my understanding, would still as well warn regarding a 100% ETSI-compliantly issued non-qualified certificate.

sleevi:
@zakird @dadrian @cpu Given that @mtgag has mostly trailed off contributing, how do you feel about removing the ETSI lints entirely, and closing out the related items?

[...]

PR #619, opened 2021-07-01
Lint that the EKU values within a SubCA certificate comply with CABF_BR 7.1.2.2 letter g)

defacto64:
Adds a new lint, identified as e_sub_ca_eku_incmpatible_values,
to check that the EKU values within a Subordinate CA certificate meet the
MUST NOT requirement of CABF Baseline Requirements 7.1.2.2, letter g).

### DIFF
--- a/v3/integration/config.json
+++ b/v3/integration/config.json
@@ -783,6 +783,9 @@
     "w_tls_server_cert_valid_time_longer_than_397_days": {},
     "e_no_underscores_before_1_6_2": {
       "ErrCount": 340
+    },
+    "e_incorrect_ku_encoding": {
+      "ErrCount": 6586
     }
   }
 }
--- a/v3/lints/rfc/lint_incorrect_ku_encoding.go
+++ b/v3/lints/rfc/lint_incorrect_ku_encoding.go
@@ -1,0 +1,80 @@
+/*
+ * ZLint Copyright 2022 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package rfc
+
+import (
+	"fmt"
+	"math/big"
+	"strings"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_incorrect_ku_encoding",
+		Description:   "RFC 5280 Section 4.2.1.3 describes the value of a KeyUsage to be a DER encoded BitString, which itself defines that all trailing 0 bits be counted as being \"unused\".",
+		Citation:      "Where ITU-T Rec. X.680 | ISO/IEC 8824-1, 21.7, applies, the bitstring shall have all trailing 0 bits removed before it is encoded.",
+		Source:        lint.RFC5280,
+		EffectiveDate: util.ZeroDate,
+		Lint:          func() lint.LintInterface { return &incorrectKuEncoding{} },
+	})
+}
+
+type incorrectKuEncoding struct{}
+
+func NewIncorrectKuEncoding() lint.LintInterface {
+	return &incorrectKuEncoding{}
+}
+
+func (l *incorrectKuEncoding) CheckApplies(c *x509.Certificate) bool {
+	ku := util.GetExtFromCert(c, util.KeyUsageOID)
+	return ku != nil && len(ku.Value) > 0
+}
+
+func (l *incorrectKuEncoding) Execute(c *x509.Certificate) *lint.LintResult {
+	ku := util.GetExtFromCert(c, util.KeyUsageOID).Value
+	if len(ku) < 4 {
+		return &lint.LintResult{
+			Status:  lint.Fatal,
+			Details: fmt.Sprintf("KeyUsage encodings must be at least four bytes long. Got %d bytes", len(ku)),
+		}
+	}
+	// Byte 0: Tag
+	// Byte 1: Length
+	// Byte 2: Unused bits
+	// Bytes 3..n: KeyUsage
+	declaredUnused := uint(ku[2])
+	actualUnused := big.NewInt(0).SetBytes(ku[3:]).TrailingZeroBits()
+	if declaredUnused == actualUnused {
+		return &lint.LintResult{Status: lint.Pass}
+	}
+	// Just a bit of formatting to a visualized binary form so
+	// it's easier for users to see what the exact binary that
+	// we're referring to so that they can debug their own certs.
+	binary := make([]string, len(ku))
+	for i, b := range ku {
+		binary[i] = fmt.Sprintf("%08b", b)
+	}
+	return &lint.LintResult{
+		Status: lint.Error,
+		Details: fmt.Sprintf(
+			"KeyUsage contains an inefficient encoding wherein the number of 'unused bits' is declared to be "+
+				"%d, but it should be %d. Raw Bytes: %v, Raw Binary: [%s]",
+			declaredUnused, actualUnused, ku, strings.Join(binary, " "),
+		)}
+}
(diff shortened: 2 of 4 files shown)


v3/lints/cabf_br/lint_sub_ca_eku_incompatible_values.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_sub_ca_eku_incompatible_values",
+		Description:   "Subordinate CA extkeyUsage: if serverAuth is present, then emailProtection, codeSigning, timeStamping, and anyExtendedKeyUsage MUST NOT be present.",
```
sleevi, line 45: So this is tricky, because 7.1.2.2 contains the following language:

> For Cross Certificates that share a Subject Distinguished Name and Subject Public Key with a Root Certificate operated in accordance with these Requirements, this extension MAY be present. If present, this extension SHOULD NOT be marked critical. This extension MUST only contain usages for which the issuing CA has
verified the Cross Certificate is authorized to assert. This extension MAY contain the anyExtendedKeyUsage [RFC5280] usage, if the Root Certificate(s) associated with this Cross Certificate are operated by the same organization as the issuing Root Certificate.

As is, this lint would fire an Error for those certificates, due to lacking sufficient context about whether or not the given certificate being linted meets that language, right?

Should this lint be blocked on https://github.com/zmap/zlint/issues/592 (see also https://github.com/zmap/zlint/issues/581#issuecomment-809529934 for more discussion about that issue)

v3/lints/cabf_br/lint_sub_ca_eku_incompatible_values_test.go
```diff
@@ -0,0 +1,67 @@
...
+ * permissions and limitations under the License.
+ */
+
+import (
+	"testing"
+
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/test"
+)
+
```
sleevi, line 23: It probably makes sense to switch this to a so-called "table based test" - e.g. https://github.com/zmap/zlint/blob/48b300e877f2ed3e6b7e75426a273f84455f17a6/v3/lints/mozilla/lint_e_prohibit_dsa_usage_test.go#L24

Eliminates a lot of redundancy, which in this case, every body is more or less the same other than the filename and expected result.

v3/lints/cabf_br/lint_sub_ca_eku_incompatible_values_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func TestSubCAEkuServerAndAny(t *testing.T) {
+	inputPath := "SubCAEkuServerAndAny.pem"
+	expected := lint.Error
+	out := test.TestLint("e_sub_ca_eku_incompatible_values", inputPath)
+	if out.Status != expected {
+		t.Errorf("%s: expected %s, got %s", inputPath, expected, out.Status)
+	}
+}
```
sleevi, line 67: Note: You don't have any tests where EKU is present, but serverAuth is **not** present, while other combinations are. For example, e-mail and code-signing.

v3/lints/cabf_br/lint_sub_ca_eku_incompatible_values.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (l *subCAEkuIncompatibleValues) Execute(c *x509.Certificate) *lint.LintResult {
+
+	serverAuthPresent := false
+	emailProtectionPresent := false
+	codeSigningPresent := false
+	timeStampingPresent := false
+	anyEkuPresent := false
+
```
sleevi, line 70: Note that 7.1.2.2(g) also includes this requirement:

> For Subordinate CA Certificates that are not used to issue TLS certificates, then the value id-kp-serverAuth [RFC5280] MUST NOT be present. Other values MAY be present, but SHOULD NOT combine multiple independent key purposes (e.g. including id-kp-timeStamping [RFC5280] with id-kp-codeSigning [RFC5280]).

This code doesn't seem to test for that. I think it should, right?

sleevi (changes requested):
Thanks for sending this!

While I'm requesting changes here, I'd definitely want to hear from @christopher-henderson on the issue of context, since that might be a blocker to landing this. Alternatively, we may have to make this simply a WARNING, not an error, so that it gets flagged for manual review but doesn't block issuance.

v3/lints/cabf_br/lint_sub_ca_eku_incompatible_values.go
defacto64: That's right, Ryan. My lint is specifically targeting the following part of BR 7.1.2.2(g):

`For all other Subordinate CA Certificates, including Technically ConstrainedSubordinate CA Certificates:`
`For Subordinate CA Certificates that will be used to issue TLS certificates, the valueid-kp-serverAuth[RFC5280] MUST be present. The valueid-kp-clientAuth[RFC5280] MAY be present. The valuesid-kp-emailProtection[RFC5280],id-kp-codeSigning[RFC5280],id-kp-timeStamping[RFC5280], andanyExtendedKeyUsage[RFC5280]MUST NOT be present. `

It was not my intention to cover the entirety of BR 7.1.2.2(g), but only the case where the serverAuth kp is present.

v3/lints/cabf_br/lint_sub_ca_eku_incompatible_values.go
christopher-henderson: @sleevi  So I have a branch up at https://github.com/zmap/zlint/tree/context that at least gets us the idea of what we want.

Take the following lint into consideration.


```go
type MyLint struct {
	BottlesOfBeerOnTheWall int
}

func init() {
	lint.RegisterLint(&lint.Lint{
		Name:          "e_ca_too_few_beers",
		Description:   "CA Certificates MUST have at least 99 beers.",
		Citation:      "BRs: 7.1.4.3.1",
		EffectiveDate: util.CABV148Date,
		Lint:          NewMyLint,
	})
}

// This is the new thing. You give us a struct to deserialize
// into and we will get the appropriate context into it.
//
// In this case, the lint itself holds the data in quesiton.
func (l *MyLint) Configure() interface{} {
	return l
}

func NewCaCommonNameMissing2() lint.LintInterface {
	return &MyLint{}
}

func (l *MyLint) CheckApplies(c *x509.Certificate) bool {
	return util.IsCACert(c)
}

func (l *MyLint) Execute(c *x509.Certificate) *lint.LintResult {
	if l.BottlesOfBeerOnTheWall < 99 {
	} else {
	}
}
```

From there we can build the additional context file...

```bash
$ cat >> ctx.toml <<EOF
[e_ca_too_few_beers]
BottlesOfBeerOnTheWall = 98

[e_some_other_lint]
is_web_cert = false
EOF
```

...and shove it into ZLint

```bash
$ ./zlint -nameFilter e_ca_too_few_beers -context ctx.toml cert.pem
{"e_ca_too_few_beers":{"result":"error","details":"Time for a beer run!"}}
```

Of course it needs more testing/user acceptance testing and it needs integrations into the large test corpus.

Additionally, the framework only configures on a per lint basis. That is, each section in TOML document lines up to a specific lint, however I am not passing in any sort of global context to all lints. This is only because I am apprehensive about either maintaining a well defined global struct _or_ passing anonymous `map[string]interface{}` to every lint (because that just sounds chaotic).

defacto64:
Indeed this lint may raise an error _erroneously_, so to speak, upon a cross-certificate that contains the EKU extension AND the EKU contains serverAuth. This lint was not intended to cover that case, and at any rate it does not seem possible to determine if a CA certificate under examination is a cross-certificate unless extra information is passed to Zlint, and this seems tricky.

However, considering that cross-certificates are a microscopic minority of all CA certificates, and that only some of them would trigger this lint (depending on whether they contain the EKU and depending on its value), in my opinion it is acceptable that this lint produces a "false positive" (that is, an error) in those very few cases, which can then be managed with an ad hoc procedure (e.g. by temporarily bypassing the linter). How about this?

sleevi:
We’ve intentionally tried to avoid this in the past, because unreliable
lints mistakenly encourage CAs to believe ZLint as a whole is unreliable.

My own view is that we should properly lint the requirements completely,
and use warnings for if we’re unable to detect errors.

cardonator:
I have to agree that any scenario that would return an Error (which impacts issuance) in a situation that is non-deterministic, the lint should be lowered to a Warning despite that not lining up exactly with the BR terminology as discussed in the README.

defacto64:
I appreciate the reasoning and all in all I agree. I am therefore available to revise the lint so that it returns a warning instead of an error, being the alternatives (e.g. passing additional context to Zlint) too tricky and lengthy. This is an easy change and I can do it quickly, then submit a new pull request. @sleevi, @christopher-henderson, @cardonator, would this be fine with you?

Please note that the aim of this lint was specifically that of checking compliance with the first "MUST NOT" of BRs 7.1.2.2 letter g). In view of this, the lint's description may seem too wide-scoped, and I would agree to narrow it. Otherwise, in order to also cover the case of "Subordinate CA certificates not used to issue TLS certificates" (which was not my intention) I'd have to add a bit more code and test cases. Nothing too difficult, but it could be handled as a subsequent refinement ...

Issue #620, opened 2021-07-16
Test certificate generator doesn't create a certificate chain


robplee:
In #569, @christopher-henderson added a helpful tool for generating certificates and chains.  However, it currently contains a bug in that the certificate chain produced is a chain in name only and not cryptographically as all the certs are self-signed.

This means that the script always has to be hacked a bit to generate certificates which don't return true on util.SelfSigned(), this came up in the course of generating the test data for #613 which needed an intermediate cert for which util.IsCACert() would return true but for which util.IsRootCert would return false.

Therefore, I think it would be worthwhile to fix the various newThing functions to take the signing key as an argument (for newIntermediate and newLeaf) and return the key included in the certificate created by the function so it can be passed on to the next function call.

If this issue sounds like something which should be addressed then I'm happy to work on it myself.

christopher-henderson:
@robplee I'll look at. It _should_ be generating a self signed trust anchor and then chaining that down appropriately. But it could have a bug in it, yeah.

robplee:
So, the previous cert is passed in as the parent certificate to the call to CreateCertificate but then the private key used in the call to CreateCertificate is that of the key pair generated in the function and it's that key, priv, which is used to sign according to https://pkg.go.dev/crypto/x509#CreateCertificate

christopher-henderson:
Argggh good catch @robplee. I'll fix it up this weekend :metal: 

![image](https://user-images.githubusercontent.com/8518514/126011268-2e87efab-9be0-4d43-8f41-dae1b1c39a6d.png)

PR #622, opened 2021-07-17
Test certificate generator doesn't create a certificate chain

christopher-henderson:
This PR addresses #620 wherein @robplee identified that certs were incorrectly passed their own private key as the signing key, resulting in a "chain" of self signed certificates.

I also went ahead and knocked out some other issues such as the incorrectly setting `BasicConstraintsValid = false` for signing certificates and assigned broad validity dates.


v3/cmd/genTestCerts/genTestCerts.go
```diff
@@ -334,3 +339,9 @@ var nextSerial = func() func() *big.Int {
 //		}
 //		return serial
 //	}
+
+type Certificate struct {
```
robplee, line 343: This might be entirely personal preference/the conventions I'm used to but when I was looking for where this was defined I was expecting it at the top under the imports rather than at the bottom.

Happy to concede this point if I'm the only person who cares.

robplee:
Apart from my minor formatting request it looks good to me.

v3/cmd/genTestCerts/genTestCerts.go
christopher-henderson: So my approach to this is based on a mix of locality and how "interesting" a definition is.

For example, if a packages structs are the main show for external consumers then I will order it as

```
struct A def
impl A def
  ...
  ...
  ...
struct B def
impl B def
  ...
  ...
  ...
```

If a type is _very_ specific to a local use, then I feel fine keep its definition local to that use (even if its smack dab in the middle of the file)

```
func 
func 
func
struct def
func that uses the above def as a one-off
func
func
```

Now, to me, this struct def falls under the "just not very interesting" category. It's a way to shuttle around a parsed certificate alongside with its original key. It's more of an oddity then a feature of interest (at least to me).

Issue #625, opened 2021-08-12
e_subject_common_name_not_from_san is no longer sufficient for enforcing CABF BRs


robplee:
As of the 21st August 2021, the e_subject_common_name_not_from_san lint is not strict enough according to the update that'll be merged into the BRs on that date (I'm not 100% sure how the CABF works yet so my use of terms may be wrong, sorry!).  The new text makes it clear that if the common name contains either a wildcard domain name or FQDN then it must appear _exactly_ in the SAN dNSNames field as it does in the common name.  The specific text is as follows:

> If the [subject:commonName] is a Fully-Qualified Domain Name or Wildcard Domain Name, then the value MUST be encoded as a character-for-character copy of the `dNSName` entry value from the `subjectAltName` extension. Specifically, all Domain Labels of the Fully-Qualified Domain Name or FQDN portion of the Wildcard Domain Name must be encoded as LDH Labels, and P-Labels MUST NOT be converted to their Unicode representation.

I started a comment thread on the appropriate CABF PR (linked in [1]) and @sleevi joined in and made the point that despite dns names being case insensitive, the text says that the common name must appear as a character-for-character copy which leaves little room for interpretation about case sensitivity.  Furthermore, the google root program will be considering certs with common name "example.com" and dNSName "EXAMPLE.COM" to be misissuances.

Therefore I guess we need to make the e_subject_common_name_not_from_san lint ineffective as of the 21st and create a new lint to perform a more stringent check.

Happy to do the work myself and shouldn't have any trouble opening a PR soon.

[1] https://github.com/cabforum/servercert/pull/285/files#r683444000

robplee:
Closing this now following the merging of #627 which enforces the newly refined text in CABF BRs v1.8.0

PR #627, opened 2021-08-23
e_subject_common_name_not_from_san is no longer sufficient for enforcing CABF BRs

robplee:
This is my solution to the issue I raised #625.  I've gone with a straightforward fix by making the old lint ineffective from the 21st August 2021 and making a new lint which performs a case-sensitive string comparison between the common name and SAN dNSNames.  I have also refactored the tests on the original lint to use the de facto standard table approach because it allowed me to largely copy and paste the new tests I wrote for the new lint across with just some minor tweaks.

Comments for reviewers:
1. I'm not super fond of the name for the new lint so I'm open to suggestions as I really struggled to come up with something given that lint_subject_common_name_not_from_san is already taken and I can't just use a w_/e_ prefix as both old and new lint return errors!  Maybe a suffix of '_case_sensitive' would work better?
2. ~I'm not familiar with all the lingo surrounding the CA/B forum so if someone can let me know what I should have called the CABFBRs_SC48_EffectiveDate variable then that'd be great!~ Thanks @sleevi!


v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
```diff
@@ -0,0 +1,69 @@
...
+		Lint:          NewSubjectCommonNameNotExactlyFromSAN,
+	})
+}
+
+func NewSubjectCommonNameNotExactlyFromSAN() lint.LintInterface {
+	return &subjectCommonNameNotExactlyFromSAN{}
+}
+
+func (l *subjectCommonNameNotExactlyFromSAN) CheckApplies(c *x509.Certificate) bool {
+	return c.Subject.CommonName != "" && !util.IsCACert(c)
```
sleevi, line 50: STYLE: Does Golang have any idioms re: `len(c.Subject.CommonName) != 0` vs `c.Subject.CommonName != ""`
STYLE: Should we place the bool check first? :)

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
```diff
@@ -0,0 +1,69 @@
...
+	cn := c.Subject.CommonName
+
+	for _, dn := range c.DNSNames {
+		if cn == dn {
+			return &lint.LintResult{Status: lint.Pass}
+		}
+	}
+
+	for _, ip := range c.IPAddresses {
+		if cn == ip.String() {
```
sleevi, line 63: 1. Doesn't Golang support `"<nil>"` as a value here? Should that be accounted for?
2. For IPv6, there's obviously several representations, and I missed that SC48 didn't put a normative dependency on to RFC 5952, specifically https://datatracker.ietf.org/doc/html/rfc5952#section-4 . Implicitly, I think this logically holds, but that's not explicitly stated anywhere. At the risk of significantly overthinking things, do you think we should add an extra pass that, if there are no exact matches (which pass), we do a fuzzy match for IPv6 representations (I guess attempting to parse the CN to an IP, then seeing if that exactly matches any of the raw/non-stringified values), and warn if it's a fuzzy match (aka a different string encoding)

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san_test.go
```diff
@@ -0,0 +1,61 @@
...
+		{
+			name:           "Error - common name not in SAN.DNSNames",
+			inputFile:      "SANWithoutCNSeptember2021.pem",
+			expectedOutput: lint.Error,
+		},
+		{
+			name:           "Error - common name in SAN.DNSNames but case mismatch",
+			inputFile:      "SANCaseNotMatchingCNSeptember2021.pem",
+			expectedOutput: lint.Error,
+		},
```
sleevi, line 45: Some IPv4 / IPv6 tests would be good, right?

Potentially testing some of the non-standard Ipv4 encodings (e.g. `"0x7F000001"` for a commonName, or `"[::00:01]"`)

How does Golang's stringifier handle this? Does it bracket encode as in URL encoding, or does it omit the brackets? I'm guessing it omits them. Here, again, the BRs and relevant specs are (I believe) a bit silent on what to expect, and I legitimately have no clue what systems do.

To be clear: I'm think I'm fine taking a conservative/common-sense approach here, and letting CAs raise concerns if it ends up overly restrictive (and clarifying the BRs as appropriate). I filed https://github.com/cabforum/servercert/issues/301 to track this

v3/util/time.go
```diff
@@ -63,6 +63,7 @@ var (
 	CABFBRs_1_6_9_Date          = time.Date(2020, time.March, 27, 0, 0, 0, 0, time.UTC)
 	CABFBRs_1_7_1_Date          = time.Date(2020, time.August, 20, 0, 0, 0, 0, time.UTC)
 	AppleReducedLifetimeDate    = time.Date(2020, time.September, 1, 0, 0, 0, 0, time.UTC)
+	CABFBRs_SC48_EffectiveDate  = time.Date(2021, time.August, 21, 0, 0, 0, 0, time.UTC) //TODO: get the right name for this
```
sleevi, line 66: It should go out via e-mail today that this is CABFBRs_1_8_0_Date - 1.7.9 was SC47, and SC48 finished review on Saturday

sleevi (changes requested):

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san_test.go
sleevi: Ah, @CBonnell helpfully pointed out on the issue that I'm an idiot who can't read (granted, he didn't say it, but it's an accurate self-assessment :P )

So I think the actionable thing here is still just to make sure we have tests covering the cases mentioned in https://github.com/cabforum/servercert/pull/285/files#diff-e0ac1bd190515a4f2ec09139d395ef6a8c7e9e5b612957c1f5a2dea80c6a6cfeR2092

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
sleevi: @CBonnell rightfully pointed out I can't read - and https://github.com/cabforum/servercert/pull/285/files#diff-e0ac1bd190515a4f2ec09139d395ef6a8c7e9e5b612957c1f5a2dea80c6a6cfeR2092 is the requirement

That said, can you confirm Golang's ip.String() follows that RFC?

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
robplee: So, c.Subject.CommonName is a string type which in golang has the nil value: "".  So, we could change the check to check the length but I don't think it would really make any difference in terms of function and personally I think it would be a negative on the style front.  I'll happily bow if a deciding vote is cast for len() != 0

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
robplee: So, according to the comment on the IP.String() function:
```
// String returns the string form of the IP address ip.
// It returns one of 4 forms:
//   - "<nil>", if ip has length 0
//   - dotted decimal ("192.0.2.1"), if ip is an IPv4 or IP4-mapped IPv6 address
//   - IPv6 ("2001:db8::1"), if ip is a valid IPv6 address
//   - the hexadecimal form of ip, without punctuation, if no other cases apply
```
So I think it's probably just the latter case we want to think about catching?

v3/util/time.go
robplee: That would explain why I couldn't find out what term I should be using if it hadn't been said anywhere yet.  Thanks!

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san_test.go
robplee: I believe I've now covered the test cases for correctly formatted CN/SAN IP addresses on IPv4 and v6.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
robplee: I've been thinking and I'm inclined to think that if we want to check that the SAN:IPAddresses are valid we should be doing that using a separate lint to split the CN-in-SAN check from the IPAddresses validity check.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
sleevi: StackOverflow points to https://groups.google.com/g/golang-nuts/c/7Ks1iq2s7FA saying both are accepted, so no opinion here :) Obviously influenced by C++ here (and `std::string::size` vs `operator==`)

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
sleevi: Right, the issue here isn't whether the SAN:IPAddresses are valid - those are stored in an unambiguous network bytes. The issue is the stringified form here. What I wanted to make sure is that the stringification function we're using (here, `IP.String()`) matches the rules set out in SC48.

The problem is that there are edge cases here:
- For not v4/v6 addresses, it prepends with a `?` (the documentation doesn't mention this, but the implementation does do this)
  - This isn't so much a concern, because SAN will always be v4 or v6, and that's handled by the X509 parser
- For IPv6, it appears to correctly implement `4.1` - `4.3` of https://datatracker.ietf.org/doc/html/rfc5952#section-4
- However, it does not correctly handle Section 5 - for example, an input of "::ffff:192.0.2.1" returns an output of "192.0.2.1" - https://play.golang.org/p/T7Cd-mWqZ3R
  - To be fair, I'm not sure the expected result here for IPv4-mapped addresses (from the ballot), but maybe this is what you meant by a separate lint to tackle these (e.g. should a v4 mapped address be stored *as* a v4 SAN?)

CBonnell: SC48 clarified that IPv4-mapped addresses are considered to be Reserved IP addresses and should not be present in Certificates (the definition of Reserved IP address now states that all address blocks listed in this IANA registry here are considered Reserved: https://www.iana.org/assignments/iana-ipv6-special-registry/iana-ipv6-special-registry.xhtml). IPv4-mapped addresses are entry #3 in the registry. Given this, I don't think we need to be concerned with @sleevi's last corner case.

robplee: Thanks for the reply @CBonnell.  For the record, the reserved IP ranges are checked by lint "e_subject_contains_reserved_ip" so I believe we can consider this resolved.

sleevi: I will defer further reviews until after I have both coffee in hand and have fully reviewed all of SC48, and not just looking at implementation edge cases with stringifying functions ;)

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san_test.go
```diff
@@ -0,0 +1,81 @@
...
+			name:           "Error - common name not in SAN.IPAddresses, IPv6",
+			inputFile:      "SANIPv6AddressNotMatchingCommonName.pem",
+			expectedOutput: lint.Error,
+		},
+		{
+			name:           "NE - certificate issued before 21 August 2021",
+			inputFile:      "SANWithMissingCN.pem",
+			expectedOutput: lint.NE,
+		},
+	}
```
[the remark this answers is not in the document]
robplee: Silly question, are both of those considered to be valid or is the latter invalid?  I've got test certs for both they are both accepted which I assume should be fine as in the latter case both IP addresses are equivalent but I wanted to check given the discussion on the BRs that kicked all this off about things being functionally equivalent but not exact copies.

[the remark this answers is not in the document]
robplee: I'm going to answer my own question and say, "this probably needs work" as...
This: "2001:db8:0:0:1:0:0:1 in the SAN, and 2001:db8::1:0:0:1 in the CN" passes whereas
this: "2001:db8::1:0:0:1 in the SAN, and 2001:db8:0:0:1:0:0:1 in the CN" does not.
Clearly, if you put the minimal version in the CN it works because the IP.String() function abbreviates the first 0:0 pair in the SAN version as per RFC but if it's the other way round then the SAN will remain as the minimal version and the '==' check will fail as the CN is not the correct representation.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
```diff
@@ -0,0 +1,69 @@
...
+		Lint:          NewSubjectCommonNameNotExactlyFromSAN,
+	})
+}
+
+func NewSubjectCommonNameNotExactlyFromSAN() lint.LintInterface {
+	return &subjectCommonNameNotExactlyFromSAN{}
+}
+
+func (l *subjectCommonNameNotExactlyFromSAN) CheckApplies(c *x509.Certificate) bool {
+	return c.Subject.CommonName != "" && !util.IsCACert(c)
```
sleevi, line 50: Design Question: Should we instead be checking if `len(c.Subject.CommonNames) > 0` here?

Context:
1. What's the expected behaviour if I supply an empty commonName (a zero-length string as the value, but a CN attribute present in the DN). My reading is that this will return False, and the Lint will succeed. We have `cabf_br/lint_subject_common_name_included` (as a notice for using the deprecated field), but AFAICT, no other lint will fire here to signal that the empty string is an error.
2. In the `Execute` call, my understanding is that we'll only test the 'last' commonName. We have `cabf_br/lint_extra_subject_common_names` to test if there are multiple CNs, so this is more of a design question about whether we should run this check on every commonName present or not.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san_test.go
```diff
@@ -0,0 +1,97 @@
...
+			expectedOutput: lint.Pass,
+		},
+		{
+			name:           "Error - IPv6 choice in abbreviation",
+			inputFile:      "SANIPv6AddressChoiceInAbbreviation.pem",
+			expectedOutput: lint.Pass,
+		},
+		{
+			name:           "Error - common name not in SAN.DNSNames",
+			inputFile:      "SANWithoutCNSeptember2021.pem",
```
sleevi, line 58: This should be CNWithoutSAN right? (The name tripped me up, because the CN is present, but the SAN isn't)

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
sleevi: @CBonnell Thanks for https://github.com/zmap/zlint/pull/627/files#r695901667 - but that's not presently tested for in any lint, AFAICT. https://github.com/zmap/zlint/blob/d5ab97e9ca1cef2c7b594672c0265cef3b703637/v3/util/ip.go#L90-L114 contains the list of reserved addresses, which is what the `lint_ext_san_contains_reserved_ip` uses to implement its test.

Looking at the test, neither Golang's `ip.IsGlobalUnicast` nor `reservedNetworks` covers the IPv4-mapped IPv6 addresses. Presumably, this is because Golang internally tests `ip.To4()` when performing a number of these checks, while the `util/ip.go` does not.

Originally I was just going to suggest that we add a comment mentioning that IPv4-mapped IPv6 addresses are already accounted for via `subject_contains_reserved_ip`, but at present, that's not the case.

@robplee I've had my coffee (and then some), but I'm still not sure I've fully traced through all this path. You could certainly empirically test, but I _suspect_ we may need to add `::ffff:0:0/96` to the reserved networks. This isn't quite precise from an "IP protocol" layer (since it's really treated as a v4 address smuggled in a v6 length), but for purposes of zlint, it's totally correct, because we shouldn't be smuggling such mapped addresses as IPv6 SANs. If you do that, then perhaps adding a comment here noting the edge case of `IP.ToString()`

```suggestion
		// IPv4-mapped IPv6 addresses (e.g. "::ffff:0:0:192.168.0.1") are converted to their IPv4 equivalent
		// ("192.168.0.1") by IP.String(). This is not handled here, as it's treated as a reserved addresses in
		// util.IsIANAReserved, and will be rejected by cabf_br/lint_ext_san_contains_reserved_ip.
		if cn == IP.String() {
```

sleevi (changes requested):
Hey @robplee , sorry for the delay! Hopefully got it right this time :)

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
robplee: 1. If you supply an empty common name that is equivalent to not providing one at all.  In the zcrypto/x509/Certificate struct the CommonName is a string type which in go doesn't have a nil value.  The closest thing to one is the empty string, if it were a *string then it could be nil but as it isn't then a certificate with no common name or a zero length common name would result in the same thing as far as this lint is concerned.
2. That is correct, this lint checks only the commonName field which I believe does check only the final commonName provided.  I wasn't aware that it was possible for a certificate to have more than one to be honest but if it is deemed to be valid behaviour it's an easy change to make sure they're all present.

sleevi: 1. Correct, that’s why checking `len(c.Subject.CommonNames) > 0` allows you to distinguish absent-vs-empty
2. Yeah, it’s valid in X.509, invalid in the BRs. While we have another lint that checks there is only one CN, and even though the same section (of SC48) makes it clear there’s one CN, I _think_ the approach has been to treat the lints independently - i.e. check all CNs here.

robplee: Unless, I've missed your meaning @sleevi?  When you say actively providing an empty CN then would that be considered valid or not?  I guess one could say that if there is nothing in the CN field then there should be a SAN field with nothing in it to match but that would violate the requirement for a SAN dNSName to contain either a FQDN or a Wildcard domain so this does feel like it's getting a bit academic...

I guess the thing to add is to sort of repeat my earlier point and say that with zcrypto/x509 as it is now then it may not be possible to tell the difference between a parsed certificate with no CN and one with an empty CN when it comes to looking at them both in zlint.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san_test.go
robplee: Remember when I said I'd completely confused myself with this PR...  Yeah, that!  Thanks for the catch though :)

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
sleevi: @robplee Supplying an empty CN is invalid, and this would be the one lint responsible for detecting that.

I’m not sure the point your making about ZCrypto - I’m saying it is possible. Note I’m saying `CommonNames` - plural. There’s the singular field that just contains the last value, and then the plural field that is a slice that contains all values. If the slice has any value, then we know there was at least one CN present.

CBonnell: Just wanted to point out that we might get empty CN attribute value checking “for free” by the ASN.1 decode logic, as the ASN.1 CN attribute type has a constraint of least one character.

robplee: Apologies, I replied to myself without seeing your reply.

I will make some test data and see what happens but I think it'll behave as we expect.  From looking at the zcrypto code I think it _should_ keep adding empty common names into the array of strings so I think this is the right approach for catching this issue.  I'll make some test data to confirm.

However, I think it might be better for empty CNs to get their own lint as currently zlint will return an error declaring "The common name field in subscriber certificates must include only names from the SAN extension" and saying the e_subject_common_name_not_exactly_from_san lint has failed if there is an empty CN which doesn't seem quite right!

One quick question: what would the effective date be of such a lint?  BR 1.8.0 date or earlier?

sleevi: @CBonnell Zlint doesn’t do a lot of the ASN.1 constraint checks. AFAICT, there’s no lint today for empty CN.

@robplee An empty (zero-length) CN should be seen as “a value”; that is, it is semantically distinct than omitting CN. As @CBonnell points out, it’s invalid ASN.1, but it’s something easy to address in this lint.

That is, you get this “for free” if you just check all of the CNs in `c.Subject.CommonNames`, which is why I was flagging it. That’ll handle the empty/invalid ASN.1 case (at least, to the extent there’s not an empty SAN). So just changing this to check all the names will address this - sorry for not making that clearer.

robplee: After battling to get some test cases to confirm I've extended the lint such that it applies based on the len(CommonNames) and that does give us empty CN catching for free so I think we're slowly getting this PR closer to where we want it.

robplee:
I don't know how to comment on this so I'm going to have to abuse the format a bit:

https://github.com/zmap/zlint/pull/627#discussion_r699463497

I think this is the only open question remaining and unless you've had time to look further and determine strict confirmation either way then I can add the changes you recommend of putting "::ffff:0:0/96" into the list of reserved networks and add the comment text you proposed in also and we should be covered against those concerns.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
robplee: I think this is the only open question remaining and unless you've had time to look further and determine strict confirmation either way then I can add the changes you recommend of putting "::ffff:0:0/96" into the list of reserved networks and add the comment text you proposed in also and we should be covered against those concerns.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
```diff
@@ -0,0 +1,80 @@
...
+			if cn == ip.String() {
+				cnFound = true
+				break
+			}
+		}
+		if cnFound {
+			continue
+		}
+
+		return &lint.LintResult{Status: lint.Error}
```
sleevi, line 76: Do you want to update the error message to include which name was missing?

sleevi (approved):
This looks good to me, modulo one suggestion below. I wasn’t sure if @christopher-henderson or @zakird wanted to check for Go idioms, or if @CBonnell had any feedback.

I suspect we can address the IPv4-mapped addresses as a separate follow-up. While that may cause false negatives with this lint, it’s arguably moreso an issue for the reserved address logic, and we can deal with it there. It may just merit a comment regarding the subtlety of `IP.ToString()`, but that’s not a blocker.

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
zakird: I agree that makes sense.

christopher-henderson, line 76:
```suggestion
		return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("Missing common name, '%s'", cn)}
```

v3/lints/cabf_br/lint_subject_common_name_not_exactly_from_san.go
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import (
+	"github.com/zmap/zcrypto/x509"
```
christopher-henderson, line 18:
```suggestion
	"fmt"

	"github.com/zmap/zcrypto/x509"
```

christopher-henderson (approved):
No particular desires WRT to Go idioms. There _are_ labels and a `goto` that one can use for this sort of internal break logic, but it's not like that's a particularly good solution itself.

robplee:
Looks like you guys were busy yesterday!  Thanks everyone for all the reviews, comments and for the merge

PR #635, opened 2021-10-04
Lint for Non-XN Reserved Labels

CBonnell:
SC48 added a requirement that effective October 1st, all Domain Labels within FQDNs MUST consist solely of either P-labels (XN-Labels with valid Punycode output) or Non-Reserved LDH Labels.

This PR adds a lint for checking adherence to this requirement.

### DIFF
--- a/v3/lints/cabf_br/lint_dnsname_contains_prohibited_reserved_label.go
+++ b/v3/lints/cabf_br/lint_dnsname_contains_prohibited_reserved_label.go
@@ -1,0 +1,69 @@
+/*
+ * ZLint Copyright 2021 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package cabf_br
+
+import (
+	"regexp"
+	"strings"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_dnsname_contains_prohibited_reserved_label",
+		Description:   "FQDNs MUST consist solely of Domain Labels that are P‐Labels or Non‐Reserved LDH Labels",
+		Citation:      "BRs: 7.1.4.2.1",
+		Source:        lint.CABFBaselineRequirements,
+		EffectiveDate: util.NoReservedDomainLabelsDate,
+		Lint:          NewDNSNameContainsProhibitedReservedLabel,
+	})
+}
+
+type DNSNameContainsProhibitedReservedLabel struct {
+	CompiledExpression *regexp.Regexp
+}
+
+func NewDNSNameContainsProhibitedReservedLabel() lint.LintInterface {
+	return &DNSNameContainsProhibitedReservedLabel{
+		CompiledExpression: regexp.MustCompile(`^(?P<AceId>.{2})--.*$`),
+	}
+}
+
+func (l *DNSNameContainsProhibitedReservedLabel) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.DNSNamesExist(c)
+}
+
+func (l *DNSNameContainsProhibitedReservedLabel) Execute(c *x509.Certificate) *lint.LintResult {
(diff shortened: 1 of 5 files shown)


v3/lints/cabf_br/lint_dnsname_contains_prohibited_reserved_label.go
```diff
@@ -0,0 +1,69 @@
...
+
+			if match != nil {
+				aceId := match[l.CompiledExpression.SubexpIndex("AceId")]
+
+				if strings.ToLower(aceId) != "xn" {
+					return &lint.LintResult{Status: lint.Error}
+				}
+			}
+		}
+	}
```
sleevi, line 66: So a few thoughts about this block:

1. Do we need the Regex here? I'm not sure on RegEx performance, and this may be totally fine, but it also seems like this might be as simple as a test like below. Not saying this is "must fix", just that the regex/capture group approach stood out as heavyweight for something like this.

```golang
if len(label) >= 4 && label[2] == '-' && label[3] == '-' {
  aceId := label[:2];
  if strings.ToLower(aceId) != "xn" {
     ...
   }
}
```

2. In other lints (e.g. [`e_name_constraint_not_fqdn`](https://github.com/zmap/zlint/blob/d5ab97e9ca1cef2c7b594672c0265cef3b703637/v3/lints/rfc/lint_name_constraint_not_fqdn.go)), we build up an error string of all of the errors. This will bail on the first error, but would it be useful to list all the issues here? I can go either way (as I clearly did, for better or worse, in [`e_san_dns_name_onion_invalid`](https://github.com/zmap/zlint/blob/d5ab97e9ca1cef2c7b594672c0265cef3b703637/v3/lints/cabf_br/lint_san_dns_name_onion_invalid.go)), but wanted to draw your attention do it.

v3/lints/cabf_br/lint_dnsname_contains_prohibited_reserved_label_test.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func TestDNSNameHasXNLabel(t *testing.T) {
+	inputPath := "dnsNameXNLabel.pem"
+	expected := lint.Pass
+	out := test.TestLint("e_dnsname_contains_prohibited_reserved_label", inputPath)
+	if out.Status != expected {
+		t.Errorf("%s: expected %s, got %s", inputPath, expected, out.Status)
+	}
+}
```
sleevi, line 40: More of a nit, but just noting we'd been doing a lot more table-based tests in the new lints ( e.g. https://github.com/zmap/zlint/blob/d5ab97e9ca1cef2c7b594672c0265cef3b703637/v3/lints/cabf_br/lint_san_dns_name_onion_invalid_test.go )

One of the reasons I ask here is the use of the RegEx makes me wonder how Golang would handle smuggled UTF-8 - would the Regex match the character or the rune? For example, if a string was `"{two byte character, -, -}"`, would RegEx thinks it fails (because the two byte UTF-8 character is only one rune), or would it match? This is a _very_ contrived example, because obviously `IA5String` is not supposed to convey these characters, and we'd hopefully get a decode failure far sooner, but uh, you know how messy it is with ASN.1 string types. I _believe_ the example alternative I gave might handle this better, but I'm also not sure.

That suggests adding another test might be useful, which would be a lot of copy/paste boiler plate, which is where the table approach comes in :)

sleevi (changes requested):
Flagging this as "Request Changes", but really it's trying to double check considerations.

v3/lints/cabf_br/lint_dnsname_contains_prohibited_reserved_label.go
CBonnell: 1. Actually, I originally wrote the `if` statement pretty much identically to how you have it written, but decided to use the regex as the use of 3 constants/"magic numbers" for the minimum string length and indexes looked a bit convoluted to me. I think the regex approach is cleaner even if it's less performant than the "length-and-index check" approach, but I'll change it if others think there's value.
2. I personally think the "fail fast" approach is simpler from an implementation standpoint that outweighs any additional information gleaned from a more exhaustive check. Likely whatever action taken upon seeing 1 failed domain label is the same action taken upon seeing multiple failed labels, so I'm not sure of the value for a more exhaustive check. That being said, happy to add the more exhaustive check/reporting if others agree that's the correct approach. If we want a convention in this regard moving forward, it should be documented in the `CONTRIBUTING.md` document (I read it prior to authoring this lint and didn't see any guidance there).

v3/lints/cabf_br/lint_dnsname_contains_prohibited_reserved_label_test.go
CBonnell: Looks like Go `regexp` package is Unicode-aware: https://play.golang.org/p/x1-e_ZqwM2S. I think using Regex is fine in this case given the `IA5String` encoding of `dNSNames` and this code will work unchanged if we apply to subject CNs, which might be (but hopefully not) multi-byte.

Given that there's currently only two test cases, I went with the current approach, which is in line with the `CONTRIBUTING.md` document guidance. Happy to switch to table-based, but the universal switch to table-based tests should probably be documented in that doc so new lint contributors do the right thing the first time.

Issue #504, opened 2020-11-25
How to implement LINTs for the upcoming SMIME BRGs?


RufusJWB:
Dear maintainers!

Do you have a preferred way, how to organize the lints for the upcoming SMIME BRGs? I would create a new folder `cabf_smime` within the `v2/lints' folder. Do you share this opinion?

And how do we want to handle the situation, that many of those lints will be duplicates of the `cabf_br` lints? Shall we copy those lints or do we want to somehow 'link' them?

cardonator:
What criteria would be make these lints applicable? Would they not run the cabf_br lints at all if they had the right EKU? This is a good question overall since new cert types (looking at you, Qualified certs) don't necessarily apply to cross-sectional EKU baseline requirements like BR (OV)->EVGs (EV).

sleevi:
So, as it stands right now, the SMCWG (S/MIME) is wholly independent of the SCWG (TLS); thus, the requirements are best described as a "fork" (with no common ancestor), even if they're identical. When you think about S/MIME and TLS being independent hierarchies, this is actually quite good and makes sense. The question of having some common "base" requirement is... complex and complicated and we shouldn't use this issue to debate it ;)

So as a practical level, I think the question is whether or not the ZLint structure should align with how the CA/B Forum is structuring things (naively, I'd say yes; which is to say, copy), and what to do if the CA/B Forum ever creates a third "base" document (in which case, if such a day ever comes about, we could consider adopting that).

In a world of purpose-segmented roots, this isn't so bad - you'd only ever run "one" of these suites per root. However, I suspect some CAs are still slow to modernize, and thus will want to run both TLS and S/MIME on the same PKI hierarchy. In that case, the question is, I think whether it makes sense to use a "common" lint for both TLS and S/MIME (with no restriction on CheckApplies), or whether it makes sense to use two lints (as I suggest above), with separate CheckApplies. Because they're separate documents, and increasingly likely to be subject to separate supervision and audit regimes, I'm more supportive of the "copy-and-duplicate" vs trying to share.

cardonator:
I think that makes a lot of sense. In this case, the common base document would be the RFCs, right? Would these separate linting hierarchies _imply_ compliance with RFCs? Is it something that a user would have to specify? Or should it be managed through the CheckApplies function within the Lint interface for each set of lints? I assume that would require a pretty extensive change across the corpus of existing BR lints.

sleevi:
I'm not sure I follow?

[...]

Issue #458, opened 2020-07-24
Linting OCSP responses


jsha:
In [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1588001), Apple shared a [very thorough list of lints they apply to OCSP responses](https://bug1588001.bmoattachments.org/attachment.cgi?id=9160540) from their own issuance infrastructure. It would be useful to implement these in an open-source linter. Are the folks here interested in incorporating an OCSP linter into zlint? Do you think it would be better as a separate project?

zakird:
This came up once in the past #382. I think the consensus is: Yes! We'd love to have them, but we should structure their addition that they don't make running or maintaining existing certificate lints more difficult.

cardonator:
I think this fits into our sets of lints nicely, we could even exclude OCSP/CRL lints by default.

Legrandin:
+1 for being able to linting CRLs too

kowshikRoy:
Hi Maintainers,
I would be happy to work on implementing OCSP linting. The question is: 
- Given that CAB Forum has made OCSP optional in v2.0.1 (SC63)
Please let me know. We're planning to implement something internally which we can think of contribute to zlint.
Let me know if there is interest?

PR #646, opened 2021-10-29
Add lint to detect Domain Labels that violate LDH-Label syntax

CBonnell:
While there are several lints that checks for the syntactic correctness of domain labels, there appears to be no comprehensive check to ensure that all labels of a FQDN are LDH Labels.

While some of the checks in this lint are duplicative of other lints (label too long, etc.), I think it's best to have a single lint that comprehensively checks for all issues related to LDH Labels/preferred name syntax.

As a future step, the now-redundant lints could be removed.


v3/lints/cabf_br/lint_fqdn_contains_non_ldh_label.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_fqdn_contains_non_ldh_label",
+		Description:   "The FQDN contained in the dNSName MUST be composed entirely of LDH Labels.",
+		Citation:      "BR 7.1.4.2.1",
+		Source:        lint.CABFBaselineRequirements,
+		EffectiveDate: util.CABEffectiveDate,
```
sleevi, line 32: Is this correct?

That is:
- Should we be viewing this as an RFC 5280 requirement (specifically, https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6 "The name MUST be in the "preferred name syntax", as specified by Section 3.5 of [RFC1034] and as modified by Section 2.1 of [RFC1123].")
- AFAICT, the quoted text does not appear in any BR version. The specific quote is "The Fully‐Qualified Domain Name or the FQDN portion of the Wildcard Domain Name contained in the entry MUST be composed entirely of LDH Labels joined together by a U+002E FULL STOP (“.”) character"
- That language was introduced in https://cabforum.org/2021/07/22/ballot-sc48-v2-domain-name-and-ip-address-encoding/ as part of BRs 1.8.0
- How would/should this intersect with https://cabforum.org/2018/11/12/ballot-sc-12-sunset-of-underscores-in-dnsnames/ (which briefly created an exception for underscores, despite the RFC prohibition)

My suggestion would be this should be an RFC 5280 lint, and the description updated. I'm definitely supportive of this lint, though.

sleevi (changes requested):

v3/lints/cabf_br/lint_fqdn_contains_non_ldh_label.go
CBonnell: 1. The main concern with implementing this as an RFC lint is that 5280 allows for wildcards, the specification of which is beyond the scope of that document. We need to allow for Wildcard Domain Names (as defined in the BRs), but a strict interpretation of "preferred name syntax" will yield false positives with asterisk-only labels.
2. I modified the Description from the original BR text introduced in BR 1.8.0 to make it less verbose. If the convention is that we copy verbatim from the source text, I can pull in the full quote.
3. Yes, I know. I wrote it 😃 
4. This raises a good point. The simplest solution is to make the EffectiveDate the date BR v1.8.0 became effective, at the risk of missing some historical mis-issuance. We could make a more historically accurate check by reducing the severity (or suppressing altogether) those certificates issued under SC12, but that introduces a lot of implementation complexity (and therefore potential for bugs) that is of no benefit for issuance moving forward. Given this, I'm leaning towards changing the effective date to the effective date of v1.8.0.

Thoughts?

CBonnell: There are many integration test failures caused by redacted pre-certs (e.g., a SAN dNSName of "?.example.com") that were issued several years ago. This will be another source of "false" positives if the EffectiveDate is not amended to BR v1.8.0. Given this, I think the cleanest solution is to modify the EffectiveDate to avoid the morass of exceptions that have been historically allowed.

sleevi: 1. I'm not sure I understand the point about wildcards. You handle those here as well? Are you concerned about situations like `a*b.example.com`? RFC6125-bis is specifically addressing that as part of the updates. I don't anticipate any issues adding this as an RFC lint, but happy to hear from @zakird @cpu or @dadrian if there are.
2. I think eliding is fine, but perhaps indicate where that is (e.g. '...'). Mostly, the goal is to provide reference to the text, and that's why I raised it - it wasn't exactly obvious to me where this requirement was derived from.
3. Right :) My point was that it was new language, but tied with the old date.

As to 'redacted' precerts, they were considered misissued by some root programs, so it doesn't seem unreasonable to keep the effective date as when it was? That is, they're true positives, but in the past, and so I'm not sure the issues/concerns there?

Obviously, I'm generally going to argue for the extreme end of the spectrum, and be open to be talked down from that ;) I'm obviously a vocal believer that this was always an RFC requirement, so I'd be perfectly happy with the lint always being set. I'd be curious how the other maintainers feel about situations here where some root programs allowed/ignored the error due to potential CA confusion.

Possible options:
* LDH lint with the RFC effective dates (which goes to 2459)
* LDH lint with the RFC effective date, but moved to a LDH+U lint for the period between SC12/BR 1.6.2 and the SC12 sunset date (2019-05-01)
* LDH+U lint for the period of SC12 to sunset (2019-05-01) + LDH lint for 2019-05-01 and later
* As to the wildcard issue, one way would be for the regex to be LDH+W everywhere.
  * BRs since 1.0 have required in the "left most position", and so while 1.8.0 used better terminology, it seems there was consensus that was an existing clarification, and `lint_dnsname_wildcard_only_in_left_label` already catches this.

I don't think the effective date of 1.8.0 sounds right, if only because the years of discussion (and the explicit text) seem to address this.

CBonnell: 1. Yes, I handle the wildcard syntax, as defined in the BRs, in this lint in the cabf_br package. Unless you happen to know of a RFC that constrains wildcard syntax in the same manner as the BRs that we can cite in this lint, it is not appropriate to add this lint to the RFC lints. I don't believe RFC 6125-bis is an appropriate document for this purpose, as it is merely states best practices for client-side certificate processing and doesn't constrain certificate profiles. Even if it did, it would be very strange to cite RFC 6125-bis in this lint as RFC 6125-bis hasn't been published yet as an RFC despite your wanting to make the effective date of this lint January 1999 (the publication date of RFC 2459). Given all this, I believe the cabf_br lint suite is the best location for this lint.
2. I agree with you that quoting verbatim the relevant text has value; the user can then do a copy/paste to search in the relevant document to find a fuller description of the requirement. I updated the Description of this lint in the latest commit.

> As to 'redacted' precerts, they were considered misissued by some root programs, so it doesn't seem unreasonable to keep the effective date as when it was? That is, they're true positives, but in the past, and so I'm not sure the issues/concerns there?

My understanding is that the only public notification by any Root Program that redacted pre-certificates are mis-issuances was a singular post on the CT Policy mailing list in May 2016. Given that this was the policy decision of a single Root Program, it would be historically inaccurate to classify this prohibition as a BR prohibition shared across Root Programs.

But more generally, I'm not sure of the value of faithfully replicating historical policy changes that are no longer in force as new lints. I think a bit of historical inaccuracy (in terms of false positives or negatives) is fine as long as it decreases the amount of code/maintenance burden for the project. My opinion is that we should optimize for correctness of the current requirements while allowing for some historical inaccuracy. Otherwise, the authoring and maintenance of lints becomes increasingly onerous. Like @sleevi, I would be very interested to hear the maintainers' opinions on the best direction here.

sleevi: I'm definitely looking for feedback from other maintainers, because I do strongly believe this should be an RFC lint.

I'm not sure I understand the substance of your objection, given the possible solution outlined in https://github.com/zmap/zlint/pull/646#discussion_r739577965 with respect to modifying the regex.

That is, is there a disagreement that RFC 5280 requires preferred name syntax and acknowledges the possibility of wildcards?

christopher-henderson: `ACK` @sleevi I will dedicate some time to sit down and grok this conversation more completely as I'm not as familiar with this particular historical context as ya'll are.

robstradling: > My understanding is that the only public notification by any Root Program that redacted pre-certificates are mis-issuances was a singular post on the CT Policy mailing list in May 2016. Given that this was the policy decision of a single Root Program, it would be historically inaccurate to classify this prohibition as a BR prohibition shared across Root Programs.

@CBonnell You seem to be implying that "redacted pre-certificates" were not considered misissuances until a "public notification" by "a single Root Program" declared otherwise.  I disagree with that view.

"?" has never been a valid character in a domain name label; so I claim that "redacted pre-certificates" were always considered misissuances, and one Root Program went on record to confirm this.

If 6962-bis had retained the domain label redaction mechanisms that the TRANS WG experimented with, then it's conceivable that "?" might have become a valid character in domain name labels (i.e., if/when adoption of 6962-bis occurs).  But that didn't happen.

robstradling: @CBonnell The focus of that particular item in that particular CA Communication was to find out "if it is reasonable to expect that all newly issued TLS certificates from all CA certificates in the Mozilla program are being logged".  I don't think you can infer policy from the potential survey responses.

christopher-henderson: I admit that my role in this project is largely that of a code maintainer and that I lack the depth of historical context that others in this thread display. However, I hope that we might be able to resolve this with some code organization and triage.

The following are what I think are the (rough) ordering of importance is for this project in relation to its value to the rest of the world.

1. The proactive identification of misissuances today.
2. To serve as cross CA regression tests (see [LE misissuances](https://bugzilla.mozilla.org/show_bug.cgi?id=1735247) related to #635)
3. To provide historical context on the state of the web PKI.

Of course, that is rather coarse grain, but I believe it to be a rough ordering of importance. I believe that the first priority speaks to @CBonnell's notion that the project should `...optimize for correctness of the current requirements`.

The following are also what I think are some relevant musings on how to wrangle this codebase.

* Relatively new requirements should be included as lints that are as precise as possible and as soon as possible. This is because the newer the requirement is the more likely it is for a CA to not be fully up to spec (again, see #635 where Let's Encrypt tried to implement the requirement but missed a case).
* What is effectively the same exact lint is okay to duplicate across different categories. This project is more concerned with lints being precise, isolated, easy to verify through code review, and directly tied to a written requirement than it is with typical software engineering philosophies such as DRY. We try our best, but we find that if there is even a subtle way that a single lint should actually be two, then they should probably just be two lints. The world of the web PKI is enough of a tangled mess as is without further complicating things through overly clever code sharing mechanisms that are brittle and cause cascading changes at the slightest provocation.

To that end...

@sleevi would you be amenable to having this as both an RFC lint _and_ a CABF lint? Or are you arguing that this should not be CABF at all and that it only makes sense for RFC?

With regard to the underscores and pre-redactions, could we not cut these into tickets for triage into individual lints? It sounds as though they have different efficacy dates than what we are targeting here (and maybe different error/warning levels)?

sleevi: @christopher-henderson Effectively, the question of CABF_BR vs RFC feels mostly a relitigation of past discussion in CA/Browser Forum, which is why it feels a bit silly, but also why I'm looking for feedback from other maintainers, since it's the same discussion that's been hashed out in the CABF between browsers and CAs over and over again.

At issue is that the RFC prohibited something (names in the non-preferred name syntax), which several CAs ignored (to the effect of several thousands of certificates). The CAs violating argued that the RFC didn't prohibit it, and SC12 was the result - a temporary amnesty for the CAs that were misissuing, followed by language that reiterated the RFC's existing requirements in ways that were hopefully unambiguous.

The issue this lint checks is aligned with the RFC; the Baseline Requirements are simply reiterating it. The choice to place this lint in the BRs section is concerning, primarily because it implies this is not an RFC requirement, but one added and imposed, but that's not exactly accurate; the BRs are trying to explain the logical interactions with the following paragraph from [RFC 5280, Section 4.2.1.6](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6), namely:

```text
   When the subjectAltName extension contains a domain name system
   label, the domain name MUST be stored in the dNSName (an IA5String).
   The name MUST be in the "preferred name syntax", as specified by
   Section 3.5 of [RFC1034] and as modified by Section 2.1 of
   [RFC1123].  Note that while uppercase and lowercase letters are
   allowed in domain names, no significance is attached to the case.  In
   addition, while the string " " is a legal domain name, subjectAltName
   extensions with a dNSName of " " MUST NOT be used.  Finally, the use
   of the DNS representation for Internet mail addresses
   (subscriber.example.com instead of subscriber@example.com) MUST NOT
   be used; such identities are to be encoded as rfc822Name.  Rules for
   encoding internationalized domain names are specified in Section 7.2.
```

Here, "preferred name syntax" is the LDH rule implemented by this lint.

As far as I can tell, the core concerns are:
1. If this is placed as an RFC lint, then it would not accurately reflect the SC12 amnesty of the BRs. Namely, historic issuance during the SC12 amnesty period would show as RFC violating issuance. This is indeed correct, but I suppose the issue is a lack of ability to say such misissuance is "CABF accepted".

    This is similar to critical vs non-critical `nameConstraints`, namely [RFC 5280, Section 4.2.1.10](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.10) saying `Conforming CAs MUST mark this extension as critical and SHOULD NOT impose name constraints on the x400Address, ediPartyName, or registeredID name forms.`, while [Baseline Requirements, v1.8.0, 7.1.2.2 (f)](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.8.0.pdf#page=73) says `Non‐critical Name Constraints are an exception to RFC 5280 (4.2.1.10), however, they MAY be used until the Name Constraints extension is supported by Application Software Suppliers whose software is used by a substantial portion of Relying Parties worldwide.`
2. The Baseline Requirements permits wildcard certificates, while RFC 5280 says `Finally, the semantics of subject alternative names that include wildcard characters (e.g., as a placeholder for a set of names) are not addressed by this specification.  Applications with specific requirements MAY use such names, but they must define the semantics.`. This was suggested in https://github.com/zmap/zlint/pull/646#discussion_r740272210 as being at odds with moving it to the RFC, although from the cited text, it's clear the RFC permits such a character, simply with undefined syntax. This is why I said in https://github.com/zmap/zlint/pull/646#discussion_r739577965 that simply adding the wildcard (with undefined syntax) is arguably RFC permissable, and mentioned RFC 6125-bis as something trying to further provide IETF guidance.
3. Finally, the issue of "redacted" precertificates, in which the precertificate does not structurally match the issued certificate. This was based on early drafts of RFC 6962-bis; i.e. CT v2. It was never permitted by CTv1 (i.e. what browsers and ZLint implement). Specifically, it was introduced in [RFC 6962-bis, draft 01](https://datatracker.ietf.org/doc/draft-ietf-trans-rfc6962-bis/01/) and removed in [RFC 6962-bis, draft 19](https://datatracker.ietf.org/doc/draft-ietf-trans-rfc6962-bis/19/) (later split into [a different draft](https://datatracker.ietf.org/doc/html/draft-strad-trans-redaction-01)). This is only an issue [for a single CA](http://web.archive.org/web/20160922213310/http://www.symantec.com/connect/blogs/privacy-redaction-and-certificate-transparency), based on draft-14, although the one root program requiring CT [stated they would treat it as misissuance](https://groups.google.com/a/chromium.org/g/ct-policy/c/fCt4Bm03GsI/m/jBbqE_QWBQAJ). If this lint is adopted as an RFC lint, such precertificates would be flagged (because their final certificate cannot be recovered and linted), and that seems to be the objection to making it an RFC lint.

My take is that this belongs as an RFC lint, because it's **always** been an RFC requirement. The underscore issue is, conceptually, similar to non-critical `nameConstraints` during the brief amnesty window. The concern about wildcards is easily addressed within the (RFC) regex, while still allowing the BRs to lint the more proscriptive wildcard syntax. The redaction issue is, IMO, at best of only interest historically, and even then, an arguable violation (resulting from using a still-changing draft of CT v2, which ZLint doesn't support anyways).

As far as I can tell, the only issue that would prevent this from being an RFC lint is updating the regex to allow `*` and removing the "remove prepended wildcard" check. Removing the "remove prepended wildcard" check would not require introducing a BR-specific lint, because we already have lints to cover where and how the `*` appears.

christopher-henderson: Thank you for your patient (and thorough) explanation @sleevi!

The following is what my understanding is of what our path forward might be. We could have three lints...

1. An RFC lint that is effective as of 5280, that is an error, and encodes the allowance of a wildcard.
2. A CABF lint that is effective as of SC12, ineffective as of BR 1.8.0, is a warning, and does not allow for a wildcard.
3. A CABF lint that is effective as of BR 1.8.0, is an error, and does not allow for a wildcard.

I believe that having RFC pop up as an error with SC12 popping up as a warning under CABF (with a good description) may effectively convey the notion that it was CABF accepted for the time period of that amnesty, but that RFC was never truly okay with it.

If I understand correctly then I think the thread is in agreement that redacted precertificates are not of too much interest at this time.

CBonnell: To be clear, I'm not attempting to re-litigate previous discussions in the CA/B Forum here, especially since doing so here serves no purpose. I'm merely trying to provide the rationale for my design decisions for this lint.

I don't believe adding "*" to the RFC regex is an appropriate solution, because if a specification does not allow for that character to be used as a wildcard, then the RFC lint is being overly liberal in allowing it. Which brings me to a question about the overall goals of zlint: is zlint intended to be used as a linting framework that is applicable and appropriate for non-TLS certificates, or is it geared solely towards TLS issuance? If the former, then I think it's clear that modifying the RFC regex to allow the asterisk character is a mis-step as other specifications may not allow that character in names. If the latter, then I suppose for expediency's sake we include characters that violate preferred name syntax in the allowed set of characters in the RFC lint. My understanding is that it's the former, which is why I classified this lint as I did.

One idea I have to perhaps modify or create an RFC lint for preferred name syntax that takes a function that transforms/processes a name with the specification-specific semantics for wildcards and returns a name string that is to be checked for conformance with preferred name syntax. As an example, the CABF lint can pass a function to this lint that removes any leading wildcard label. As another example, a lint for redacted precerts would remove all leading "?" labels in the "wildcard handler" function prior to checking for preferred name syntax. I think such a solution cleanly separates the concerns while not shoehorning WebPKI TLS serverAuth-specific semantics in the RFC lints.

Thoughts?

sleevi: @CBonnell To your point about the RFC prohibiting it, namely
> if a specification does not allow for that character to be used as a wildcard, then the RFC lint is being overly liberal in allowing it.

I tried to address in https://github.com/zmap/zlint/pull/646#discussion_r744892571 the relevant text from the RFC, namely:
> Finally, the semantics of subject alternative names that include wildcard characters (e.g., as a placeholder for a set of names) are not addressed by this specification. Applications with specific requirements MAY use such names, but they must define the semantics.

Similarly, I don’t think your solution is good: redacted precerts are not a valid thing, and have never been. This really is solely about wildcards here, and in the context of a DNSName lint, then beyond RFC 3280/5280, you have RFC 2818 and 6125 making it clear they’re permitted and refining the syntax.

As to the meta-goals for ZLint, to date it has been focused on what is being actively maintained. This is why, for example, there are discussions of removing the ETSI lints absent more demonstrable maintenance. However, that largely seems irrelevant here, because it seems the crux of the argument is “the RFC forbids wildcards”, but it doesn’t.

CBonnell: > you have RFC 2818 and 6125 making it clear they’re permitted and refining the syntax

RFC 6125 makes it clear that the asterisk ("*") can appear within a label and does not have to be the sole character, despite this being disallowed by CABF rules. Additionally, the ABNF for "preferred name syntax" in RFC 1035 makes it clear that asterisks are not allowed: https://datatracker.ietf.org/doc/html/rfc1034#section-3.5. So I disagree that allowing asterisks in a generic RFC-level lint is an appropriate solution.

> As to the meta-goals for ZLint, to date it has been focused on what is being actively maintained.

There has been a desire to add support for linting other certificate types (#504) or even other document types (#458). Adding TLS-flavoring to RFC lints that are general across certificate types will complicate the implementation of such lints.

sleevi: What part of the MAY here mentioned directly in RFC 5280 is problematic here? 

I’m not sure how/why you’re arguing that the following language prohibits wildcards:
> Applications with specific requirements MAY use such names, but they must define the semantics.

sleevi (changes requested):
Just reflecting in GitHub don’t think we should land this as a BR lint, and it should be an RFC lint. I’m trying to keep an open mind in light of new information/perspective, but in the present form, think it should be changed.

v3/lints/cabf_br/lint_fqdn_contains_non_ldh_label.go
CBonnell: > Applications with specific requirements MAY use such names, but they must define the semantics.

This is exactly why I classified this lint as a CABF lint. The specification in question (CABF BRs) defines the semantics of wildcards. None of the RFCs define the semantics of wildcards that match that of the CABF BRs.

sleevi: I think, in the discussion, a few things are getting lost here, so let me try to restate:

## Wildcards

1. What characters does RFC 5280 allow within the `dNSName`?
  * Unambiguously, we know any characters in the "preferred name syntax" (a-z, A-Z, 0-9, -) are allowed.
  * We also can see that the RFC permits wildcards, but defers to other RFCs to define the syntax and semantics of where those wildcards can be placed.
2. In the context of TLS using `dNSNames`, as [RFC 6125, Appendix B](https://datatracker.ietf.org/doc/html/rfc6125#appendix-B) captures, several RFCs have defined a variety of syntaxes, which also supports the "Wildcard is allowed"

This lint is not a lint for "left-most label" wildcard syntax. We already have a lint for this. As noted by the name, this is merely a lint for LDH / "preferred name syntax", including the syntactical rules from preferred name syntax (e.g. prohibitions on `-` as a leading/trailing character).

This lint can fully be an RFC lint if `nonLDHCharacterRegex` is updated to permit `*` as a character, given that RFC 5280 permits such characters. This would also obviate the need to `util.RemovePrependedWildcard` - because the regex would be satisfied by such labels.

Existing lints for wildcard labels, within the BR section, will handle the BR compliance. There's no need to try to lint for that in this lint - and ample reason not to (conflating multiple requirements)

## Underscores

The question I raised in https://github.com/zmap/zlint/pull/646#discussion_r739451253 was related to the Effective Date of the lint. I believe we agree that the "preferred name syntax" is an existing RFC requirement since the first RFC requirement. Preferred name syntax prohibits underscores.

The CABF, briefly, had an exception to the RFC (SC12 adoption to sunset date). The question was whether RFC lints should be selectively enforced based on CABF requirements, for historic purposes, and it seems there's agreement that's of limited value. That is, underscores were prohibited, and presently are prohibited, so we don't need to worry about disabling the RFC lint because the CABF said it was OK.

If we apply this logic to `nameConstraints`, it would equally be possible to imagine there is a "name constraints MUST be critical" RFC lint, which treats it as an error, and require the CA to selectively ignore such lints (e.g. because the BRs permit), along with a BR lint "name constraints MAY be critical" (which could be sunset once the BRs remove that exception, which is doable within a few months)

## Question Marks

You raised a concern with making the LDH lint effective since the RFC start, as it would flag pre-certificates using redaction as misissued. However, this is correct as well - there was no support for redaction in CTv1 (which ZLint supports), and while there were several drafts for redaction in CTv2 using different approaches, they were just that - drafts. At issue here is that while SCTs are versioned, at the time, the precertificates stored in logs weren't unambiguously so, and so the signed data used for a RFC6962-bis-draft-14 log (i.e. Symantec's logs) could be replayed against a CTv1 log (i.e. every other log) and be treated as a signed certificate from the CA.

However, as with underscores, ZLint doesn't necessarily need to concern itself with that now-quite-old historic accident, which was equally misissuance at the time anyways. So this is moot.

## Changes Requested

To recap, I'm requesting you:
1. Change this from a BR lint to an RFC lint
2. Change the effective date to the RFC date
3. Change the regex to allow `*`
  * Which means you can safely remove the `util.RemovePrependedWildcard` without breaking any certs that would be accepted today.
  * Will ensure you don't cause errors on certs that are BR-errors not RFC-errors

vanbroup:
Should this lint also check for any reserved LDH Labels that are not P-Labels?

In relation to:
https://bugzilla.mozilla.org/show_bug.cgi?id=1740493

CBonnell:
> Should this lint also check for any reserved LDH Labels that are not P-Labels?
> 
> In relation to: https://bugzilla.mozilla.org/show_bug.cgi?id=1740493

Hi Paul, that lint was implemented as part of #635.

PR #648, opened 2021-11-28
Support for Configurable Lints

### LINKED ISSUE #624: RFC for Configurable Lints
# Purpose

Today, the only facts that are provided to ZLint are those that are present within the target x509 certificate. That is to say, the only input to the linter is the certificate itself. This works fairly well, in general, due to the fact that most requirements refer to data encoded within the certificate itself.

This is not, however, universally true. A lint may require access to the full certificate chain in order to deterministically come to a conclusion ([#491, Ocsp eku check for non tls certificates](https://github.com/zmap/zlint/pull/491)). Or language may specify that the requirement only takes effect if certain conditions regarding the certificate's cross signer are true ([#619, Lint that the EKU values within a SubCA certificate comply with CABF_BR 7.1.2.2 letter g)](https://github.com/zmap/zlint/pull/619)). And ETSI lints have had a history of being inaccurate due to requiring outside  data ([#581, Bug: ESI lints are not consistent with ETSI EN 319 412-5 requirements regarding non-qualified certificates](https://github.com/zmap/zlint/issues/581)).

If the certificate in isolation is in its self insufficient for the accurate execution of a given lint, then it seems that additional context must be shuttled to the lint in some way.

# Proposal

This proposal is to introduce a second, optional, input file into the ZLint command line interface which represents a configuration/context for that individual run. A supporting framework is also proposed for lints to declare an interest in arbitrary configurations.

### Goals
* Non-breaking change. All current invocations of ZLint, as well as all extant lints, MUST remain valid.
* Operators SHOULD be able to easy discover, and generate, configurations relevant to the lints that they are targeting.
* Lint authors SHOULD NOT be required to add or edit any further files than they already do. That is, lints should stay localized to their individual files and tests.
* Lints that are configurable MUST have a testing framework available to them so that they may construct unit tests against any number of arbitrary configuration combinations.

# Code Interface
Lints may implement a new, optional, interface - `Configurable`...

```go
type Configurable interface {
    Configure() interface{}
}
```

...where the returned `interface{}` is a pointer to the target struct to deserialize your configuration into.

This struct may encode any arbitrary data that may be deserialized from [TOML](https://toml.io/en/). Examples may include:

* PEM encoded certificates or certificate chains
* File paths
* Resolvable DNS entries or URIs
* Dates or Unix timestamps

...and so on. How stable and/or appropriate a given configuration field is is left as a code review exercise on a per-lint basis.

If a lint is `Configurable` then a new step is injected at the beginning of its lifecycle.

---

---

# Example Lints

### Contrived Examples

The following is a set of complete examples showcasing how one might write lints to utilize this configuration framework. They are contrived and solve no real problem in-and-of themselves. They are merely demonstrative.

The following lint encodes its configuration within the `struct` that is the lint itself. The lint itself is rather silly in that it is asserting that the signing certificate authority must play certain games while on a road trip - information that is clearly not encoded in any certificate ever issued.

```go
package cabf_br

import (
	"github.com/zmap/zlint/v3/lint"
	"github.com/zmap/zlint/v3/util"
)


type LintItselfIsConfigurable struct {
	BottlesOfBeerOnTheWall int `comment:"This MUST be set to the number of bottles of beer that were on the signing CAs wall at the time of issuance."`
	ISpyWithMyLittleEye    struct {
		Descriptions []string `comment:"Any number of descriptions that are valid for the subject."`
		Subject      string
        // Go naming convention does not often match external naming conventions, so the ability
        // to alias the name used within the target configuration is indeed useful.
	} `toml:"i_spy_with_my_little_eye"`
}

func init() {
	lint.RegisterLint(&lint.Lint{
		Name:          "e_ca_too_few_beers",
		Description:   "CA Certificates MUST have at least 99 beers.",
		Citation:      "BRs: 7.1.4.3.1",
		EffectiveDate: util.CABV148Date,
		Lint:          NewLintItselfIsConfigurable,
	})
}

func NewLintItselfIsConfigurable() lint.LintInterface {
	// Go code encourages that the zero value for a struct be valid, however
	// if you wish for your Lint to have non-zero defaults then your constructor is
	// the place to do so. The value initialized in this constructor WILL appear
	// in the example configuration printed through the `-exampleConfig flag`.
	return &LintItselfIsConfigurable{ISpyWithMyLittleEye: struct {
		Descriptions []string `comment:"Any number of descriptions that are valid for the subject."`
		Subject      string
	}{Descriptions: []string{"larger than a bread box", "smaller than a barn"}, Subject: "A car"}}
}

// In this case, the struct that is the lint itself is the target for configuration.
func (l *LintItselfIsConfigurable) Configure() interface{} {
	return l
}

func (l *LintItselfIsConfigurable) CheckApplies(c *x509.Certificate) bool {
	return util.IsCACert(c)
}

func (l *LintItselfIsConfigurable) Execute(c *x509.Certificate) *lint.LintResult {
	if l.BottlesOfBeerOnTheWall < 99 {
	} else {
	}
}
```

The following lint takes on a different strategy of aggregating a configurable `struct` rather than being the target `struct` itself.

```go
package cabf_br

import (
	"github.com/zmap/zlint/v3/lint"
	"github.com/zmap/zlint/v3/util"
)


type LintEmbedsAConfiguration struct {
        // The configuration field is not exported, however it is
        // explicitly returned in the Configure method, so it will be
        // targeted for deserialization.
	configuration                        embeddedConfiguration
        // This field is exported, however it is not a member of
        // returned by Configure, thus effectively making it
        // private in the eyes of the configuration framework.
	SomeOtherFieldThatWeDontWantToExpose int
}

// This lints configuration has only one question - is the certificate intended for the web PKI?
type embeddedConfiguration struct {
	IsWebPKI bool `toml:"is_web_pki" comment:"Indicates that the certificate is intended for the Web PKI."`
}

func init() {
	lint.RegisterLint(&lint.Lint{
		Name:          "w_web_pki_cert",
		Description:   "CA Certificates SHOULD....something....about the web pki",
		Citation:      "BRs: 7.1.4.3.1",
		EffectiveDate: util.CABV148Date,
		Lint:          NewLintEmbedsAConfiguration,
	})
}

// A pointer to an embedded struct may be passed to the framework
// if the author does not wish to expose certain fields in their primary struct.
func (l *LintEmbedsAConfiguration) Configure() interface{} {
	return &l.configuration
}

func NewLintEmbedsAConfiguration() lint.LintInterface {
	return &LintEmbedsAConfiguration{configuration: embeddedConfiguration{}}
}

func (l *LintEmbedsAConfiguration) CheckApplies(c *x509.Certificate) bool {
	return util.IsCACert(c)
}

func (l *LintEmbedsAConfiguration) Execute(c *x509.Certificate) *lint.LintResult {
	if l.configuration.IsWebPKI {
		return &lint.LintResult{Status: lint.Warn}
	} else {
	}
}
```

Given the above lints, operators may then generate a sample configuration which contains the defaults for every lint that is configurable.

```bash
./zlint -exampleConfig
```

```toml
[Apple]

[CABF_BR]

[CABF_EV]

[Community]

[ETSI_ESI]

[Mozilla]

[RFC5280]

[RFC5480]

[RFC5891]

[e_ca_too_few_beers]
# This MUST be set to the number of bottles of beer that were on the signing CAs wall at the time of issuance.
BottlesOfBeerOnTheWall = 0

[e_ca_too_few_beers.i_spy_with_my_little_eye]
# Any number of descriptions that are valid for the subject.
Descriptions = ["larger than a bread box", "smaller than a barn"]
Subject = "A car"

[w_web_pki_cert]
# Indicates that the certificate is intended for the Web PKI.
is_web_pki = false
```

Note that comments declared in the provided `structs` ARE present within example configuration generated by for operators. As such, lint authors SHOULD provide comments describing the purposes of their fields as well as the impact that those fields have on their given lint.

<a name="higher_scoped_configurations"></a>
# Higher Scoped Configurations

Lints may embed within theselves either pointers or structs to the following definitions.

```go
type Global struct {}
type RFC5280Context struct{}
type RFC5480Context struct{}
type RFC5891Context struct{}
type CABFBaselineRequirementsContext struct {}
type CABFEVGuidelinesContext struct{}
type MozillaRootStorePolicyContext struct{}
type AppleRootStorePolicyContext struct{}
type CommunityContext struct{}
type EtsiEsiContext struct{}
```
> **_NOTE:_**  These structs are currently empty and the following are merely examples of the sort of data that may be stored in them. For the sake of maintainability, predictability, and usability by CAs these structs do _not_ allow for arbitrary key:value mappings at runtime and thus must be statically defined. New fields may be added to these structs on a case-by-case bases when there is a consensus that a particular configuration may be valuable to share among many lints of a particular class. 

Doing so allows for the reuse of common configurations among entire classes of lints.

```toml
some_global = true
some_other_global = "So long and thanks for all the fish"

[Apple]

[CABF_BR]

[CABF_EV]

[Community]

[ETSI_ESI]
is_web_pki = true

[Mozilla]

[RFC5280]

[RFC5480]

[RFC5891]

[e_must_be_good_weather]
was_sunny_during_signing_ceremony = true
```

```go
type CaresAboutWebPKI struct {
    Etsi lint.EtsiEsiContext
}

type CaresAboutETSIandGlobal struct {
    Etsi    lint.EtsiEsiContext
    Global  lint.Global
}

type MustBeGoodWeather struct {
    Etsi lint.EtsiEsiContext
    WasSunnyDuringSigningCeremony bool `toml="was_sunny_during_signing_ceremony" comment="It MUST be a nice day"`
}
```

Lints receive _copies_ of higher scoped configurations (even in the event of embedding a pointer) in order to avoid surprising mutations across many lints.

<a name="live_example"></a>
# Live Example

The following example implementation is derived from #619. It serves as real world, and immediate, usecase for these new configuration facilities. There may be, however, a more stable and/or correct implementation - this is merely serving as a motivating example.

```go
package cabf_br

import (
	"github.com/zmap/zlint/v3/lint"
	"github.com/zmap/zlint/v3/util"
)

func init() {
	lint.RegisterLint(&lint.Lint{
		Name:          "e_sub_ca_eku_incompatible_values",
		Description:   "Subordinate CA extkeyUsage: if serverAuth is present, then emailProtection, codeSigning, timeStamping, and anyExtendedKeyUsage MUST NOT be present.",
		Citation:      "BRs: 7.1.2.2",
		EffectiveDate: util.CABFBRs_1_7_1_Date,
		Lint:          NewSubCAEkuIncompatibleValues,
	})
}

type subCAEkuIncompatibleValues struct {
	CrossSignerPEM string `comment:"Section 7.1.2.2: For Cross Certificates that share a Subject Distinguished Name and Subject Public Key with a Root Certificate operated in accordance with these Requirements, this extension MAY be present. If present, this extension SHOULD NOT be marked critical. This extension MUST only contain usages for which the issuing CA has verified the Cross Certificate is authorized to assert. This extension MAY contain the anyExtendedKeyUsage [RFC5280] usage, if the Root Certificate(s) associated with this Cross Certificate are operated by the same organization as the issuing Root Certificate."`
}

func (l *subCAEkuIncompatibleValues) Configure() interface{} {
	return l
}

func NewSubCAEkuIncompatibleValues() lint.LintInterface {
	return &subCAEkuIncompatibleValues{}
}

func (l *subCAEkuIncompatibleValues) CheckApplies(c *x509.Certificate) bool {
	return util.IsSubCA(c) && util.IsExtInCert(c, util.EkuSynOid)
}

func (l *subCAEkuIncompatibleValues) Execute(c *x509.Certificate) *lint.LintResult {
	// Check if l.CrossSigner was provided.
	// Parse to cert.
	// Do appropriate checks against section 7.1.2.2
	...
}
```

This structure generates the following configuration.

```toml
[e_sub_ca_eku_incompatible_values]
# Section 7.1.2.2:  For Cross Certificates that share a Subject Distinguished Name and Subject Public Key with a Root Certificate operated in accordance with these Requirements, this extension MAY be present. If present, this extension SHOULD NOT be marked critical. This extension MUST only contain usages for which the issuing CA has verified the Cross Certificate is authorized to assert. This extension MAY contain the anyExtendedKeyUsage [RFC5280] usage, if the Root Certificate(s) associated with this Cross Certificate are operated by the same organization as the issuing Root Certificate.
CrossSignerPEM = ""
```

With the cross signer certificate filled in...
```toml
[e_sub_ca_eku_incompatible_values]
CrossSignerPEM = """
-----BEGIN CERTIFICATE-----
MIIBBDCBrKADAgECAgEBMAoGCCqGSM49BAMCMAAwIhgPMDAwMTAxMDEwMDAwMDBa
GA85OTk4MTEzMDAwMDAwMFowADBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABDaq
E7tJ1D0lKPlIYcMylZ47WsEAFypRlBVUni+Q8ZY4S+NPVHFDxpZ/Qzx8qnXmHvf6
8caqXJthylrSzH1HVGGjEzARMA8GA1UdEwEB/wQFMAMBAf8wCgYIKoZIzj0EAwID
RwAwRAIgJklSU+eEIOc+oUHa/ugiNsBpTsrrrJudXuZ41hayNQsCIEwxc3quig63
ZI7LhYni7wlyXJBiXUlQkuacCGdqkv+/
-----END CERTIFICATE-----
"""
```

# Example CLI Usages

Operators will be able to generate a configuration file which contains all available configurations with ZLint initialized to their defualt values.

### Retrieving an Example Configuration
```bash
$ ./zlint -exampleConfig > config.toml
# Aside from the global configurations, this example shows
# that ZLint has exactly one configurable lint, which has one
# string value (for which the empty string is the default value).
$ cat config.toml
[e_sub_ca_eku_incompatible_values]
CrossSignerPEM = ""
```

### Running Against a Configuration.

Running against a configuration is as simple as pointing the `-config` flag to a file path.

```bash
$ cat >> config.toml << EOF
[e_sub_ca_eku_incompatible_values]
CrossSignerPEM = """
-----BEGIN CERTIFICATE-----
MIIBBDCBrKADAgECAgEBMAoGCCqGSM49BAMCMAAwIhgPMDAwMTAxMDEwMDAwMDBa
GA85OTk4MTEzMDAwMDAwMFowADBZMBMGByqGSM49AgEGCCqGSM49AwEHA0IABDaq
E7tJ1D0lKPlIYcMylZ47WsEAFypRlBVUni+Q8ZY4S+NPVHFDxpZ/Qzx8qnXmHvf6
8caqXJthylrSzH1HVGGjEzARMA8GA1UdEwEB/wQFMAMBAf8wCgYIKoZIzj0EAwID
RwAwRAIgJklSU+eEIOc+oUHa/ugiNsBpTsrrrJudXuZ41hayNQsCIEwxc3quig63
ZI7LhYni7wlyXJBiXUlQkuacCGdqkv+/
-----END CERTIFICATE-----
"""
EOF
$ ./zlint -config config.toml /path/to/cert/to/lint
```

# Testing

The linting `test` package provides the convenience function `test.TestLint(lintName string, testCertFilename string)`.

```go
func TestCaCommonNameNotMissing(t *testing.T) {
	inputPath := "caCommonNameNotMissing.pem"
	expected := lint.Pass
	out := test.TestLint("e_ca_common_name_missing", inputPath)
	if out.Status != expected {
	}
}
```

In addition to the above, a new facility shall be added that allows for the execution of a test against an arbitrary TOML string - `test.TestLintWithCtx(lintName string, testCertFilename string, context string)`.

```go
func TestCaCommonNameNotMissing22(t *testing.T) {
	inputPath := "caCommonNameNotMissing.pem"
	expected := lint.Pass
	ctx := `
            [e_ca_common_name_missing]
            some_field = true
	`
	out := test.TestLintWithCtx("e_ca_common_name_missing", inputPath, ctx)
	if out.Status != expected {
	}
}
```

# Choice of TOML

## vs. JSON
JSON does not allow for comments in its documents. This alone is the primary disqualifier.

Comments from lint authors to CAs are very important for conveying the purpose and impact of individual fields. Additionally, it is likely that CAs will version control (and possibly even template) their own configurations, in which case it is an attractive proposition that CAs be able to leave comments in their internal configurations  in order to track their own policy decisions.

## vs. YAML

YAML allows for comments. However, it is a far richer (and thus far more complex) lanuage than TOML is. It is not uncommon for humans to have difficulty writing and editing YAML by hand.

Should use cases arise such that the complexity of YAML becomes desirable then it may be a consideration for usage.

# Noted Dissatisfactions
## Long Single Line Comments
Lint authors provide comments in their configurations via the `comment:"<text>"` Go tag. Unfortunately, Go does not allow for multiline tags nor does it allow for usages of `const` or `var` strings within tags - all tags must be string literals. This means that, as is, it is very likely that the majority of comments are going to break code style rules simply by being even reasonably descriptive. This is seen immediately in the example provided in [Live Example](#live_example).

There may be remediations to this. One is to have an implicit, reflective, interface.

1. Find all public fields within the struct.
2. For every public field, attempt to find a public method named `<Field>Comment` that returns a string.

```go
type MyLint struct {
    CrossSigner string
}

func CrossSignerComment() string {
    return `
For Cross Certificates that share a Subject Distinguished Name and 
Subject Public Key with a Root Certificate operated in accordance 
with these Requirements, this extension MAY be present. If present, 
this extension SHOULD NOT be marked critical. This extension MUST 
only contain usages for which the issuing CA has verified the 
Cross Certificate is authorized to assert. This extension MAY contain 
the anyExtendedKeyUsage [RFC5280] usage, if the Root Certificate(s) 
associated with this Cross Certificate are operated by the same 
organization as the issuing Root Certificate.
`
}
```

However, spike work has not yet been done on accomplishing an interface such as this. As such it is not yet known how conducive our TOML framework would be towards being extended in such a way.

## Manual Certificate Parsing

The TOML serde framework does not have a notion of x509 certificates and, as such, lint authors are backed into the wall of declaring `string` fields and documenting that they MUST be properly formatted PEMs and then do the parsing and error handling themselves. This is seen in the [Live Example](#live_example).

Possible remediations include:

* Providing a shim `struct` that both knows how to deserialize from TOML _and_ how to convert itself into a `*x509.Certificate`.

```go
type MyLint struct {
    CrossSigner *TomlCertificate
}

func (m *MyLint) Execute() {
    var cert *x509.Certificate = m.CrossSigner.GetCertificate()
}
```

* Provide `utils` facilites.

```go
type MyLint struct {
    CrossSigner string
}

func (m *MyLint) Execute() {
    var cert *x509.Certificate = utils.CertificateFromToml(m.CrossSigner)
}
```

christopher-henderson:
Howdy folks,

The following is an implementation of the notion of configurable lints that we have been working on.

First of all, there two fortunate attributes to this code review...

1. All of the documentation written up in #624 still apply, so if there are any questions on how this behaves or is used, then that document is still accurate.
2. The _wide_ majority of this PR are tests and docs (over 75%).

However, it still may be conducive to provide a walkthrough (or, at least, give an ordering of the file that should be read carefully). I'll endeavor to at least provide said ordering.

1. [v3/lint/configuration.go](https://github.com/zmap/zlint/pull/648/files#diff-ef5cbe9f39c9fefd0fb1ea33b4450798e887987247717128374187802544e13e) - This is the most involved file in the PR. Most others are plumbing.
2. [v3/lint/global_configurations.go](https://github.com/zmap/zlint/pull/648/files#diff-943896debd9f4e05b78cc4761938c132ba6641d8f4d6886e99de28d6d267ac87) - This is a list of the pre-defined configuration sections that ZLint will ship with. Of course they are all empty at the moment, but I think it still bares taking a look at.
3. [CONTRIBUTING.md](https://github.com/zmap/zlint/pull/648/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055) - Docs on how to work with this system from a lint are important.

The rest are miscellaneous plumbing required to actually hookup the feature end-to-end as well as enabling configurations from within unit tests where desired.

This PR resolves #624


v3/cmd/zlint/main.go
```diff
@@ -47,6 +47,8 @@ var ( // flags
 	includeSources  string
 	excludeSources  string
 	printVersion    bool
+	config          string
```
christopher-henderson, line 50: The changes in `main` are to:

1. Add the new flags that either point to a configuration file are prints an example default one to the terminal.
2. If printing the default example is selected, then that is done and the problem exists. Any other flag is ignored. This behavior is consistent with printing a list of all available lints as well as printing ZLint's version.

v3/lint/configuration.go
```diff
@@ -0,0 +1,227 @@
+package lint
```
christopher-henderson, line 1: This file is the meat of the change.

v3/lint/configuration.go
```diff
@@ -0,0 +1,242 @@
...
+	if tree := c.tree.Get(namespace); tree != nil {
+		err := tree.(*toml.Tree).Unmarshal(target)
+		if err != nil {
+			return err
+		}
+	}
+	return c.resolveHigherScopedReferences(target)
+}
+
+// resolveHigherScopeReferences takes in an interface{} value and attempts to
```
christopher-henderson, line 134: Indeed, this method is what accomplished the somewhat fancy scoping feature by reflecting into the struct provided by the lint and trying to find and data member that is one of the higher scoped configuration fields.

v3/lint/configuration_test.go
```diff
@@ -0,0 +1,913 @@
+package lint
```
christopher-henderson, line 1: These tests require custom structs for each case that we would like to exercise. That, coupled with the fact the types are not values in Go, means that table driven testing cannot reasonably be done (at least, I don't immediate see a way - please speak up if you can think of one).

christopher-henderson:
Howdy folks,

Between the log4j shenanigans and the holidays I myself took December largely off from the repo, however I'm hoping to push this feature over the finish line because I think it can really unlock a good number of doors for us with regard to ZLint's power and accuracy.

I do appreciate that 1.7k additions is a bit daunting, but I hope that knowing that more than 50% of that is just unit tests that are forced to be a bit redundant can help ease the apprehension.

Please do not hesitate if you would like more from me before attempting to tackle this review.

@cpu @zakird @sleevi

zakird:
I'm just beyond underwater this month, but can take a pass on January 5th.

cpu:
I'm taking a break from volunteer open source work for the short term and won't be available to review this PR.

v3/lints/cabf_br/lint_dsa_correct_order_in_subgroup_test.go
```diff
@@ -41,7 +41,7 @@ func TestDSANotCorrectOrderSubgroup(t *testing.T) {
 	pMinusOne.Sub(dsaKey.P, big.NewInt(1))
 	dsaKey.Y = pMinusOne
 	expected := lint.Error
-	out := test.TestLintCert("e_dsa_correct_order_in_subgroup", c)
+	out := test.TestLintCert("e_dsa_correct_order_in_subgroup", c, lint.NewEmptyConfig())
```
zakird, line 44: Is there a reason to pass in an EmptyConfig here every time instead of the Default Config?

README.md
```diff
@@ -96,6 +96,12 @@ Example ZLint CLI usage:
...
 	echo "Lint mycert.pem with all of the lints except for ETSI ESI sourced lints"
 	zlint -excludeSources=ETSI_ESI mycert.pem
+    
+    echo "Receive a copy of the full (default) configuration for all configurable lints"
+    zlint -exampleConfing
+
+    echo "Lint mycert.pem using a custom configuration for any configurable lints"
+    zlint -config configFile.toml mycert.pem
 
 See `zlint -h` for all available command line options.
```
zakird, line 107: Do we need to update library usage to explain how to pass in a configuration? Do we also want someone to be able to programmatically provide configurations instead of requiring a toml based interface? It's a bit awkward to have a TOML based structure to this instead of TOML encoding an internal structure that the library expects. This could be solved by having Configuration{} object be initializable using data instead of toml.

README.md
```diff
@@ -96,6 +96,12 @@ Example ZLint CLI usage:
 
 	echo "Lint mycert.pem with all of the lints except for ETSI ESI sourced lints"
 	zlint -excludeSources=ETSI_ESI mycert.pem
+    
+    echo "Receive a copy of the full (default) configuration for all configurable lints"
+    zlint -exampleConfing
```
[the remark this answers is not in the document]
christopher-henderson: So the idea here was there is likely only going to be a small subset of lints that are actually configurable (of course, starting with 0). This makes the following questions difficult to answer - `Which lints are configurable? Is there documentation for these fields? What are their default values? If I am lint author, how do I convey the information to the user?`.

To date, the project has done a pretty good job of making it such that all a new lint has to do is add one new file, one new test file, and any necessary PEM data for tests. That is, everything is isolated making new lints rather easy for people unfamiliar to the codebase to contribute. However, if we naively asked that any new lint configurations kindly update the documentation on the current configuration, then that opens up a world of bookkeeping that can be quite lossy and error prone.

So what I've done here is make the following contract with lint authors...

> The code is going to require that you give me a default anyways, and our TOML framework lets you annotate your struct with comments, so whatever you give me as being your default configuration I will serialize into TOML and show to the end user on your behalf (comments and all).

In this way, new lints don't have to update markdown anywhere or otherwise leave the safety of their one file...

```go
type MyConfig struct {
   TheBestSongInTheWorld string `coment:"Totally your opinion"`
}

func (m (*MyLinT) Configure() interface{} {
    return &MyConfig{TheBestSongInTheWorld: "But I think it's Triangle Man by They Might be Giants"}
}
```

....and automagically that shows up in this "example" default configuration.

```toml
[MyLint]
# Totally your opinion
TheBestSongInTheWorld = "But I think it's Triangle Man by They Might be Giants"
```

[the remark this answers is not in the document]
christopher-henderson:
```suggestion
    zlint -exampleConfig
```

README.md
christopher-henderson: hmmm I admint that I completely forgot about using ZLint as a library.

I'll have to come back to this one, but at the end of the day it does all resolve down to Go structs. But honestly my answer to this depends on which layer the the library interface is exposed, which I'll get on top of.

v3/cmd/zlint/main.go
```diff
@@ -95,6 +99,16 @@ func main() {
 		return
 	}
 
+	if exampleConfig {
+		b, err := registry.DefaultConfiguration()
+		if err != nil {
+			fmt.Println(err)
```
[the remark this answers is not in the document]
christopher-henderson:
```suggestion
			log.Fatalf("a critical error occurred while generating a configuration file, %s", err)
```

v3/cmd/zlint/main.go
```diff
@@ -95,6 +99,16 @@ func main() {
 		return
 	}
 
+	if exampleConfig {
+		b, err := registry.DefaultConfiguration()
+		if err != nil {
+			log.Fatalf("a critical error occurred while generating a configuration file, %s", err)
+			os.Exit(1)
```
christopher-henderson, line 106:
```suggestion
```

v3/lint/base.go
```diff
@@ -87,14 +93,33 @@ func (l *Lint) CheckEffective(c *x509.Certificate) bool {
...
 	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) {
 		return &LintResult{Status: NA}
 	}
-	lint := l.Lint()
+	return l.execute(l.Lint(), cert, config)
+}
+
+func (l *Lint) execute(lint LintInterface, cert *x509.Certificate, config Configuration) *LintResult {
+	switch configurable := lint.(type) {
+	case Configurable:
```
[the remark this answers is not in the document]
christopher-henderson: Good...point. Don't know why I did that, honestly.

v3/lints/cabf_br/lint_dsa_correct_order_in_subgroup_test.go
christopher-henderson: Hmmm it's because there is actually merge logic at play here. As in, we take a configuration from the frontend, we take the defaults given by the lint, and merge them together with preference give to the frontend. So if an empty confguration is given by the frontend then the lint's default values are used.

v3/lint/global_configurations.go
```diff
@@ -0,0 +1,155 @@
...
+func (g Global) namespace() string {
+	return "Global"
+}
+
+// RFC5280Config is the higher scoped configuration which services as the deserialization target for...
+//
+// [RFC5280Config]
+// ...
+// ...
+type RFC5280Config struct{}
```
aarongable, line 38: Is the idea for these pre-defined config structs to gain various useful fields as part of this change, or is this PR only going to add the scaffolding while individual configurable fields will be added to these later?

v3/lint/registration.go
```diff
@@ -87,6 +92,8 @@ type Registry interface {
 	// WriteJSON writes a description of each registered lint as
 	// a JSON object, one object per line, to the provided writer.
 	WriteJSON(w io.Writer)
+	SetConfiguration(config Configuration)
+	GetConfiguration() Configuration
```
aarongable, line 96: As a library consumer of zlint, I just want to say that these new interface methods will work great for us. We [build a registry](https://github.com/letsencrypt/boulder/blob/7dcbf69536eba2d7133067c56e959873f00b4d0b/linter/linter.go#L148-L163) with our own configurable set of skip-lints; adding one more step to load and apply a config from a TOML file will be very easy.

v3/lint/registration.go
```diff
@@ -292,6 +301,63 @@ func (r *registryImpl) WriteJSON(w io.Writer) {
 	}
 }
 
+func (r *registryImpl) SetConfiguration(ctx Configuration) {
+	r.configuration = ctx
```
aarongable, line 305: nit: save `ctx` for `context.Context` just to avoid confusing readers that are used to that custom.
```suggestion
func (r *registryImpl) SetConfiguration(cfg Configuration) {
	r.configuration = cfg
```

v3/lint/registration.go
```diff
@@ -313,7 +379,7 @@ var globalRegistry = NewRegistry()
 // registration process.
 //
 // IMPORTANT: RegisterLint will panic if given a nil lint, or a lint with a nil
-// Lint pointer, or if the lint's Initialize function errors, or if the lint
+// Lint pointer, or if the lint's Initialize function errors, or if the lint`
```
aarongable, line 382: typo: stray backtick
```suggestion
// Lint pointer, or if the lint's Initialize function errors, or if the lint
```

v3/test/configuration_test_framework_test.go
```diff
@@ -0,0 +1,203 @@
...
+func TestCaCommonNameNotMissing2(t *testing.T) {
+	inputPath := "caCommonNameNotMissing.pem"
+	expected := lint.Pass
+	config := `
+[e_ca_common_name_missing2]
+BeerHall = "liedershousen"
+`
+	out := TestLintWithConfig("e_ca_common_name_missing2", inputPath, config)
+	if out.Details != "liedershousen" {
+		panic("noooooooooooo")
```
aarongable, line 93: Why panic instead of `t.Errorf(...)` here? (And a few other places below.)

v3/test/helpers.go
```diff
@@ -35,7 +39,15 @@ import (
...
 func TestLint(lintName string, testCertFilename string) *lint.LintResult {
-	return TestLintCert(lintName, ReadTestCert(testCertFilename))
+	return TestLintWithConfig(lintName, testCertFilename, "")
+}
+
+func TestLintWithConfig(lintName string, testCertFilename string, context string) *lint.LintResult {
+	ctx, err := lint.NewConfigFromString(context)
+	if err != nil {
+		panic(err)
+	}
```
aarongable, line 49: I think the more idiomatic thing to do would be to take `t *testing.T` as the first argument, declare this a `t.Helper()`, and then use `t.Fatal()` instead of directly `panic`ing.

But I recognize that that's not the pattern used by the other helpers in this file, so it's probably better to do it this way for now.

aarongable (commented):
Just dropping by to say that I'm very excited for this.

The use-case I have in mind right now is writing a lint which checks that the SCTs embedded in a cert abide by the new Chrome "two different operators" CT Policy. Doing that lint properly requires having a list of which logs are run by which operators, which means loading https://www.gstatic.com/ct/log_list/v3/log_list.json from disk, which means having a config which points to that file on disk.

README.md
```diff
@@ -96,6 +96,12 @@ Example ZLint CLI usage:
 
 	echo "Lint mycert.pem with all of the lints except for ETSI ESI sourced lints"
 	zlint -excludeSources=ETSI_ESI mycert.pem
+    
+    echo "Receive a copy of the full (default) configuration for all configurable lints"
+    zlint -exampleConfig
+
+    echo "Lint mycert.pem using a custom configuration for any configurable lints"
+    zlint -config configFile.toml mycert.pem
```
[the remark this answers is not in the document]
christopher-henderson: Tricksy four spaces instead of tabs, indeed ;)

v3/test/configuration_test_framework_test.go
christopher-henderson: Simply some first-pass yolo test code that I missed when sprucing everything up :smiley:

v3/test/helpers.go
christopher-henderson: Right, I wouldn't be opposed to that were it not for the massive pile of lint tests that have already been written using this interface (a perfunctory glance shows ~550) :sob:

v3/lint/global_configurations.go
christopher-henderson: The later (if I understood you correctly).

So the idea for these structs is to have a ready-to-go facility for when someone (inevitably) finds a configuration that is valid for a broad classification of lints. What I've done here is, as you say, set up the scaffolding such that when that day does eventually arrive it is quite clear how it would be accomplished.

v3/lint/configuration.go
```diff
@@ -0,0 +1,248 @@
...
+// find any field within its inner value that is either a struct or a pointer
+// to a struct that is one of our global configurable types. If such a field
+// exists then that higher scoped configuration will be copied into the value
+// held by the provided interface{}.
+//
+// This procedure is recursive.
```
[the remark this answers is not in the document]
christopher-henderson: You are correct that there is a caching opportunity here, however I was apprehensive when building this in providing lints with shared data for which I do not believe I can reasonably control the mutability.

```go
struct A {
    Config *RFC5280Config
}

func (a *A) Execute(cert *x509.Certificate) {
    A.Config.Field = "now something else entirely"
}

struct A {
    Config *RFC5280Config
}

func (b *B) Execute(cert *x509.Certificate) {
    // I sure hope that no one modified my config before I ran because that would be super sneaky
}
```

I suppose that for the common configurations I could do a series of `getters`, although I cannot do that in general for all configurations as I cannot know their beforehand.

However, a quick measurement shows that this is not much a problem (at least not at the moment).

```go
func BenchmarkConfig(b *testing.B) {
	config, err := lint.NewConfigFromString(`
[e_ca_common_name_missing2]
BeerHall = "liedershousenssss"
`)
	if err != nil {
		b.Fatal(err)
	}
	lint := NewCaCommonNameMissing()
	for i := 0; i < b.N; i++ {
		err = config.Configure(lint, "e_ca_common_name_missing2")
		if err != nil {
			b.Fatal(err)
		}
	}
}
```

...yields...

```
goos: linux
goarch: amd64
pkg: github.com/zmap/zlint/v3/test
cpu: Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
BenchmarkConfig
BenchmarkConfig-8   	  931068	      1250 ns/op
PASS
```

And indeed, simply tight looping this reconstruction 100k times take about 0.13s.

christopher-henderson:
Thank you @zakird and @aarongable for your patience - other low hanging fruit kept presenting itself and it was just too tempting to go for the small wins too often.

@aarongable thank you so much for hopping in here without even being asked to! It is very much appreciated to receive community interest and support, and I have been very interested in receiving feedback from consumers on the usability if this framework as well as candidate usecases.

@zakird I looked into how this would all work from the perspective of a library usage and thankfully not much had to be done beyond a bug fix, some extra tests, and the following documentation.

### From the README
Library Usage
-------------
...
...
...

To lint a certificate in the presence of a particular configuration file, you must first construct the configuration and then make a call to `SetConfiguration` in the `Registry` interface.

A `Configuration` may be constructed using any of the following functions:

* `lint.NewConfig(r io.Reader) (Configuration, error)`
* `lint.NewConfigFromFile(config string) (Configuration, error)`
* `lint.NewConfigFromString(config string) (Configuration, error)`

The contents of the input to all three constructors must be a valid TOML document.

```go
import (
	"github.com/zmap/zlint/v3"
)

var certDER []byte = ...
parsed, err := x509.ParseCertificate(certDER)
if err != nil {
	// If x509.ParseCertificate fails, the certificate is too broken to lint.
	// This should be treated as ZLint rejecting the certificate
	log.Fatal("unable to parse certificate:", err)
}
configuration, err := lint.NewConfigFromString(`
        [some_configurable_lint]
        IsWebPki = true
        NumIterations = 42
        
        [some_configurable_lint.AnySubMapping]
        something = "else"
        anything = "at all"
`)
if err != nil {
	log.Fatal("unable to parse configuration:", err)
}
lint.GlobalRegistry().SetConfigutration(configuration)
zlintResultSet := zlint.LintCertificate(parsed)
```

v3/cmd/zlint/main.go
```diff
@@ -96,6 +100,15 @@ func main() {
 		return
 	}
 
+	if exampleConfig {
+		b, err := registry.DefaultConfiguration()
+		if err != nil {
+			log.Fatalf("a critical error occurred while generating a configuration file, %s", err)
+		}
+		fmt.Printf("%s\n", string(b))
```
aarongable, line 108: nit: I believe `fmt.Println(string(b))` would be equivalent here.

README.md
```diff
@@ -149,6 +155,45 @@ if err != nil {
 zlintResultSet := zlint.LintCertificateEx(parsed, registry)
 ```
 
+To lint a certificate in the presence of a particular configuration file, you must first construct the configuration and then make a call to `SetConfiguration` in the `Registry` interface.
+
+A `Configuration` may be constructed using any of the following functions:
+
+* `lint.NewConfig(r io.Reader) (Configuration, error)`
+* `lint.NewConfigFromFile(config string) (Configuration, error)`
```
aarongable, line 163: nit: The argument to this function is called `path`, not `config`, in the actual function definition.

v3/cmd/zlint/main.go
```diff
@@ -203,6 +216,11 @@ func trimmedList(raw string) []string {
 // use.
 //nolint:cyclop
 func setLints() (lint.Registry, error) {
+	config, err := lint.NewConfigFromFile(config)
```
aarongable, line 219: nit: this kind of shadowing (here, `config` is now of type `Configuration`, instead of type `string`) can be difficult to read. I'd use a different variable name here to avoid any chance of confusion.

v3/test/helpers.go
```diff
@@ -35,7 +39,15 @@ import (
 // lint result is nil.
 //nolint:revive
 func TestLint(lintName string, testCertFilename string) *lint.LintResult {
-	return TestLintCert(lintName, ReadTestCert(testCertFilename))
+	return TestLintWithConfig(lintName, testCertFilename, "")
+}
+
+func TestLintWithConfig(lintName string, testCertFilename string, context string) *lint.LintResult {
+	ctx, err := lint.NewConfigFromString(context)
```
aarongable, line 46: nit: calling it a `context` and naming the variable `ctx` is likely to be confusing to people who tightly associate that name with `context.Context`. I'd rename these `configStr` and `config`, or `config` and `cfg`, or something like that.

The same also applies a couple times below where the name `ctx` is used for a `lint.Configuration`.

aarongable (commented):
All the core logic code looks good to me, just a few nits around the documentation and naming!

christopher-henderson:
Howdy @aarongable, could you take one more perfunctory glance at this after I addresses some of the nits (I honestly have no idea why GH won't let me cycle your review status...). I think we're just about ready to get this change in in preparation for a minor version bump :smile:

aarongable (approved):

PR #652, opened 2021-12-10
Entrust Datacard rebranded to Entrust

vanbroup:


christopher-henderson (approved):
I admit that I had to go look this up real quick to verify :laughing: thank you!

PR #660, opened 2022-02-06
Permit underscores in DNSNames if-and-only-if those certificates are valid for less than 30 days and during BR 1.6.2's permissibility period

christopher-henderson:
This lint encodes a brief period from December 10th 2018 to April 1st 2019 wherein CABF BR permitted underscores within DNS names if-and-only if those certificates were valid for fewer than 30 days.


v3/util/time.go
```diff
@@ -60,8 +60,10 @@ var (
 	MozillaPolicy24Date         = time.Date(2017, time.February, 28, 0, 0, 0, 0, time.UTC)
 	MozillaPolicy241Date        = time.Date(2017, time.March, 31, 0, 0, 0, 0, time.UTC)
 	MozillaPolicy27Date         = time.Date(2020, time.January, 1, 0, 0, 0, 0, time.UTC)
+	BALLOT_SC_12_Ineffective    = time.Date(2019, time.April, 1, 0, 0, 0, 0, time.UTC)
```
sleevi, line 63:
```suggestion
	BallotSC12Ineffective    = time.Date(2019, time.April, 1, 0, 0, 0, 0, time.UTC)
```

WDYT? (e.g. lines 47-62)

sleevi (changes requested):
A minor nit here, and then I suspect that the dupe certs will get resolved from the other PR. If you decide to not go with that, then I think this is pretty good.

v3/util/time.go
christopher-henderson: Thanks, I hop back-and-forth between Go/Rust/Python/Java often enough that the naming conventions bleed over each other sometimes.

v3/lints/cabf_br/lint_underscore_present_with_too_long_validity.go
```diff
@@ -0,0 +1,59 @@
...
+		EffectiveDate:   util.CABFBRs_1_6_2_Date,
+		IneffectiveDate: util.CABFBRs_1_6_2_UnderscorePermissibilitySunsetDat,
+		Lint:            func() lint.LintInterface { return &UnderscorePresentWithTooLongValidity{} },
+	})
+}
+
+type UnderscorePresentWithTooLongValidity struct{}
+
+func (l *UnderscorePresentWithTooLongValidity) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.DNSNamesExist(c)
```
sleevi, line 42: Can/should the check from 47 - 49 be moved here?

```suggestion
	return util.IsSubscriberCert(c) &&
	       util.DNSNamesExist(c) &&
	       c.NotBefore.AddDate(0, 0, 30).Before(c.NotAfter)
```

Might be a way to simplify further. But is a bit of a suggestion

v3/testdata/dNSNoUnderscoresLongValidity.pem
```diff
@@ -0,0 +1,37 @@
...
+                    04:9d:46:d8:ce:3d:5f:ff:e2:70:6c:30:04:3f:0e:
+                    aa:b0:7c:50:8c:48:89:6e:ee:19:e6:e4:e4:31:3a:
+                    d0:41:e9:2d:b8:de:ef:99:fc:5e:26:fd:ef:b9:e0:
+                    1a:6a:ab:c5:15:8a:b8:b5:2c:7a:ac:7e:b8:70:61:
+                    58:88:2c:52:38
+                ASN1 OID: prime256v1
+                NIST CURVE: P-256
+        X509v3 extensions:
+            X509v3 Subject Alternative Name: 
+                DNS:no.underscore.com
```
sleevi, line 24: ditto re: `.test` fixes

christopher-henderson: Done (thank you for your patience on this, I've been traveling for business and I almost missed a flight trying to juggle these similiar-but-different certs so I just punted for the week).

v3/lints/cabf_br/lint_underscore_present_with_too_long_validity.go
christopher-henderson: I don't think so? If we do this check here then what we would be saying is that this lint does not apply to this certificate at all (which is usually reserved for scenarios such as being an irrelevant cert type or not having a particular field).

I do prefer that method of adding 30 days, though. Far more correct than the naive `time.Hour*24*30`.

christopher-henderson:
@sleevi gentle re-engagement on this

v3/lints/cabf_br/lint_underscore_present_with_too_long_validity.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:            "e_underscore_present_with_too_long_validity",
+		Description:     "From 20128-12-10 to 2019-04-01, DNSNames may contain underscores if-and-only-if the certificate is valid for less than thirty days.",
```
cardonator, line 29: nit: there is a typo on this date field (20128 vs 2018)

cardonator (approved):
Thi looks good other than the one nit.

dadrian (approved):

PR #662, opened 2022-02-06
No underscores are allowed in DNSNames after BR 1.6.2's permissibility period

christopher-henderson:
This lint encodes that, post the brief permissibility period from December 10th 2018 to April 1st 2019, CABF BR does not permit the underscore character anywhere within DNS names.


v3/lints/cabf_br/lint_underscore_not_permissible_in_dnsname.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/zmap/zlint/v3/lint"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_underscore_not_permissible_in_dnsname",
+		Description:   "DNSNames MUST NOT contain underscore characters",
+		Citation:      "BR 7.1.4.2.1",
+		Source:        lint.CABFBaselineRequirements,
+		EffectiveDate: util.BALLOT_SC_12_Ineffective,
```
sleevi, line 33: I realize we're all over the board with dates, so this is very much a "nit" (and already the suggestion elsewhere to rename this, which I suppose this would replace), but wdyt about

`CABFBRs_1_6_2_UnderscoreSunsetDate`

It's very long, so it may be a terrible name though.

sleevi (approved):
I realize since this is a chain of PRs, the bits like the domain names will be addressed, and this may be tweaked first in those other CLs. But LG

v3/lints/cabf_br/lint_underscore_not_permissible_in_dnsname.go
christopher-henderson: The interesting thing about this repo and is that how often "breaking the rules" is actually  the preferred option.

`LintThisButOnlyWhenThatAndOnlyOnTheseDays`

This is an awful name, but the alternative would be vague (at best) and most likely misleading.

PR #669, opened 2022-03-26
Skip checking for a Tor Descriptor Hash if the provided cert contains a V3 Onion address.

### LINKED ISSUE #667: Zlint incorrectly requires TorServiceDescriptors for onion V3 domains
Example cert: https://crt.sh/?id=6357116915

ZLint version: 3.3.0

christopher-henderson:
As a part of the discussion in #667 it was agreed upon that EV certificates do not require a Tor Descriptor Hash IF the certificate is encoding a V3 Onion address(es).

I think perhaps one of the more interesting decisions I made here was that if a certificate contains even a single non-v3 address then the entire certificate is considered non-v3. This _seems_ reasonable to me, although I am uncertain if there is such a thing a V2/V3 heterogeneous certificate out in the wild.

Mentioning @mimi89999 since I can't bring you on as reviewer explicitly.

This PR resolves #667


cardonator:
These changes seem positive to me. I know an onion cleanup ballot is being discussed in the forum right now. Will that affect how these lints should work going forward? I think we are planning to deprecate V2 Onion EV certs at least.

zakird (approved):

mimi89999 (approved):

christopher-henderson:
@cardonator I'm not familiar with the work being drafted up for onion, but I can say that even though certain language and technologies get deprecated, that ZLint tends towards keeping those lints around as-is (for both historical interest as well as the likelihood that there is _someone_ out there using these lint result to litigate their own point in private discussions regarding some in-house implementations/deployments). This is precisely why we have the `Inneffective` date declaration so that certificates can exist within a bubble of time without either being completely ignored or poisoning future ecosystems.

cardonator:
Ah, right, good call. I forgot about the ineffective date. That sounds good.

PR #672, opened 2022-04-09
Update to Go 1.18 and update GolangCI Linter

christopher-henderson:
*sigh* golangci broke itself and began to fail to compile which forced an upgrade to a version that added just a ton of useless and asinine lints.


zakird (approved):

.golangci.yaml
```diff
@@ -33,6 +33,7 @@ linters:
     - goconst
     - golint
     - godot
+    - forbidigo
```
christopher-henderson, line 36: `forbidgo` is profoundly useless. It literally bans certain regexes from your codebase - notable any use of `fmt.Print(f|ln)` (because I guess it's illegal to print to stdout?)

zakird:
@christopher-henderson I didn't merge this tests seemed to be failing (and now merge conflict). I can if we need to force it through (post merge)

christopher-henderson:
Thanks so @zakird for hopping on a batch of topics yesterday :pray: 

This one...this one is getting an exasperated sigh from me. Master is broken because the linter itself no longer works and requires an  upgrade (including an upgrade to Go 1.18 which is all well and good).

Unfortunately, the linter update is coming bundled with an absolute ocean of useless and frankly, idiotic, lints For example, the variable `ip` no longer appears to be good enough for the linter (it is clearly requiring that anything with a global scope have a minimum name length which is...naive). And it's not a warning, the variable `ip` is a full on error.

So I'm fighting the urge to blindly disable this thing and pecking through its demands,

zakird:
I vote disable it, get master back to healthy, and then we can deal with
this independently of everything else.

Issue #676, opened 2022-05-18
Zlint incorrectly requires TorServiceDescriptors if V3 onion and DNS name


mimi89999:
Zlint incorrectly requires TorServiceDescriptors if a certificate contains a V3 onion name and DNS name. I failed to notice that when reviewing https://github.com/zmap/zlint/pull/669.

A certificate https://crt.sh/?id=6748280018 was issued that contains a V3 onion name and a DNS name, that I believe is valid, but has the `e_ext_tor_service_descriptor_hash_invalid` lint error.

PR #677, opened 2022-05-21
Zlint incorrectly requires TorServiceDescriptors if V3 onion and DNS name

christopher-henderson:
This patch addresses #676 wherein the presence of a normal DNS entry would trigger a `false` return on `IsOnionV3`, resulting in the Onion V3 cert being checked as though it were V2.

I believe that the solution here is to change this from checking whether-or-not all names are V3 into checking whether-or-not at least one name is V2.

@mimi89999


mimi89999:
Hmmm. I think that it might be better to write `IsOnionV2` and `IsOnionV3` that would check `.onion` names. They could be reused in another lint that would check if `.onion` names are valid.

mimi89999:
In https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_san_dns_name_onion_invalid.go#L133 you only check the address length, but in https://github.com/zmap/zlint/blob/master/v3/util/onion.go#L25-L39 you check the length and the last byte.

christopher-henderson:
@mimi89999 

> I think that it might be better to write IsOnionV2 and IsOnionV3 that would check .onion names. They could be reused in another lint that would check if .onion names are valid.

Sure thing! Indeed, with our newfound access to generics in Go, I went ahead and added some utility functions that help fill the gap of `Any` and `All` which would help with this.

> In https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_san_dns_name_onion_invalid.go#L133 you only check the address length, but in https://github.com/zmap/zlint/blob/master/v3/util/onion.go#L25-L39 you check the length and the last byte.

Fair enough, although I think the only extra thing I can reasonably do here is check that it is lowercase base32 encoded. That is, V3 has a number of extra semantics to it, however a V2 address is nothing but a b32 encoded hash of length 16.

mimi89999 (approved):

v3/util/onion.go
```diff
@@ -31,24 +31,74 @@ func IsOnionV3(dnsName string) bool {
...
+func IsOnionV2Address(dnsName string) bool {
+	if !strings.HasSuffix(dnsName, "onion") {
+		return false
+	}
+	labels := strings.Split(dnsName, ".")
+	if len(labels) < 2 {
+		return false
 	}
-	return isV3
+	if len(labels[0]) != 16 {
```
mathewhodson, line 55: Based on the comment at the top of this function, this should check the second last label, which would be something like this:
```
address := labels[len(labels)-2]
if len(address) != 16 {
```

v3/util/onion.go
```diff
@@ -31,24 +31,74 @@ func IsOnionV3(dnsName string) bool {
...
+		return false
+	}
+	_, err := onionBase32Encoding.DecodeString(labels[0])
```
mathewhodson, line 58: @christopher-henderson Same here

PR #686, opened 2022-08-21
Lint to check for invalid KU lengths

christopher-henderson:
@aarongable since you're the one that thought of this lint

As per [RFC 5280 Section 4.2.1.3](https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.3) a key usage is a bit string with only eight possible flags.

```
      KeyUsage ::= BIT STRING {
           digitalSignature        (0),
           nonRepudiation          (1), -- recent editions of X.509 have
                                -- renamed this bit to contentCommitment
           keyEncipherment         (2),
           dataEncipherment        (3),
           keyAgreement            (4),
           keyCertSign             (5),
           cRLSign                 (6),
           encipherOnly            (7),
           decipherOnly            (8) }

```

This lint checks that, if a KU is present, that it's bitstring is exactly one byte long.

### DIFF
--- a/v3/lints/rfc/lint_key_usage_incorrect_length.go
+++ b/v3/lints/rfc/lint_key_usage_incorrect_length.go
@@ -16,6 +16,7 @@ package rfc
 
 import (
 	"fmt"
+	"math/big"
 
 	"github.com/zmap/zcrypto/x509"
 	"github.com/zmap/zlint/v3/lint"
@@ -27,7 +28,7 @@ type keyUsageIncorrectLength struct{}
 func init() {
 	lint.RegisterLint(&lint.Lint{
 		Name:          "e_key_usage_incorrect_length",
-		Description:   "The key usage is a bit string with exactly eight bits",
+		Description:   "The key usage is a bit string with exactly nine possible flags",
 		Citation:      "RFC 5280: 4.2.1.3",
 		Source:        lint.RFC5280,
 		EffectiveDate: util.RFC5280Date,
@@ -50,8 +51,13 @@ func (l *keyUsageIncorrectLength) Execute(c *x509.Certificate) *lint.LintResult 
 	// Unused: keyUsage[2]
 	// The actual key usage...
 	content := keyUsage[3:]
-	if len(content) != 1 {
-		return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("key usage is %d bytes long (should be exactly one)", len(content))}
+	// Any combination of the nine bit flags is legal from perspective
+	// of this lint (although requirements elsewhere may limit the combinations).
+	//
+	// As such, any value greater than 512 (2**9) is out of range of the possible
+	// values for a key usage bit string.
+	if big.NewInt(0).SetBytes(content).Int64() > 0b111111111 {
+		return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("the key usage (%v) contains a value that is out of bounds of the range of possible KU values.", content)}
 	} else {
 		return &lint.LintResult{Status: lint.Pass}
 	}
--- a/v3/lints/rfc/lint_key_usage_incorrect_length.go
+++ b/v3/lints/rfc/lint_key_usage_incorrect_length.go
@@ -15,8 +15,11 @@ package rfc
  */
 
 import (
+	"encoding/asn1"
 	"fmt"
 	"math/big"
+
+	"golang.org/x/crypto/cryptobyte"
 
 	"github.com/zmap/zcrypto/x509"
 	"github.com/zmap/zlint/v3/lint"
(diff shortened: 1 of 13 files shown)


zakird (approved):

CBonnell:
I believe there's 9 possible values as defined in X.509 08/2005 and RFC 5280, so it is possible to have a KU BIT STRING whose value component is 2 octets.

zakird:
Nice catch @CBonnell. It does look that way to me too, even in the Enum in the description.

christopher-henderson:
Yikes, that's an embarassing misstep. Thanks @CBonnell and @zakird!

I swapped up the lint to do the following instead:

1. Map the byte contents to an `int64`.
2. Check for any value >= 512.

My reasoning is that this lint is primarily enforcing that there are no unknown flags present. That is, it does not concern itself with what combinations of flags may-or-may not be valid vis-a-vis a particular requirement, but rather only that there are no `1` flags set beyond the 9th bit.

v3/lints/rfc/lint_key_usage_incorrect_length.go
```diff
@@ -0,0 +1,64 @@
...
+	// Length: keyUsage[1]
+	// Unused: keyUsage[2]
+	// The actual key usage...
+	ku := keyUsage[3:]
+	// Any combination of the nine bit flags is legal from perspective
+	// of this lint (although requirements elsewhere may limit the combinations).
+	//
+	// As such, any value greater-than-or-equal-to 512 (2**9) is out of range of the possible
+	// values for a key usage bit string.
+	if big.NewInt(0).SetBytes(ku).Int64() > 0b111111111 {
```
aarongable, line 59: Suppose we have the keyUsage ASN1
```
0x30 0x03 0x07 0x80 0x80
````
This is
* `0x30`: bitString,
* `0x03`: 3 bytes (including the unused-indicator byte),
* `0x07`: 7 bits unused in the last byte,
* `0x80 0x80`: the bit string `10000000 10000000`, or rather `100000001` followed by 7 unused 0s.

This is a valid keyUsage value for the purposes of this lint: it sets flags 0 (digitalSignature) and 8 (decipherOnly).

However, this code doesn't cut off the unused bits before turning them into a bigint. It wouldn't run `big.NewInt(0).SetBytes(0b100000001)`, it would run `big.NewInt(0).SetBytes(0b1000000010000000)`. And that's [32896](https://go.dev/play/p/TCK6Y1-NMv6), significantly greater than 512. So I _think_ my hypothetical valid KU value would show up as invalid via this lint.

And even if I'm wrong (which I hope I am!) the fact that I'm not sure means this lint isn't obviously-correct.

I think that the clearer thing to do here would be to rely on cryptobyte to parse the bitstring value for us:
```go
keyUsageExt := cryptobyte.String(util.GetExtFromCert(c, util.KeyUsageOI).Value)
var keyUsageVal encoding_asn1.BitString
ok := keyUsageExt.ReadASN1BitString(&out)
if !ok {
		return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("the key usage (%v) extension is not parseable.", ku)}
}
if out.BitLength > 9 {
		return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("the key usage (%v) contains a value that is out of bounds of the range of possible KU values.", ku)}
}
```
Or something like that.

christopher-henderson: Thank you @aarongable!

So cryptobyte isn't truncating away the trailing bytes for us either, but I swapped over to using it because it is doing some extra validation which is always nice.

In order to address the issue, I've taken to simply bit shifting the `int64` representation value to the right `unused` number of times before comparing to 512.

v3/lints/rfc/lint_key_usage_incorrect_length.go
```diff
@@ -0,0 +1,66 @@
...
+
+func keyUsageIncorrectLengthBytes(kuBytes []byte) *lint.LintResult {
+	keyUsageExt := cryptobyte.String(kuBytes)
+	var keyUsageVal asn1.BitString
+	ok := keyUsageExt.ReadASN1BitString(&keyUsageVal)
+	if !ok {
+		return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("the key usage (%v) extension is not parseable.", kuBytes)}
+	}
+	unused := kuBytes[2]
+	if big.NewInt(0).SetBytes(keyUsageVal.Bytes).Int64()>>unused >= 512 {
```
aarongable, line 58: If the bitstring were so large as to not be representable as an Int64, then the result here would be undefined, and possibly something like "all zeros" which then passes then "less than 512" test.

Even though that seems super unlikely, I'd add a test for that explicitly:

```suggestion
val := big.NewInt(0).SetBytes(keyUsageVal.Bytes)
if !val.IsInt64() || val.Int64()>>unused >= 512 {
```

aarongable (approved):
LGTM with one small improvement

PR #687, opened 2022-08-30
Add lint for checking that server certificates do not contain any OUs in the Subject

defacto64:
Add lint for checking that server certificates do not contain any OUs in the Subject


defacto64:
I am correcting test data and will open a new pull request.

PR #675, opened 2022-05-10
Add new lint to block organisational unit names as of 1st September 2022

robplee:
I've attempted to address #628.  ~I took a guess at what the appropriate version number for the BRs will be but please let me know if I've guessed wrong.~  (Nevermind, I realised that the version of the document is not as important as the rule change has a particular date mentioned and I can just describe that instead of a version number for the BRs)   I ran make integration -overwriteExpected to try and get an entry in the config.json record for the new lint but it didn't seem to change the file so I think I might have made a mistake somewhere but I can't work out where as it looks right to me...

So far, I've just tackled the easy bit of SC47 and banned them utterly from 1st September.  I'll try to suss out the combination rules and add a rule effective sooner but I am less certain if that's needed if it's outright banned in September but I'll see what I can work out from the PR Corey linked in the issue he opened.

Also, I'm English so can reviewers be extra careful to check that I've correctly, incorrectly spelled "organizational" across all the files I've added/changed 😄

PR #689, opened 2022-08-31
 Add lint for checking that server certificates do not contain any OUs in the Subject

defacto64:
Add lint for checking that server certificates do not contain any OUs in the Subject, according to BRs 7.1.4.2.2, letter i) as modified by ballot SC47 (since BR 1.7.9).


robplee:
Sorry to self-cite but can you explain what this PR is adding that wasn't covered by #675 ?

defacto64:
@robplee You are right, I just was not aware of #675, so I am going to close my pull request.

However, the prohibition of OUs only affects Subscriber (i.e. TLS server) certificates, so it should not be applied to CA certificates or other types of certificates. Therefore the CheckApplies() function should return false if the certificate is not a TLS server certificate.

defacto64:
Closing this pull request as it's a duplicate of #675 that I was not aware of.

robplee:
@defacto64, I've done some digging(/reminding myself about how zlint works) and I think the lint I merged in is fine and the CheckApplies() is fine how it is despite just returning true.  The reason is because the lint source is the CA/B Forum BRs.  If you check here:

https://github.com/zmap/zlint/blob/44e12c12ca43a4af86f0dc2da4a71493ac9f8345/v3/lint/base.go#L100

Then you'll see that CABF BR lints are only applied against certs with EKU Any or EKU ServerAuth so I think we are ok here.  However, the above link does have a bit of a code smell about it to me as I'd prefer to see the CABF BR lints have adequate CheckApplies functions rather than having this extra check elsewhere that is easily forgotten about and probably causes frustration for some of the people who import zlint in to their projects.

One day I'll open an issue and PR to try and address this.  But it may not be that soon :)

defacto64:
@robplee I do not think that is sufficient. The util.IsServerAuthCert() util just checks that the certificate has either Any or ServerAuth in the EKU, but keep in mind that intermediate CAs for issuing TLS Server certs MUST (per the BRs) also have ServerAuth in their EKU, so this util function is not enough. At least, IMO, you should also add a check that the certificate is not a CA cert.

robplee:
Ach!  So close, I'll open an issue/PR today.  Good catch!

PR #711, opened 2023-04-27
Always perform e_cert_unique_identifier_version_not_2_or_3 

mtgag:
Lint e_cert_unique_identifier_version_not_2_or_3 checks whether the unique identifier(s) is present and then checks that the version is correct. Lint lint_cert_extensions_version_not_3.go takes another approach. It always performs the linting and the check is done during execution as one step. Both are very similar cases and the implementation could be aligned. Proposal: always perform e_cert_unique_identifier_version_not_2_or_3 and check inside the execution the presence and correct values. (Alternatively align lint_cert_extensions_version_not_3 to the approach of e_cert_unique_identifier_version_not_2_or_3.)


christopher-henderson (approved):
I suppose this makes sense since it technically applies to all certificates.

Issue #712, opened 2023-05-02
Lint Coverage of SMIME BRs version 1.0.0


robplee:
- [ ] SMIME certificates SHALL have cRLDistributionPoints (7.1.2.3.b)
- [ ] Strict and Multipurpose SMIME certificates SHALL have the cRLDistributionPoints URI scheme as HTTP others are not permitted (7.1.2.3.b)
- [ ] Strict and Multipurpose SMIME certificate AIA fields: OCSP Responder "When provided, every accessMethod SHALL have the URI scheme HTTP." (7.1.2.3.c.1)
- [ ] Strict and Multipurpose SMIME certificate AIA fields: caIssuers "When provided, every accessMethod SHALL have the URI scheme HTTP." (7.1.2.3.c.1)
- [ ] Key usage, RSA certs, strict policies: prevent all key usages other than digitalSignature, nonRepudiation, keyEncipherment (7.1.2.3.e)
- [ ] Key usage, RSA certs, multipurpose/legacy policies: prevent all key usages other than digitalSignature, nonRepudiation, keyEncipherment and dataEncipherment (7.1.2.3.e)
- [ ] Key usage, EC certs, all: prevent all key usages other than digitalSignature, nonRepudiation, keyAgreement, encipherOnly, decipherOnly (7.1.2.3.e)
- [ ] Key usage, EC certs, all: encipherOnly/decipherOnly are permitted only when keyAgreement is set (7.1.2.3.e)
- [ ] Extended key usage, strict: emailProtection SHALL be present.  Other values SHALL NOT BE PRESENT (7.1.2.3.f)
- [ ] Extended key usage, multipurpose/legacy: emailProtection SHALL be present. Other values MAY be present  (7.1.2.3.f)
- [ ] Extended key usage, all: serverAuth, codeSigning, timeStamping, anyExtendedKeyUsage SHALL NOT BE PRESENT (7.1.2.3.f)
- [ ] authorityKeyIdentifier, all: SHALL be present, SHALL NOT be critical.  keyIdentifier SHALL be present, authorityCertIssuer and authorityCertSerialNumber SHALL NOT be present (7.1.2.3.g)
- [ ] subjectAlternativeName, all: SHALL be present (7.1.2.3.h)
- [ ] subjectAlternativeName, all: SHALL NOT be marked critical unless subject field is empty (7.1.2.3.h)
- [ ] subjectDirectoryAttributes, strict/multipurpose: field is Prohibited (7.1.2.3.j)
[...]

PR #713, opened 2023-05-02
Add lint enforcing the restrictions on subject DN fields for mailbox validated SMIME certificates

robplee:
This PR is the first to attempt to address something from the big todo list in #712.  It's a simple lint enforcing the field MAY/SHALL NOT table for subject DN fields in mailbox validated certificates.  I think the changes are fairly straightforward so we should be able to discover if there are any extra things needed for us to start supporting SMIME BR linting.

My current wonder is about the TestCorpus for integration testing as I assume it won't have any applicable certs in it but I guess that's something we can try to work out now?


v3/lints/cabf_smime_br/mailbox_validated_enforce_subject_field_restrictions.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_mailbox_validated_enforce_subject_field_restrictions",
+		Description:   "SMIME certificates complying to mailbox validated profiles MAY only contain commonName, serialNumber or emailAddress attributes in the Subject DN",
+		Citation:      "SMIME BRs: 7.1.4.2.3",
+		Source:        lint.CABFSMIMEBaselineRequirements,
+		EffectiveDate: util.CABF_SMIME_BRs_1_0_0_Date,
+		Lint: func() lint.LintInterface {
```
aaomidi, line 38: You may want to use CertificateLintInterface now as LintInterface is deprecated

v3/lints/cabf_smime_br/mailbox_validated_enforce_subject_field_restrictions.go
```diff
@@ -0,0 +1,94 @@
...
+)
+
+// mailboxValidatedEnforceSubjectFieldRestrictions - linter to enforce MAY/SHALL NOT requirements for mailbox validated SMIME certificates
+type mailboxValidatedEnforceSubjectFieldRestrictions struct {
+	forbiddenSubjectFields map[string]string
+	allowedSubjectFields   map[string]string
+}
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
```
aaomidi, line 32: This also ends up being RegisterCertificateLint

v3/lints/cabf_smime_br/mailbox_validated_enforce_subject_field_restrictions.go
```diff
@@ -0,0 +1,94 @@
...
+import (
+	"fmt"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+// mailboxValidatedEnforceSubjectFieldRestrictions - linter to enforce MAY/SHALL NOT requirements for mailbox validated SMIME certificates
+type mailboxValidatedEnforceSubjectFieldRestrictions struct {
```
aaomidi, line 26: This seems specifically be targeting the "strict" profile from the smime BRs. Do you think the lint name should reflect that?

v3/testdata/mailboxValidatedStrictWithEmailAddress.pem
```diff
@@ -0,0 +1,39 @@
+Certificate:
+    Data:
+        Version: 3 (0x2)
+        Serial Number: 3 (0x3)
+        Signature Algorithm: ecdsa-with-SHA256
+        Issuer: 
+        Validity
+            Not Before: Sep  2 00:00:00 2023 GMT
+            Not After : Nov 30 00:00:00 9998 GMT
+        Subject: emailAddress = brains@tracyisland.com
```
aaomidi, line 10: I would recommend for test data to use `example.com`

v3/lints/cabf_smime_br/mailbox_validated_enforce_subject_field_restrictions.go
robplee: So, this lint is actually targeting all three of the levels of mailbox-validated SMIME lints.  The table in 7.1.4.2.3 has Legacy, Multipurpose and Strict mailbox-validated certs allowed the same fields

aaomidi (approved):
Looks reasonable to me! But my approval doesn't carry much weight either :)

v3/lints/cabf_smime_br/mailbox_validated_enforce_subject_field_restrictions.go
```diff
@@ -0,0 +1,96 @@
...
+// CheckApplies is returns true if the certificate's policies assert that it conforms to the mailbox validated SMIME BRs
+func (l *mailboxValidatedEnforceSubjectFieldRestrictions) CheckApplies(c *x509.Certificate) bool {
+	return util.IsMailboxValidatedCertificate(c)
+}
+
+// Execute applies the requirements on what fields are allowed for mailbox validated SMIME certificates
+func (l *mailboxValidatedEnforceSubjectFieldRestrictions) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, rdnSeq := range c.Subject.OriginalRDNS {
+		for _, field := range rdnSeq {
+			oidStr := field.Type.String()
```
christopher-henderson, line 84: I'm assuming that you chose to use the stringer method rather than the parsed OID because you wished to do a lookup into the `forbiddenSubjectField` member? And because the Go `map` type doesn't support user defined key types (thsu forcing us into using `strings`)?

v3/lint/base.go
```diff
@@ -221,6 +221,9 @@ func (l *CertificateLint) Execute(cert *x509.Certificate, config Configuration)
 	if l.Source == CABFBaselineRequirements && !util.IsServerAuthCert(cert) {
 		return &LintResult{Status: NA}
 	}
+	if l.Source == CABFSMIMEBaselineRequirements && !util.IsEmailProtectionCert(cert) {
```
christopher-henderson, line 224: Opening salvo!

v3/util/ca.go
```diff
@@ -62,3 +62,15 @@ func IsServerAuthCert(cert *x509.Certificate) bool {
 	}
 	return false
 }
+
+func IsEmailProtectionCert(cert *x509.Certificate) bool {
+	if len(cert.ExtKeyUsage) == 0 {
```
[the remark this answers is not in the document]
robplee: I can try.  I'm sure I can come up with something better than "Because the function above for IsServerAuthCert does it this way."  I guess the main reason to return true where there are no EKUS is that this approach errs on the side of applying linting but it also seems like a mistake as the BRs apply where a certificate policy is present so this seems a backwards.  If anything we should have lints that enforce EKUs based on certificate policies rather than applying rules based on EKU...

v3/lints/cabf_smime_br/mailbox_validated_enforce_subject_field_restrictions.go
robplee: Exactly, using strings instead of straight OIDs for both the naughty and nice subject fields makes it possible to look them up in a map which allows for nicer error messages as well as keeping the implementation a bit cleaner and creates a pattern that can easily be reused for lints covering the SHALL/MAY/SHALL NOT lists of the other types of SMIME certs (organization/sponsor/individual-validated Legacy/Multipurpose/Strict)

v3/util/ca.go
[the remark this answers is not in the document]
cardonator: Policy wise, I think that makes sense, however, I think in the ServerAuth case it's more likely because ServerAuth linting is the "default" linting behavior of zlint. In other words, due tot he way that zlint is user pre-issuance, it may be possible that a linted cert doesn't have its EKUs yet or something (it seems wrong to me, but also seems like something that could happen).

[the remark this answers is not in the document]
robplee: I've taken a swing at writing a comment for this function to explain the reasoning.  Let me know if you have any comments/complaints/suggestions

christopher-henderson (approved):

christopher-henderson:
> My current wonder is about the TestCorpus for integration testing as I assume it won't have any applicable certs in it but I guess that's something we can try to work out now?

@robplee I'll take on this investigation as it's an interesting question. The test corpus from a [query](https://github.com/zmap/zlint-test-corpus/blob/master/query.sql) against the censys.io database, so I am as yet unsure of whether-or-not these certificates will be available to us. My guess would be `no` since it's my understanding that Censys get its certificate database by crawling the open internet.

cardonator:
I believe this PR sets some basic precedents and expectations required to start plowing through #712, so what else needs to be done here do you think? Does this need further review or just testing against a large corpus of SMIME certs? I could try to help run against an internal corpus here to at least get some initial testing results if that's needed before we feel comfortable merging this PR.

christopher-henderson:
@cardonator 

> Does this need further review or just testing against a large corpus of SMIME certs?

I have actually been pecking away at this over the past several weeks. We got a new batch of certs 100k+ large from Censys, but it's been a lot of work to format the corpus, deduplicate, sanitize, and smoke that the new failures make sense.

> ...before we feel comfortable merging this PR.

In the interest of time I think that we would be fine merging this now. If things go truly south in the immediate absence of more test certs then it will be easy to soft "back out" changes by simply not running SMIME checks,

christopher-henderson:
@cardonator 

> Does this need further review or just testing against a large corpus of SMIME certs?

I have actually been pecking away at this over the past several weeks. We got a new batch of certs 100k+ large from Censys, but it's been a lot of work to format the corpus, deduplicate, sanitize, and smoke that the new failures make sense.

> ...before we feel comfortable merging this PR.

In the interest of time I think that we would be fine merging this now. If things go truly south in the immediate absence of more test certs then it will be easy to soft "back out" changes by simply not running SMIME checks,

mtgag:
Would it be meaningful to change 

https://github.com/zmap/zlint/blob/master/v3/lints/cabf_smime_br/mailbox_validated_enforce_subject_field_restrictions.go#L80

``
	return util.IsMailboxValidatedCertificate(c)
``

to:

``
	return util.IsMailboxValidatedCertificate(c) && util.IsSubscriberCert(c)
``

because this is under Section 7.1.4.2 which regards subscriber certificates?

PR #717, opened 2023-05-17
Use help Method beforeoron instead of 

mtgag:
Two lints use the before method to the calculate the validity of certificates, while the BeforeOrOn help method should be used. Two certificates that test the edge case are also added.


v3/lints/cabf_br/lint_sub_cert_valid_time_longer_than_39_months.go
```diff
@@ -42,7 +42,7 @@ func (l *subCertValidTimeLongerThan39Months) CheckApplies(c *x509.Certificate) b
 }
 
 func (l *subCertValidTimeLongerThan39Months) Execute(c *x509.Certificate) *lint.LintResult {
-	if c.NotBefore.AddDate(0, 39, 0).Before(c.NotAfter) {
+	if util.BeforeOrOn(c.NotBefore.AddDate(0, 39, 0), c.NotAfter) {
```
robplee, line 45: Is this actually correct to make this change?  The Description of the lint says that the lint validity period must be "no greater than 39 months".  A certificate that is valid for exactly 39 months doesn't have a validity period greater than 39 months so I think the original code is the correct version here.

v3/lints/cabf_ev/lint_ev_valid_time_too_long.go
```diff
@@ -47,7 +47,7 @@ func (l *evValidTooLong) CheckApplies(c *x509.Certificate) bool {
 }
 
 func (l *evValidTooLong) Execute(c *x509.Certificate) *lint.LintResult {
-	if c.NotBefore.AddDate(0, 27, 0).Before(c.NotAfter) {
+	if util.BeforeOrOn(c.NotBefore.AddDate(0, 27, 0), c.NotAfter) {
```
robplee, line 50: Same comment here.  The lint description says the validity period must be "27 months... or less", a 27 month validity would therefore be acceptable and yet this change makes that an error case.

robplee (changes requested):
Something isn't quite lining up for me, either the lint descriptions aren't accurate representations of the BRs and EV rules or else this change makes the lints too strict.

mtgag:
The implementation in the pull request and jzlint was motivated by the following discussions/pull requests:

https://github.com/zmap/zlint/issues/467
https://github.com/zmap/zlint/pull/469

Certificate eeServerCertValidOver398.pem
(https://github.com/zmap/zlint/blob/master/v3/testdata/eeServerCertValidOver398.pem) from the testdata, as the name suggests, is valid over 398 days and the test  lint_e_server_cert_valid_time_longer_than_398_days_test.go treats it as an error for the corresponding lint.

Accordingly, a certificate which has a notBefore "Jan 1 00:00:00 2023" and a notAfter "Feb 1 00:00:00 2023" has a validity of two months, which I believe is correct when the granularity of the duration is expressed in months.

CBonnell:
The definition of Validity Period in the BRs was not aligned with the RFC 5280 definition until SC31. Changing the validity period calculation to align with 5280 for certificates issued prior to the SC31 effective date will yield false positive findings/errors.

Given this, I agree with @robplee that the changes proposed in this PR do not align with the BRs as written.

mtgag:
Ok. Then, let us keep the certificates and change the assertion in the test to have an explicit test for this behaviour (or simply close this PR). I will commit this soon.

zakird:
@CBonnell, are we in better shape after the latest change from @mtgag?

zakird (approved):

PR #718, opened 2023-05-23
Remove references in comments to Initialize() method of lints

toddgaunt-gs:
Hello Zlint team!

Just a small PR to update some comments that still refer to lints having an Initialize method. This appears to no longer be the case but a warning in the comments for RegisterLint, RegisterCertificateLint, and RegisterRevocationListLint was still referencing lints having such a method.

Regards,
Todd Gaunt


christopher-henderson (approved):
Thank you for the careful reading and the contribution!

PR #722, opened 2023-06-09
Bc critical

mtgag:
fatal (or error) may be more appropriate in this case. The Not Applicable status is covered by CheckApplies. When it enters Execute the certificate always has the BC extension and then only Pass or Error/Fatal seem appropriate for this lint.


v3/lints/rfc/lint_basic_constraints_not_critical.go
```diff
@@ -60,7 +60,6 @@ func (l *basicConstCrit) Execute(c *x509.Certificate) *lint.LintResult {
 		} else {
 			return &lint.LintResult{Status: lint.Error}
 		}
-	} else {
-		return &lint.LintResult{Status: lint.NA}
 	}
+	return &lint.LintResult{Status: lint.Fatal}
```
christopher-henderson, line 64:
```suggestion
	return &lint.LintResult{Status: lint.Error}
```

In general, `Fatal` denotes that something has gone wrong in the lint itself (parsing errors, type cast errors, etc.) although I'm sure that you can find an errant exception to this somewhere in the codebase.

christopher-henderson (changes requested):

v3/lints/rfc/lint_basic_constraints_not_critical.go
mtgag: Error is more appropriate. Maybe add an error message to distinguish between the two errors in the execute method? A commit for this follows soon.

christopher-henderson (approved):

PR #724, opened 2023-06-09
Max length check applies

mtgag:
Several lints about the maximum length in the value of several components of a subject are always triggered (CheckApplies returns true). It seems more appropriate to check if this component is present in the subject. Lint e_subject_dn_serial_number_max_length for example takes this approach,

(Addendum: Current max length lints check only the subject of a certificate, we could think of new lints checking also the issuer field)


christopher-henderson (approved):
Seems reasonable to me.

PR #729, opened 2023-06-29
Include LintMetadata in the LintResult

aaomidi:
This reduces the need to do a lookback on what the lint was based on its name when all the user needs is the metadata.

This also allows the name field in the results to be eventually sunset.


dadrian (commented):
I understand wanting this in code, but are we sure it needs to be output in JSON? Should it be a `json:"-"`?

aaomidi:
I didn't really think about if we want it in JSON or not when writing this and was following the rest of the fields. No opinions on it either way and I'm happy to exclude it from JSON serialization.

dadrian:
I think the JSON output would be extremely redundant when ran with ZCertificate if LintMetadata was included, so let's leave it out of the JSON.

dadrian (approved):

PR #732, opened 2023-07-05
E ext cert policy disallowed any policy qualifier refactor

mtgag:
Lint e_ext_cert_policy_disallowed_any_policy_qualifier checks whether the CpsOID or UserNoticeOID policy qualifiers are present. It does not evaluate whether anyPolicy is present. In this implementation it is iterated over all policies, checked whether a policy is the anyPolicy and then checked whether the qualifiers for this policy are the allowed ones. The implementation is based on code snippets from v3/util/oid.go GetMappedPolicies function. Also the check applies does not only check whether the extension is present but also whether one of the policies is the anyPolicy, and has a small TODO for discussion whether to implement this as a utility function "HasAnyPolicyOID". Also new certificates have been issued to test the behaviour of the lint.


v3/lints/rfc/lint_ext_cert_policy_disallowed_any_policy_qualifier.go
```diff
@@ -49,16 +57,113 @@ func NewUnrecommendedQualifier() lint.LintInterface {
...
+			if !policyQualifierId.Equal(util.CpsOID) && !policyQualifierId.Equal(util.UserNoticeOID) {
 				return &lint.LintResult{Status: lint.Error}
 			}
 		}
 	}
+
 	return &lint.LintResult{Status: lint.Pass}
 }
+
+func getCertificatePolicies(c *x509.Certificate) (error, []policyInformation) {
```
christopher-henderson, line 126: Errors typically come second in the return list.

```suggestion
func getCertificatePolicies(c *x509.Certificate) ( []policyInformation, error) {
```

christopher-henderson (approved):
> ...and has a small TODO for discussion whether to implement this as a utility function "HasAnyPolicyOID"

The code is indeed a little bit painful. But it is isolated. I think I'm okay merging this and then possibly taking a look at a cleanup on my own.

PR #738, opened 2023-08-20
Inclusion of approximately 190000 email protection certificates into the test corpus

christopher-henderson:
The ZLint project has begun writing lints for email protection certificates (https://github.com/zmap/zlint/pull/713) so it would appropriate to have a wide corpus for testing.

These certs were pulled from Censys using the following query.

```sql
SELECT fingerprint_sha256, raw FROM `censys-io.certificates_v2.certificates` 
WHERE (parsed.extensions.extended_key_usage.any = true 
or parsed.extensions.extended_key_usage.email_protection = true
or parsed.extensions.extended_key_usage is null)
and validation.nss.ever_valid = true
```

Aside from reformatting and sanitization, I have also gone through the task of deduplicating these certificates against the corpus because (as it turns out) there are about 1500 certs that are both email protection certs _and_ server auth certs.

### DIFF
--- a/v3/integration/config.json
+++ b/v3/integration/config.json
@@ -240,19 +240,100 @@
     {
       "Name": "xch.csv",
       "URL": "https://github.com/zmap/zlint-test-corpus/raw/master/certificates/xch.bz2"
+    },
+    {
+      "Name": "xch.csv",
+      "URL": "https://github.com/zmap/zlint-test-corpus/raw/master/certificates/xch.bz2"
+    },
+    {
+      "Name": "xde.csv",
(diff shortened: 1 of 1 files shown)


zakird (approved):

cardonator:
No concerns with this having been merged. I am curious if the corpus that was pulled from Census is compatible with the SMIME BRs that go into effect on Sept 1. Is there a good way to determine that?

cardonator:
@christopher-henderson logically I think what you have done is sound. The concern I have is that a significant corpus of these additions probably would not be comlpiant once zlint has a full set of SMIME checks. Perhaps that is what we would expect, but newly generated SMIME certs posted to CT post-Sept 1 will need to conform to the BR profiles where they don't currently need to.

PR #739, opened 2023-09-12
Part 1 of SC-62 related updates to zlint

cardonator:
These are two of the  more critical changes required by the SC-62 ballot that goes into effect on Sept 15th. The following changes were made in this PR:

* Updated lint for common name handling. The definition for the CN field has switched from deprecated to NOT RECOMMENDED (essentially SHOULD NOT). An IneffectiveDate was added to the original lint. [This was based on nomenclature adopted from https://datatracker.ietf.org/doc/html/rfc2119#section-4]
* Added a new lint for subscriber cert basic constraints checking. Post-SC62, basicConstraints MAY be included but MUST be critical if present.
* Added a date for SC62 Effective

I did want to get eyes on the approach here. We don't usually use Ballot numbers in the effective dates, however these changes present a unique challenge since we have one lint that is no longer effective while a new lint becomes effective; additionally SC-62 makes changes to both the TLS BRs and the EVGs. Perhaps I could change that to the version of the TLS BRs going into effect on that date anyway? paging @christopher-henderson for feedback.

### DIFF
--- a/v3/lints/cabf_br/lint_sub_cert_basic_constraints_not_critical.go
+++ b/v3/lints/cabf_br/lint_sub_cert_basic_constraints_not_critical.go
@@ -1,0 +1,61 @@
+package cabf_br
+
+/*
+ * ZLint Copyright 2023 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import (
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+type subCertBasicConstCrit struct{}
+
+/************************************************
+CA/Browser Forum BRs: 7.1.2.7.6 Subscriber Certificate Extensions
+
+| __Extension__                     | __Presence__    | __Critical__ | __Description__ |
+| ----                              | -               | -            | ----- |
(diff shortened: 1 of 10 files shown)


v3/lints/cabf_br/lint_sub_cert_basic_constraints_not_critical.go
```diff
@@ -0,0 +1,61 @@
...
+func (l *subCertBasicConstCrit) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && util.IsExtInCert(c, util.BasicConstOID)
+}
+
+func (l *subCertBasicConstCrit) Execute(c *x509.Certificate) *lint.LintResult {
+	if e := util.GetExtFromCert(c, util.BasicConstOID); e != nil {
+		if e.Critical {
+			return &lint.LintResult{Status: lint.Pass}
+		} else {
+			return &lint.LintResult{Status: lint.Error, Details: "Basic Constraints extension is present and marked as non-critical"}
```
christopher-henderson, line 57: Printing the OID of the offending constraint may help someone troubleshoot in the future.

```suggestion
			return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("Basic Constraints extension is present (%v) and marked as non-critical", e.Id)}
```

christopher-henderson (commented):
@cardonator 

> We don't usually use Ballot numbers in the effective dates...

I believe that the issue that we are seeing here is that we have an efficacy date for a requirement that does not line up with the efficacy date of its containing BR.

While it's not what is _usually_ done (indeed, most efficacy dates are tied to BR dates) there is precedence in [utils/time.go](https://github.com/zmap/zlint/blob/71d5e4b1cbf7636331bbf1c839d24abb4a945633/v3/util/time.go#L24) of efficacy dates being declared against very specific requirements.

So, so long as the name of the efficacy date is at least moderately descriptive of where the date came from (and it is), then I believe that what you've done here is perfectly fine.

Restated, I think it's most important that it's clear where the date came from for tracing purposes. That the given date does not lineup with a release, and is referenced in multiple documents, should be fine so long as it is traceable to this SC.

---

Aside from that, I had two tiny code nits. It otherwise looks great.

cardonator:
Thanks, I accepted both of those changes and agree with your reasoning. Appreciate the review!

christopher-henderson (approved):

PR #744, opened 2023-09-30
Lint for CABF SMIME 7.1.2.3.h - subjectAlternativeName, all: SHALL be present (7.1.2.3.h)

### LINKED ISSUE #743: Lint for CABF SMIME 7.1.2.3.h - subjectAlternativeName, all: SHALL be present

christopher-henderson:
Resolves #743. Addresses ` subjectAlternativeName, all: SHALL be present (7.1.2.3.h)` from #712

### DIFF
--- a/v3/lints/cabf_smime_br/lint_san_shall_be_present.go
+++ b/v3/lints/cabf_smime_br/lint_san_shall_be_present.go
@@ -1,0 +1,53 @@
+/*
+ * ZLint Copyright 2023 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package cabf_smime_br
+
+import (
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_san_shall_be_present",
+		Description:   "Subject alternative name SHALL be present",
+		Citation:      "7.1.2.3.h",
+		Source:        lint.CABFSMIMEBaselineRequirements,
+		EffectiveDate: util.CABF_SMIME_BRs_1_0_0_Date,
+		Lint:          NewSubjectAlternativeNameShallBePresent,
+	})
+}
+
+type SubjectAlternativeNameShallBePresent struct{}
+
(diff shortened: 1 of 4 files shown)


zakird (approved):

v3/lints/cabf_smime_br/lint_san_shall_be_present.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+type SubjectAlternativeNameShallBePresent struct{}
+
+func NewSubjectAlternativeNameShallBePresent() lint.LintInterface {
+	return &SubjectAlternativeNameShallBePresent{}
+}
+
+func (l *SubjectAlternativeNameShallBePresent) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c)
```
robplee, line 41: So, there's a bit of a circular dependency on when this lint might apply.  Currently, this lint applies to subscriber certificates with emailProtection EKU.  However, by section 1.1 of the SMIME BRs, a certificate isn't an SMIME certificate if it merely contains emailProtection and no subjectAltName extension.  So, by failing this lint, it means that it shouldn't apply...

Relevant quote: "An S/MIME Certificate for the purposes of this document can be identified by the existence of an Extended Key Usage (EKU) for id-kp-emailProtection (OID: 1.3.6.1.5.5.7.3.4) and the inclusion of a rfc822Name or an otherName of type id-on-SmtpUTF8Mailbox in the subjectAltName extension."

I guess one option is to refactor lint.base.go to remove the need for emailProtection on SMIME lints or to instead change it to require _either_ (emailProtection AND a SAN extension) OR an SMIME BR policy OID and returning N/A whenever neither of those conditions are met.

It's occurred to me that my #747 PR probably has the same issue as I've a lint requiring emailProtection but if it's not present then the lint would be N/A.  For simplicities sake, let's keep the discussion about this here and I'll put a comment on my PR to point to this one.

v3/lints/cabf_smime_br/lint_san_shall_be_present.go
```diff
@@ -0,0 +1,53 @@
...
+		Name:          "e_san_shall_be_present",
+		Description:   "Subject alternative name SHALL be present",
+		Citation:      "7.1.2.3.h",
+		Source:        lint.CABFSMIMEBaselineRequirements,
+		EffectiveDate: util.CABF_SMIME_BRs_1_0_0_Date,
+		Lint:          NewSubjectAlternativeNameShallBePresent,
+	})
+}
+
+type SubjectAlternativeNameShallBePresent struct{}
```
robplee, line 34: Does this need to be exported?  I'd have thought subjectAlternativeNameShallBePresent is fine

robplee (changes requested):
I think we need some changes for this PR and #713 re: checkApplies+CertificateLint.Execute.

v3/lints/cabf_smime_br/lint_san_shall_be_present.go
christopher-henderson: I believe this to be a separate issue and I have begun tracking it at #748.

---

That said, I am inclined to move forward with merging this (and other lints) as it implements `7.1.2.3.h` quite precisely

> subjectAlternativeName, all: SHALL be present

Any separate issues with regard to the general applicability of SMIME certificates from within the framework is out of scope of this change.

PR #746, opened 2023-09-30
Lint for CABF SMIME 7.1.2.3.h - subjectAlternativeName SHOULD NOT be marked critical unless the subject field is an empty sequence

### LINKED ISSUE #745: Lint for CABF SMIME 7.1.2.3.h - subjectAlternativeName SHOULD NOT be marked critical unless the subject field is an empty sequence
Sourced from #712 

> This extension SHOULD NOT be marked critical unless the subject field is an empty sequence.

christopher-henderson:
Resolves #745. Address `subjectAlternativeName, all: SHALL NOT be marked critical unless subject field is empty (7.1.2.3.h)` from #712.

Note that the SHALL NOT from the original tracker is incorrect. The full text of the section is...

> h. subjectAlternativeName (SHALL be present)
>
> This extension SHOULD NOT be marked critical unless the subject field is an empty sequence.
>
> The value of this extension SHALL be encoded as specified in Section 7.1.4.2.1.

### DIFF
--- a/v3/lints/cabf_smime_br/lint_san_should_not_be_critical.go
+++ b/v3/lints/cabf_smime_br/lint_san_should_not_be_critical.go
@@ -25,7 +25,7 @@ import (
 
 func init() {
 	lint.RegisterLint(&lint.Lint{
-		Name:          "e_san_should_not_be_critical",
+		Name:          "w_san_should_not_be_critical",
 		Description:   "subjectAlternativeName SHOULD NOT be marked critical unless the subject field is an empty sequence.",
 		Citation:      "7.1.2.3.h",
 		Source:        lint.CABFSMIMEBaselineRequirements,
@@ -53,20 +53,6 @@ func (l *SubjectAlternativeNameNotCritical) Execute(c *x509.Certificate) *lint.L
 		return &lint.LintResult{Status: lint.Pass}
 	} else if isCritical && !emptySubject {
 		// Critical, but there's a non-empty SAN.
-		return &lint.LintResult{
-			Status:  lint.Warn,
-			Details: "subject is not empty, but subjectAlternativeName is marked critical",
-		}
-	} else if !isCritical && emptySubject {
-		// Not critical, but there's an empty SAN.
-		//
-		// Is this a warning? That is, does the clause...
-		//
-		//     ...SHOULD NOT be marked critical unless the subject field is an empty sequence.
-		//
-		// ..also imply the inverse? That is...
-		//
-		//    ...SHOULD be marked critical if the subject field is an empty sequence.
 		return &lint.LintResult{
 			Status:  lint.Warn,
 			Details: "subject is not empty, but subjectAlternativeName is marked critical",
--- a/v3/lints/cabf_smime_br/lint_san_should_not_be_critical.go
+++ b/v3/lints/cabf_smime_br/lint_san_should_not_be_critical.go
@@ -1,0 +1,78 @@
+/*
+ * ZLint Copyright 2023 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


v3/lints/cabf_smime_br/lint_san_should_not_be_critical.go
```diff
@@ -0,0 +1,78 @@
...
+		// Is this a warning? That is, does the clause...
+		//
+		//     ...SHOULD NOT be marked critical unless the subject field is an empty sequence.
+		//
+		// ..also imply the inverse? That is...
+		//
+		//    ...SHOULD be marked critical if the subject field is an empty sequence.
+		return &lint.LintResult{
+			Status:  lint.Warn,
+			Details: "subject is not empty, but subjectAlternativeName is marked critical",
```
zakird, line 72: I think this should say "Subject is empty but SAN is not marked critical".

I also don't think that the text supports this second one being a warning based on the text of the standard. I think that it could be a separate informational lint if we wanted it to be.

v3/lints/cabf_smime_br/lint_san_should_not_be_critical.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zcrypto/x509/pkix"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "e_san_should_not_be_critical",
```
zakird, line 28: I think that this should be `w_` not `e_` since it returns a warning not an error.

zakird (changes requested):

CBonnell:
A few observations:

The SMIME BRs are not introducing any requirements above and beyond the requirements prescribed in RFC 5280. Section 4.1.2.6 says:

> If the subject field
   contains an empty sequence, then the issuing CA MUST include a
   subjectAltName extension that is marked as critical.  When including
   the subjectAltName extension in a certificate that has a non-empty
   subject distinguished name, conforming CAs SHOULD mark the
   subjectAltName extension as non-critical.


Given this, I believe this lint should be a RFC lint. Additionally, the status of the lint result returned upon encountering a non-critical SAN with empty subject name sequence should be an error.

cardonator:
@CBonnell that quoted text is from 4.2.1.6 but I'm guessing this similarly written section in 4.1.2.6 implies the same behavior.

> If subject
> naming information is present only in the subjectAltName extension
> (e.g., a key bound only to an email address or URI), then the subject
> name MUST be an empty sequence and the subjectAltName extension MUST
> be critical.

Does `lint_subject_empty_without_san` cover this scenario already?

christopher-henderson:
@CBonnell

> Additionally, the status of the lint result returned upon encountering a non-critical SAN with empty subject name sequence should be an error.

I believe that I agree with @zakird's interpretation that the branch should simply be removed from this lint altogether.

The branch does not address the clause `subjectAlternativeName SHOULD NOT be marked critical unless the subject field is an empty sequence` an any direct way. As such, any interpretation of the branch would be sourced from a separate document-or-clause which itself should have its own dedicated lint.

---

> Given this, I believe this lint should be a RFC lint.

Indeed, however the issue is irrespective as ZLint does not operate on the concept of "shared" lints between governing documents.

That is to say, even if two documents happen to contain the same wording, they still receive their own version of the given lint.

While it is indeed moderately redundant, it does drastically reduce the risk of a code change interfering with the interpretation of an unrelated governing document.

So we are going to need a lint for this requirement under the CABF SMIME umbrella, irrespective if we happen to have (or need) a doppelganger lint for 5280.

---
@cardonator
>  Does lint_subject_empty_without_san cover this scenario already?

([lint_subject_empty_without_san.go](https://github.com/zmap/zlint/blob/master/v3/lints/rfc/lint_subject_empty_without_san.go))

It at least does something similar, albeit without criticality testing.

v3/lints/cabf_smime_br/lint_san_should_not_be_critical.go
christopher-henderson: I agree that it should simply be a separate lint backed by its own citation, rather than an implied requirement of this citation.

cardonator:
That is an RFC lint at any rate, but I think I agree with you philosophically that if the SMIME BRs outline behavior, it makes sense for the lints designated for those standards to cover the described behavior, even if duplicative.

PR #752, opened 2023-10-15
CABF SMIME 7.1.4.2.h If present, the subject:emailAddress SHALL contain a single Mailbox Address

christopher-henderson:
Addresses #753 and accomplishes the following from #712

> subject:emailAddress, all: if present, the subject:emailAddress SHALL contain a single Mailbox Address. (7.1.4.2.2.h)

### DIFF
--- a/v3/lints/cabf_smime_br/lint_single_email_if_present.go
+++ b/v3/lints/cabf_smime_br/lint_single_email_if_present.go
@@ -1,0 +1,57 @@
+/*
+ * ZLint Copyright 2023 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


v3/lints/cabf_smime_br/lint_single_email_if_present.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+type singleEmailIfPresent struct{}
+
+func NewSingleEmailIfPresent() lint.LintInterface {
+	return &singleEmailIfPresent{}
+}
+
+func (l *singleEmailIfPresent) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && c.EmailAddresses != nil && len(c.EmailAddresses) != 0
```
zakird, line 43: Just want to check that this last part `len(c.EmailAddresses) != 0` is correct. It would seem that, if present, the field shouldn't be empty either. (I suspect though that this could just be an idiosyncrasy with Go X.509 and how fields are parsed)

christopher-henderson: > I suspect though that this could just be an idiosyncrasy with Go X.509 and how fields are parsed

I precisely wanted to avoid any such shenanigans being possible. I was uncertain what upstream systems would generate in a `not present` scenario. I reckoned that both `nil` and empty would count as `not being present` (generally speaking, rather than speaking to the specific Go implementation).

zakird (approved):

PR #760, opened 2023-11-01
Ensure AIA URLs point to public paths

cardonator:
As part of the SMIME work, I discovered at least one example of a cert that was issued using an outdated OCSP URL that would have been valid prior to the SMIME BRs. The rules around the AIA paths are not super strict but they are similar for both TLS and SMIME requirements, which read:
```
This extension MAY be present. If present, it MUST NOT be marked critical, and it MUST contain the
HTTP URL of the CA’s CRL service.
```
For publicly trusted certificates that are present in CT logs, these URLs are required to point to a resolvable location and shouldn't contain internal names. However, that requirement is not strictly codified in the requirements but is implied by what the resource is intended to point to. This lint intends to check and warn for this condition that may not be an actual problem, but probably is.

Note: I don't believe the Legacy lint is quite right here. The rule for Legacy reads
```
When provided, at least one accessMethod SHALL have the URI scheme HTTP. Other schemes (LDAP, FTP, ...) MAY be present.
```
The current lint tries to parse the url and make sure the DNS name has a valid TLD. Should I instead try to validate the scheme of the field before parsing in that lint only?

I also combined the Strict and Multipurpose lints into one as they have the same requirements. 

Feedback welcome and desired!

### DIFF
--- a/v3/lints/cabf_br/lint_sub_cert_aia_contains_internal_names.go
+++ b/v3/lints/cabf_br/lint_sub_cert_aia_contains_internal_names.go
@@ -58,7 +58,7 @@ func (l *subCertAIAInternalName) Execute(c *x509.Certificate) *lint.LintResult {
 	for _, u := range c.OCSPServer {
 		purl, err := url.Parse(u)
 		if err != nil {
-			return &lint.LintResult{Status: lint.Fatal}
+			return &lint.LintResult{Status: lint.Error}
 		}
 		if !util.HasValidTLD(purl.Hostname(), time.Now()) {
 			return &lint.LintResult{Status: lint.Warn}
@@ -67,7 +67,7 @@ func (l *subCertAIAInternalName) Execute(c *x509.Certificate) *lint.LintResult {
 	for _, u := range c.IssuingCertificateURL {
 		purl, err := url.Parse(u)
 		if err != nil {
-			return &lint.LintResult{Status: lint.Fatal}
+			return &lint.LintResult{Status: lint.Error}
 		}
 		if !util.HasValidTLD(purl.Hostname(), time.Now()) {
 			return &lint.LintResult{Status: lint.Warn}
--- a/v3/lints/cabf_br/lint_sub_cert_aia_contains_internal_names.go
+++ b/v3/lints/cabf_br/lint_sub_cert_aia_contains_internal_names.go
@@ -1,0 +1,77 @@
+package cabf_br
+
+/*
+ * ZLint Copyright 2023 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import (
+	"net/url"
+	"time"
+
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+type subCertAIAInternalName struct{}
+
+/************************************************************************
+BRs: 7.1.2.10.3
+CA Certificate Authority Information Access
+This extension MAY be present. If present, it MUST NOT be marked critical, and it MUST contain the
+HTTP URL of the CA’s CRL service.
+
+id-ad-ocsp        A HTTP URL of the Issuing CA's OCSP responder.
+id-ad-caIssuers   A HTTP URL of the Issuing CA's Certificate.
+*************************************************************************/
+
+func init() {
+	lint.RegisterLint(&lint.Lint{
+		Name:          "w_sub_cert_aia_contains_internal_names",
+		Description:   "Subscriber certificates authorityInformationAccess extension should contain the HTTP URL of the issuing CA’s certificate, for public certificates this should not be an internal name",
+		Citation:      "BRs: 7.1.2.10.3",
+		Source:        lint.CABFBaselineRequirements,
+		EffectiveDate: util.CABEffectiveDate,
+		Lint:          NewSubCertAIAInternalName,
+	})
+}
+
+func NewSubCertAIAInternalName() lint.LintInterface {
+	return &subCertAIAInternalName{}
+}
+
+func (l *subCertAIAInternalName) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c)
+}
+
+func (l *subCertAIAInternalName) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, u := range c.OCSPServer {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
+		}
+		if !util.HasValidTLD(purl.Host, time.Now()) {
+			return &lint.LintResult{Status: lint.Warn}
+		}
+	}
+	for _, u := range c.IssuingCertificateURL {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
+		}
+		if !util.HasValidTLD(purl.Host, time.Now()) {
+			return &lint.LintResult{Status: lint.Warn}
+		}
+	}
+	return &lint.LintResult{Status: lint.Pass}
+}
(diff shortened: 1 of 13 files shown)


cardonator:
@christopher-henderson this is ready for review. I can't add you as a reviewer myself.

v3/lints/cabf_br/lint_sub_cert_aia_contains_internal_names.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (l *subCertAIAInternalName) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c)
+}
+
+func (l *subCertAIAInternalName) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, u := range c.OCSPServer {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
```
christopher-henderson, line 61: I believe that what we're doing here is saying that if the URL is parseable and has a valid TLD, then it is probably resolvable (which is more tolerable than _actually_ doing an DNS lookup).

So just a minor clerical thing that `lint.Fatal` is typically used to signal an internal error (although you can find discrepancies...) so this parsing failure would be an example of a `lint.Error`.

```suggestion
			return &lint.LintResult{Status: lint.Error}
```

v3/lints/cabf_br/lint_sub_cert_aia_contains_internal_names.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		if !util.HasValidTLD(purl.Hostname(), time.Now()) {
+			return &lint.LintResult{Status: lint.Warn}
+		}
+	}
+	for _, u := range c.IssuingCertificateURL {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
```
christopher-henderson, line 70:
```suggestion
			return &lint.LintResult{Status: lint.Error}
```

v3/lints/cabf_smime_br/lint_legacy_aia_contains_internal_names.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (l *smimeLegacyAIAContainsInternalNames) CheckApplies(c *x509.Certificate) bool {
+	return util.IsLegacySMIMECertificate(c)
+}
+
+func (l *smimeLegacyAIAContainsInternalNames) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, u := range c.OCSPServer {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
```
christopher-henderson, line 63:
```suggestion
			return &lint.LintResult{Status: lint.Error}
```

v3/lints/cabf_smime_br/lint_legacy_aia_contains_internal_names.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		if !util.HasValidTLD(purl.Hostname(), time.Now()) {
+			return &lint.LintResult{Status: lint.Warn}
+		}
+	}
+	for _, u := range c.IssuingCertificateURL {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
```
christopher-henderson, line 72:
```suggestion
			return &lint.LintResult{Status: lint.Error}
```

v3/lints/cabf_smime_br/lint_legacy_aia_contains_internal_names.go
```diff
@@ -0,0 +1,79 @@
...
+
+func NewSMIMELegacyAIAInternalName() lint.LintInterface {
+	return &smimeLegacyAIAContainsInternalNames{}
+}
+
+func (l *smimeLegacyAIAContainsInternalNames) CheckApplies(c *x509.Certificate) bool {
+	return util.IsLegacySMIMECertificate(c)
+}
+
+func (l *smimeLegacyAIAContainsInternalNames) Execute(c *x509.Certificate) *lint.LintResult {
```
christopher-henderson, line 59: > ...at least one accessMethod SHALL have the URI scheme HTTP

I reckon that we can also enforce this clause a bit more closely by checking up on the parse scheme

```go
atLeastOneHttp := false
...
...
if purl.Scheme == "http" {
   atLeastOneHttp = true
}
...
...
if !atLeastOneHttp {
}
```

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (l *smimeStrictAIAContainsInternalNames) CheckApplies(c *x509.Certificate) bool {
+	return util.IsStrictSMIMECertificate(c) || util.IsMultipurposeSMIMECertificate(c)
+}
+
+func (l *smimeStrictAIAContainsInternalNames) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, u := range c.OCSPServer {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
```
christopher-henderson, line 63:
```suggestion
			return &lint.LintResult{Status: lint.Error}
```

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		if !util.HasValidTLD(purl.Hostname(), time.Now()) {
+			return &lint.LintResult{Status: lint.Warn}
+		}
+	}
+	for _, u := range c.IssuingCertificateURL {
+		purl, err := url.Parse(u)
+		if err != nil {
+			return &lint.LintResult{Status: lint.Fatal}
```
christopher-henderson, line 72:
```suggestion
			return &lint.LintResult{Status: lint.Error}
```

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names.go
```diff
@@ -0,0 +1,79 @@
...
+
+func NewSMIMEStrictAIAInternalName() lint.LintInterface {
+	return &smimeStrictAIAContainsInternalNames{}
+}
+
+func (l *smimeStrictAIAContainsInternalNames) CheckApplies(c *x509.Certificate) bool {
+	return util.IsStrictSMIMECertificate(c) || util.IsMultipurposeSMIMECertificate(c)
+}
+
+func (l *smimeStrictAIAContainsInternalNames) Execute(c *x509.Certificate) *lint.LintResult {
```
christopher-henderson, line 59: Much like the legacy lint, but here we can fail fast on the `every accessMethod SHALL have the URI scheme HTTP` clause.

```go
if purl.Scheme != "http" {
}
```

christopher-henderson (changes requested):
Thank you for the lints! Looks like a pretty straight forward change.

> The current lint tries to parse the url and make sure the DNS name has a valid TLD. Should I instead try to validate the scheme of the field before parsing in that lint only?

I think we can let the URL parser do its thing as it will parse out the scheme for us. If there is no scheme, then that is a a problem in-and-of-itself that the parser will report to us (that is, other schemes are allowed however I doubt that not _any_ scheme attached is okay).

v3/lints/cabf_smime_br/lint_legacy_aia_contains_internal_names.go
cardonator: I implemented my take on this one, but I became unsure about it so if you can take a look that would be helpful. Gist is, in the requirement is mentions that at least one accessMethod MUST be http, however I believe it applies independently to the OCSPServers and IssuingCAs. If you think I'm being overly aggressive here, let me know.

cardonator:
Thanks for the review @christopher-henderson I had one question above but otherwise should be ready to re-review. 👍

v3/lints/cabf_smime_br/lint_legacy_aia_contains_internal_names.go
christopher-henderson: hrmmm I think that my only thought is that this would also require that there be a minimum of one at least OCSP and at least one CRL (that is, if the loop never enters then `atLeastOnHttp` will be `false`).

Regardless, if there is such a requirement (at least one OCSP and at least one CRL) then it should be a separate lint (trying to encode all surrounding facts into each lint is how they commonly end up in deadlock).

So I guess we could just check the empty condition before failing out.

```go
	if !atLeastOneHttp && len(c.OCSPServer) != 0 {
		return &lint.LintResult{Status: lint.Error, Details: "at least one accessMethod MUST have the URI scheme HTTP"}
	}
```

cardonator: Okay, I agree and made that change. Do you assume that the same change is not needed for the IssuingCertificateURL list?

christopher-henderson: If you extend this small change to CRLs as well then I think we'd be `gtg` on these!

cardonator: done

christopher-henderson (approved):

PR #761, opened 2023-11-04
CABF SMIME lints failed to be registered in the CLI due to a missing blank import

### LINKED ISSUE #759: Regression in SMIME lint dispatch from the CLI
Refers to discussion in #757

christopher-henderson:
Thank you for your patience @cardonator 

The infrastructure for a lint to register itself leans on Go's magical `init` function which is automatically executed at program startup.

However, the `init` function is not executed if the package is not referred to by the main program, which is why `zlint.go` features a short list of imports performed solely for the side effect of triggering these `init` functions.

Unfortunately, we missed this import when building the CABF SMIME framework. The cause is that almost all of the testing, tooling, and integration tests are focused on either the library usage of Zlint or on testing lints against the corpus of certs scraped from the wild. Such a CLI integration test would be a valuable thing to consider spending time on, however I believe it is out of scope of this particularly pernicious issue.

Resolves #759

### DIFF
--- a/v3/lint/source.go
+++ b/v3/lint/source.go
@@ -51,7 +51,7 @@ func (s *LintSource) UnmarshalJSON(data []byte) error {
 	}
 
 	switch LintSource(throwAway) {
-	case RFC5280, RFC5480, RFC5891, CABFBaselineRequirements, CABFEVGuidelines, MozillaRootStorePolicy, AppleRootStorePolicy, Community, EtsiEsi:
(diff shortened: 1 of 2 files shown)


cardonator:
Nice, thanks! I had almost fixed this myself before bringing it up but I missed the import in zlint.go. Good to know about that detail!

zakird (approved):

Issue #748, opened 2023-10-08
Revisit S/MIME applicability rules


christopher-henderson:
Stemming from the comment at https://github.com/zmap/zlint/pull/744#discussion_r1347031645 we may want to revisit the logic at [IsEmailProtectionCert](https://github.com/zmap/zlint/blob/386a8dc413add9bb92d80badbb4d86f833f6a4e5/v3/util/ca.go#L70)

> An S/MIME Certificate for the purposes of this document can be identified by the existence of an Extended Key Usage (EKU) for id-kp-emailProtection (OID: 1.3.6.1.5.5.7.3.4) and the inclusion of a rfc822Name or an otherName of type id-on-SmtpUTF8Mailbox in the subjectAltName extension.

robplee:
So, my proposal to address this issue is to change this line: https://github.com/zmap/zlint/blob/master/v3/lint/base.go#L224

To something like this:
`
if l.Source == CABFSMIMEBaselineRequirements && !((util.IsEmailProtectionCert(cert) && util.HasEmailSAN(cert) || util.IsSMIMEBRCertificate(cert)) {
`

Where HasEmailSAN (name TBC) checks for the presence of at least one san:rfc822Name or a san:otherName of type id-on-SmtpUtf8Mailbox

PR #764, opened 2023-11-09
Revisit S/MIME applicability rules

robplee:
PR to address #748, I had to regenerate a couple of certificates but I don't think any of the changes I made to them substantially changes the effectiveness of the unit tests using them.


christopher-henderson (approved):
Thank you! This is a great change.

Issue #765, opened 2023-11-10
Gauging interest in doing a large-scale code change migrating everything from Lints -> CertificateLints


aaomidi:
In https://github.com/zmap/zlint/pull/699 I introduced the ability to have more than just certificate lints. I now want to finish off this migration by moving all the existing lints to CertificateLints.

The next step will be fully removing the `Lint` interface, which is potentially a more contentious move since it might break lints folks have already made.

Do we have any concerns with doing the first part of this?

zakird:
Makes sense to me.

aaomidi:
Agreed on the last count. I've written a small tool to convert all the lints to CertificateLint in this project. I can imagine that being useful for folks who have built their own lints separately. 

The decision to remove Lint/RegisterLint will be up to the maintainers here.

When I introduced the CertificateLint/RevocationListLint, generics wasn't officially "ready". Maybe before any long term decision is made, we can potentially look into generics for these lints again too?

christopher-henderson:
Migration from `lint.Lint` to `lint.CertificateLint` via your tooling is fairly straight forward, and (as you've said) the real issue is how/when/if to remove the `lint.Lint` interface altogether.

> ...we can potentially look into generics for these lints again too?

Presuming that that this would be considered during the removal of `lint.Lint` then it makes sense as it is certainly a breaking change for library consumers. So bundling the consequences together would help reduce friction overall.

PR #766, opened 2023-11-11
Convert all Lint to CertificateLint

aaomidi:
Solves #765 

Generation code: https://github.com/aaomidi/zlint-migration

The formatting isn't...what I exactly wanted it to be. I'm still trying to see if I can go directly to a well formatted struct and not have to run gofmpt on everything.

I've asked about the formatting issue: https://groups.google.com/g/golang-nuts/c/C16T4H7slA4


aaomidi:
@zakird, I'm curious since you're a major maintainer here. Would you accept the formatting of the RegisterCertificateLint methods as is, or should I try to spend some more time getting the formatting just right?

Also just curious, how do you feel about the fact that I ended up running gofumpt on all of these lints?

aaomidi:
Well, I actually think I have a correct solution in #767 now

PR #699, opened 2022-11-30
Add CRL linting infrastructure

aaomidi:
ZLint is an invaluable asset for pre-linting certificate issuances and creating a technical enforcement of the various rules around certificate issuance. The recent incidents [1], [2] involving CRLs have created a need to extend linting to CRLs as well.

This PR aims to bring CRL linting and create the necessary code infrastructure to address #458 to add OCSP linting support to ZLint without having to do any large modifications to the codebase.

The major requirement with this PR is to not unintentionally break the existing lints. The decisions made in this PR are generally a reflection of this goal. The major highlights of these changes are as follows:

* Introduce `CertificateLintInterface` and `RevocationListLintInterface`. Deprecate the existing `LintInterface`.
* Introduce `CertificateLint` and `RevocationListLint`. Deprecate the existing `Lint` struct.
   * As part of this, introduce `LintMeta` to reduce repeating code between the two new lints.
   * Create an adapter to go from the deprecated `Lint` to `CertificateLint` and vice versa.
* Introduce `LinterLookup` as a generic interface that acts as the storage backend for the different types of lints.
   * The global registry's ByName and BySource methods have now been deprecated, and these methods are now scoped down to the specific stores.
* Registering a `Lint` is now deprecated, and it is recommended to use RegisterCertificateLint and RegisterRevocationListLint for Certificates and CRLs respectively.

Some other notes:
* Currently [zcrypto] does not have support for RevocationLists. A PR will be made to add support for that. In the meantime, ox509 is used to denote "original x509", which has the RevocationList struct. 
   * Before this PR is submitted, an update to zcrypto should happen so I can replace ox509 with x509.
      * [I've started on a potential draft for this update in my fork of zcrypto][zcrypto-patch], although I feel like we should probably merge from upstream and re-apply the zcrypto patches instead of trying to make this work
* Maintainers, please comment on how the deprecation of these methods should happen. For now I've added `@deprecated` to the godoc description of these methods.

Future work:
* In a future PR, I will migrate all the existing `Lint`s to be `CertificateLint` instead. This migration is only necessary due to the introduction of `LintMeta`.
* In a future PR, I will add all the lints that [Boulder's CRL Linting] currently has.
* Adding OCSP linting is not in-scope, and I probably won't get around to doing it.

[1]: https://bugzilla.mozilla.org/show_bug.cgi?id=1793467
[2]: https://bugzilla.mozilla.org/show_bug.cgi?id=1793642
[zcrypto]: https://github.com/zlint/zcrypto
[...]

PR #767, opened 2023-11-12
[LSC] Convert all Lints to CertificateLints

aaomidi:
Partially addresses #765 

Using the code on: https://github.com/aaomidi/zlint-migration, I've converted all `lint.RegisterLint` to `lint.RegisterCertificateLint`. 

To make it easier for reviewers: look at the code here: https://github.com/aaomidi/zlint-migration, re-run this on the lints, it should create no diff compared to this PR.

Note that I've reviewed every file manually as well just in case, and I didn't spot anything.

### DIFF
--- a/v3/lints/apple/lint_ct_sct_policy_count_unsatisfied.go
+++ b/v3/lints/apple/lint_ct_sct_policy_count_unsatisfied.go
@@ -27,13 +27,15 @@ import (
 type sctPolicyCount struct{}
 
 func init() {
-	lint.RegisterLint(&lint.Lint{
-		Name:          "w_ct_sct_policy_count_unsatisfied",
-		Description:   "Check if certificate has enough embedded SCTs to meet Apple CT Policy",
-		Citation:      "https://support.apple.com/en-us/HT205280",
-		Source:        lint.AppleRootStorePolicy,
-		EffectiveDate: util.AppleCTPolicyDate,
-		Lint:          NewSctPolicyCount,
+	lint.RegisterCertificateLint(&lint.CertificateLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "w_ct_sct_policy_count_unsatisfied",
+			Description:   "Check if certificate has enough embedded SCTs to meet Apple CT Policy",
+			Citation:      "https://support.apple.com/en-us/HT205280",
+			Source:        lint.AppleRootStorePolicy,
+			EffectiveDate: util.AppleCTPolicyDate,
+		},
+		Lint: NewSctPolicyCount,
 	})
 }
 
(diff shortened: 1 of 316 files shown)


robplee:
I'm not sure _how_ one would do this exactly, but I wonder if we could/should add something to the build process to try to make it impossible to add _new_ `lint.Lint`s or `lint.LintInterface`s in PRs?  Something on a similar line to Chris' magic that blocks people from merging changes to genTestCerts.go but that works everywhere?

I say this because I'm pretty sure you're going to have to fix some lints with this PR that I've had merged in since #699 was merged so I think there might be value in having a good solution that isn't "We'll just review it better" because there's enough stuff to check with zlint PRs to "just" make sure they're accurately implementing the rule they're adding.  This feels like something we probably can check automatically so we probably should try.

aaomidi:
I can actually extend the code I wrote to actively look for that in the CI step (?)

aaomidi:
Are we waiting on anything else for this PR?

cardonator:
I meant to try to use these changes in our build environment to make sure they wouldn't have any immediate effect on our usage. I was wondering if this change is significant enough that we should consider incrementing the version. Not sure if there is anything else blocking.

aaomidi:
I _think_ this change should be invisible for most folk. Good point though it makes sense to leave it ready for a few folks to test it out on their environment.

Also if this change does accidentally break anyone, we should definitely roll it back.

aaomidi:
@robplee: I have this PR (which is branched off of master, so it'll fail for now) #770.

cardonator:
Alright, I did some high level testing against this PR in our integration and confirmed that there were no issues for our use case. Ours is definitely not super complicated. I think the second step you described above would definitely break us so we should probably make sure we can cut a new release prior to that to give time for people to migrate over to the new interfaces. 👍

aaomidi:
Thank you so much for testing this change on your end!

zakird:
This looks reasonable to me. @christopher-henderson would you also be willing to take a look?

christopher-henderson:
> I'm not sure how one would do this exactly, but I wonder if we could/should add something to the build process to try to make it impossible to add new lint.Lints or lint.LintInterfaces in PRs? Something on a similar line to Chris' magic that blocks people from merging changes to genTestCerts.go but that works everywhere?

Indeed, doing so is possible (it parses the AST of lints and does something similar to what @aaomidi's code migrator is doing).

I'll be able to take this on (pending the conversation on implementing the breaking changes), as well as some repo cleanup activities regarding lint templating being slightly off.

christopher-henderson (approved):
Thank you @aaomidi for automating this transition!

PR #768, opened 2023-11-13
 CABF SMIME BR Appendix A A.1 - countryName matches registration scheme id

eliot-gs:
Adding lint to check the 2 character countryName in the subject matches the countryName in the registration scheme id as per Appendix A A.1: _The country code used in the Registration Scheme identifier SHALL match that of the subject:countryName in the Certificate as specified in Section 7.1.4.2.2._


v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Regex to match the start of an organization identifier: 3 character registration scheme identifier and 2 character ISO 3166 country code
+var countryRegex = regexp.MustCompile(`^([A-Z]{3})([A-Z]{2})`)
+
+func init() {
+	lint.RegisterCertificateLint(&lint.CertificateLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "e_registration_scheme_id_matches_subject_country",
+			Description:   "The country code used in the Registration Scheme identifier SHALL match that of the subject:countryName in the Certificate as specified in Section 7.1.4.2.2",
+			Citation:      "Appendix A A.1",
```
robplee, line 34:
```suggestion
			Citation:      "Appendix A.1",
```

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
```diff
@@ -0,0 +1,94 @@
...
+			EffectiveDate: util.CABF_SMIME_BRs_1_0_0_Date,
+		},
+		Lint: NewRegistrationSchemeIDMatchesSubjectCountry,
+	})
+}
+
+type registrationSchemeIDMatchesSubjectCountry struct{}
+
+// NewRegistrationSchemeIDMatchesSubjectCountry creates a new linter to enforce SHALL requirements for registration scheme identifiers matching subject:countryName
+func NewRegistrationSchemeIDMatchesSubjectCountry() lint.LintInterface {
```
robplee, line 45:
```suggestion
func NewRegistrationSchemeIDMatchesSubjectCountry() lint.CertificateLintInterface {
```

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
```diff
@@ -0,0 +1,94 @@
...
+    return &registrationSchemeIDMatchesSubjectCountry{}
+}
+
+// CheckApplies returns true if the provided certificate contains subject:countryName and one-or-more of the following SMIME BR policy identifiers:
```
robplee, line 49: Firstly, this comment doesn't mention the update to need a _partially valid_ subject.organizationID.

Secondly, I wonder if you could compress the list of policies by just saying "and the certificate contains an Organization or Sponsor Validated policy OID"

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
```diff
@@ -0,0 +1,94 @@
...
+//   - Organization Validated Strict
+//   - Organization Validated Multipurpose
+//   - Organization Validated Legacy
+//   - Sponsor Validated Strict
+//   - Sponsor Validated Multipurpose
+//   - Sponsor Validated Legacy
+func (l *registrationSchemeIDMatchesSubjectCountry) CheckApplies(c *x509.Certificate) bool {
+	if c.Subject.Country == nil {
```
robplee, line 57: It's a bit picky but I wonder if this would be better if you also checked that there was a non-empty-string value (or make sure that there's at least one value that's two characters long?) in c.Subject.Country?

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
```diff
@@ -0,0 +1,94 @@
...
+			return false
+		}
+	}
+
+	return util.IsOrganizationValidatedCertificate(c) || util.IsSponsorValidatedCertificate(c)
+}
+
+// Execute applies the requirements on matching subject:countryName with registration scheme identifiers
+func (l *registrationSchemeIDMatchesSubjectCountry) Execute(c *x509.Certificate) *lint.LintResult {
+	country := c.Subject.Country[0]
```
robplee, line 73: I wonder how well this combines with the CheckApplies.  I have got a comment in this review suggesting you loop to find one countryName of length 2 but I wonder if you should be saying this lint applies if something like this is true:
`c.Subject.Country[0] == <TWO CAPITAL LETTERS>`
Given that the Execute function of this lint is assuming that the first countryName value is present and valid.

robplee (changes requested):
Nothing super major, but a couple of nits to address.

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
eliot-gs: Updated

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
eliot-gs: Should have seen LintInterface was deprecated. Updated as suggested.

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
eliot-gs: Made comment more relevant to full content of CheckApplies

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
eliot-gs: I can check if c.Subject.Country is of length 2 and if not return false.

eliot-gs: Added subsequent check that c.Subject.Country[0] is 2 characters in length because as you raised in a different comment, the Execute function of this lint is assuming that the first countryName value is present and (at least could be) valid - e.g. ISO 3166 or XX

robplee (approved):

robplee:
Sorry about this @eliot-gs, but there's currently some discussion about removing this requirement to have the country in the organizationIdentifier match the provided countryName attribute (will link the first email on the subject below).  I wonder if we should wait for some additional clarity before merging this PR?

I guess on the flip-side though, the change hasn't been made and as it's a substantive one I imagine it will be brought in in the future making this lint useful for certs issued since 20230901 (yyyymmdd).  So, perhaps this _could_ be merged with the caveat that we may need to add an ineffective date once there's a date for when any future changes are coming into force?

SMIME WG email: https://lists.cabforum.org/pipermail/smcwg-public/2023-November/000863.html

robplee:
Hi @eliot-gs, I've been thinking about this PR a bit and I think we can merge it in.  There is a decent chance that the rule you're enforcing with your lint is going to be removed, but that change is a substantive change to the BRs and should be made effective from a future date.  Therefore, the requirement from appendix A1 is effective from 1st September 2023 until the TBC future date.

My thinking is we should merge this in, keep our eyes open on the conversations around SMCWG and once that conversation is resolved we can do another PR to set the ineffective date for the lint in this PR.

What do you think @christopher-henderson?

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
```diff
@@ -0,0 +1,92 @@
...
+		return false
+	}
+
+	if len(c.Subject.Country[0]) != 2 {
+		return false
+	}
+
+	for _, id := range c.Subject.OrganizationIDs {
+		submatches := countryRegex.FindStringSubmatch(id)
+		if len(submatches) < 3 {
```
christopher-henderson, line 61: Thankfully `len(nil)` won't explode here :smile:  (I had to double check this myself, it returns 0 which makes sense).

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
```diff
@@ -0,0 +1,92 @@
...
+			return &lint.LintResult{Status: lint.Error, Details: err.Error()}
+		}
+	}
+	return &lint.LintResult{Status: lint.Pass}
+}
+
+// verifySMIMEOrganizationIdentifierContainSubjectNameCountry verifies that the country code used in the subject:organizationIdentifier matches subject:countryName
+func verifySMIMEOrganizationIdentifierContainsSubjectNameCountry(id string, country string) error {
+	submatches := countryRegex.FindStringSubmatch(id)
+	// Captures the country code from the organization identifier
```
christopher-henderson, line 84:
```suggestion
	// Captures the country code from the organization identifier
	// Note that this raw indexing into the second position is only safe
	// due to a length check done in CheckApplies
```

I often find it useful for the poor unfortunate souls in the future to leave a note as to why an unchecked index/dereference is safe, just in case those safety guarantees are broken by someone in the future and now the cause has to be hunted down.

christopher-henderson (commented):
Thank you for your thorough policy reviews @robplee and thank you the lint @eliot-gs!

I had one minor request on a code comment, however it appears that a file needs to be go fmted before we can merge.

v3/lints/cabf_smime_br/lint_registration_scheme_id_matches_subject_country.go
eliot-gs: I have added this suggested comment.

eliot-gs:
Thanks @christopher-henderson, I have addressed the issues you raised.

chrisbn:
I'd like to highlight that the current requirement effectively prevents the use of LEI and INT as registration schemes. 
As per 7.1.4.2.2 d note 2 and Appendix A A.1, the country code shall be set to XG (example INTXG, LEIXG).

The Subject:CountryName SHALL contain the country code associated with the **location** of the Subject verified, which cannot be XG, and so this lint would fail.

Also, the existing lint e_subject_country_not_iso does not recognize XG as a valid country, so even if the Subject:CountryName would be set to XG, this lint would block it.

robplee:
Hi @chrisbn, thanks for the comment however the point you're raising is not a zlint problem.  zlint exists to apply the rules *as they are written* and, to me, your comment sounds like the SMIME BRs are broken because of the inconsistency between the parts of the document you've highlighted.  As you've pointed out, the 7.1.4.2.2 d note 2 text says that in some cases the country code in the organizationIdentifier should be "XG", Appendix A.1 says the country codes in the organizationIdentifier and countryName must match, and 7.1.4.2.2 n doesn't support the countryName being set to "XG".  This is a failure of the SMIME BRs and should be fixed there before any changes to zlint are made.

tl;dr
We don't write the rules here, we just enforce them :)

PR #770, opened 2023-11-16
Detect if the deprecated RegisterLint method was used

aaomidi:


aaomidi:
@christopher-henderson I actually have the detection code here, but feel free to go with something else for this too!

PR #777, opened 2023-12-06
refactor of SMIME aia contains

mtgag:
the effective date of this lint has been updated and in checkApplies a check to see whether the extension is present has been added. 

Also, another proposal would be to break this lint into two lints. One that is checking whether the scheme is HTTP and produces either error or pass and a second lint that is checking the internal names and produces either warn or pass. I believe that most lints try to focus on one aspect and not to lint many aspects in one lint. I would be willing to contribute on this.

Best,

Vangelis


h
zakird, line 1: I think that this file may have been accidentally committed.

mtgag: thank you. Just removed.
Vangelis

cardonator:
Separating the scheme check to a separate lint does seem appropriate. There are several lints for AIA checking that have a similar structure so we will want to make sure they all have consistent behavior.

mtgag:
I did not find any text in the SMIME CABF Specifications which deals with internal names in AIA/OCSP. I propose to remove this part in the lint and the smime package and if this is a requirement from another specification, implement it as a new lint in the corresponding package. This proposal is implemented in commit https://github.com/zmap/zlint/pull/777/commits/a1eee503e9c2219408699bab3ed8cb98b015032a

cardonator:
I'll disagree with that one. I don't think a requirement is spelled out in any of the BRs, however the ocsp has to be publicly accessible as does the aia. Is there a use case you're thinking of where an internal name would or should be allowed? Pinging @CBonnell for visibility.

CBonnell:
I agree with @cardonator. While the S/MIME BRs do not explicitly state that the URIs contained within the CRLDP and AIA extensions must be publicly accessible, the language cannot reasonably be interpreted as allowing the inclusion of such internal domain names.

If one reads the current language in the SMBRs as permitting internal domain names in CRLDP and AIA extensions, then one would also interpret the SMBRs as allowing the issuance of publicly trusted  S/MIME certificates that have no publicly available revocation information.

More generally, the inclusion of such internal domain names is of no use (and actively harmful due to connection attempts that will inevitably fail) in the common case, as the referenced resources cannot be accessed by most Relying Parties.

mtgag:
The motivation to remove the check about internal names was to keep different aspects of lints separated and left the removed part to future work.

Commit https://github.com/zmap/zlint/pull/777/commits/2a6b8870db1235774d8b869f960955a7299cdc90 tries to address the issues discussed in this PR. On the one hand two distinct lints exist to keep a narrow focus on each lint, and on the other hand the check for internal names is extended to all SMIME certificates. I hope that the current proposal and implementation reflects the discussion.

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names.go
```diff
@@ -34,39 +35,40 @@ values for each of the following types:
...
 
-For Strict and Multipurpose: When provided, every accessMethod SHALL have the URI scheme HTTP. Other schemes SHALL NOT be present.
 *************************************************************************/
 
 func init() {
 	lint.RegisterCertificateLint(&lint.CertificateLint{
 		LintMetadata: lint.LintMetadata{
-			Name:          "w_smime_strict_aia_contains_internal_names",
-			Description:   "SMIME Strict certificates authorityInformationAccess When provided, every accessMethod SHALL have the URI scheme HTTP. Other schemes SHALL NOT be present.",
+			Name:          "w_smime_aia_contains_internal_names",
```
cardonator, line 43: The filename of this lint and test should be renamed to match this updated lint name.

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names_test.go
```diff
@@ -19,14 +19,29 @@ func TestSMIMEStrictAIAInternalName(t *testing.T) {
...
 			InputFilename:  "smime/aiaWithInternalNamesStrict.pem",
 			ExpectedResult: lint.Warn,
 		},
+		{
+			Name:           "warn - aia with internal names in AIA CA issuers ",
+			InputFilename:  "smime/aiaWithInternalNamesCaIssuersStrict.pem",
+			ExpectedResult: lint.Warn,
+		},
+		{
+			Name:           "warn - aia with valid names, one is ldap",
```
cardonator, line 32: I believe this version of the lint will now cover Legacy profile, which currently allows LDAP scheme to appear in the AIA. That may be unintentional or just missed. Along with that, I think if we're going to do this refactor for the Strict/MP profiles, that we should also do it for the legacy profiles which are additional lints that should be reviewed.

cardonator (changes requested):

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names.go
mtgag: renamed

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names_test.go
mtgag: In the contains_internal_names lint, it is checked whether it is a BR-SMIME certificate, the scheme part is not checked at all and only the host part of the URL is checked. The certificate containing the LDAP URL is a Pass Test. I believe the current implementation covers these apsects. Has something not been taken into account? If there is something missing or is not correct I could extend the implementation.

timfromdigicert:
A future browser alignment ballot might help here, as it is my opinion is that a non-public OCSP or CRL server for public S/MIME would violate Mozilla policy:

"CA operators MUST NOT issue certificates that have:

[...]; or
cRLDistributionPoints or OCSP authorityInfoAccess extensions for which no operational CRL or OCSP service exists."

In the context of Mozilla policy for public S/MIME, I would interpret "operational" to include "publicly accessible".  Because *PUBLIC*  S/MIME.

And this is all just another data point showing that the correct reading of the S/MIME BRs includes the common implicit assumption that CRL and OCSP servers need to be publicly available and operating in a manner generally consistent with how CRL and OCSP servers are generally accepted to operate.

v3/lints/cabf_smime_br/lint_strict_aia_contains_internal_names_test.go
mtgag: Commit https://github.com/zmap/zlint/pull/777/commits/a5711df13cef54288ec3f735d5718643ee008805 covers this.

cardonator:
Great, this LGTM now. @christopher-henderson do you want to take a look?

christopher-henderson:
Thank you @cardonator, @mtgag, and @CBonnell for the thorough policy discussion - it always helps a ton.

Issue #776, opened 2023-12-05
Cutting zlint 3.6.0


robplee:
By my count, there've been over 30 PRs merged including quite a number of new lints since we last cut a release so should we make zlint 3.6.0?

I've tested a zlint integration with importing the latest changes and don't see any issues necessitating a major release. Particular as #775 was scoped to checking new zlint code rather than existing code that may exist around zlint as it is imported.
 
I feel that after #775 and #767, the logical next step is to remove the deprecated interfaces but that'll definitely have to be called 4.0.0.  Therefore, I suggest we cut a minor release now to get the new lints around SMIME and CRLs out into the wild and then 4.0.0 can be about removing LintInterface and we can try to keep release notes for that short.

zakird:
Makes sense to me! In general, I'd love to get in the habit of doing smaller point releases more often.

cardonator:
Agreed, was going to comment the same thing today and forgot. Regarding 4.0.0 release, should we wait a bit until the SMIME lints are more robust before making another major bump? I know we are anxious for more SMIME but will need to do some refactoring of other code when we do remove the old interfaces.

christopher-henderson:
I was literally thinking on this the other day. To several points...

#### Minor Bump, not Major
I agree that now is a good time to do a minor bump. I reckon that we will eserve the major for when we have 

1. Completed deleted the the `Lint` interface 
2. Taken inventory on other items that are on peoples' wishlist  that would necessitate a major release (just so that we get everything in one go).

#### SMIME
Releasing the extant lints in a minor release is a good canary in the coal mine, I think. It's tempting to wait for a release to have more of them, however I admit that I don't see anything inherently wrong with trickling them out.

Given no objections, I can certainly cut a release before the holiday storm is upon us.

christopher-henderson:
Release candidate up at https://github.com/zmap/zlint/releases/tag/v3.6.0-rc1

christopher-henderson:
Cut in https://github.com/zmap/zlint/releases/tag/v3.6.0 :tada:

PR #781, opened 2023-12-14
Limit e_registration_scheme_id_matches_subject_country to no longer apply to LEI or INT organizationIdentifiers

robplee:
Following on from the comments on https://github.com/zmap/zlint/pull/768 and https://github.com/cabforum/smime/issues/216 this PR limits the organizationIdentifier country code lint to no longer compare the country codes in LEI and INT organizationIdentifiers to the subject:countryName.  This should prevent zlint 3.6.0 breaking some SMIME issuance that is considered valid by the SMIME WG.

I think this PR, or something like it, is needed before #776 

Regarding the change, I assume the SMIME WG are only expecting certs to contain a single organizationIdentifier but zcrypto/x509 supports having more than one in the certs it parses so I implemented accordingly.  The discussion on the SMIME BR was saying that for INT/LEI schemes that the "XG" isn't part of identifying the scheme so need not match the countryName so I'm skipping the lint execution if all organizationIdentifiers are INT and/or LEI and if there are some organizationIdentifiers for which the check applies then it's applied but I skip the actual check against the countryName for any INT/LEI organizationIdentifiers.  I don't think this is a realistic case, I'm pretty sure SMIME certs should only contain one organizationIdentifier but I think it makes the lint better this way?


christopher-henderson (approved):

christopher-henderson:
> I'm pretty sure SMIME certs should only contain one organizationIdentifier but I think it makes the lint better this way?

Indeed, there is commonly a mismatch between what the ASN1 can technically encode and what requirements actually refer to. A best-effort attempt to at least consider these unlikely edge cases is always welcomed (although, sometimes we get bogged down with `well, actually, technically...` when considering some of these edge cases precisely because their behavior is not defined by the governing body in question...).

Issue #749, opened 2023-10-09
Subject Key Identifiers in Subscriber TLS Certificates


robstradling:
RFC5280 4.2.1.2 says "this extension SHOULD be included in all end entity certificates", hence https://github.com/zmap/zlint/blob/386a8dc413add9bb92d80badbb4d86f833f6a4e5/v3/lints/rfc/lint_ext_subject_key_identifier_missing_sub_cert.go#L48

However, TLS BRs v2 section 7.1.2.7.6 (Subscriber Certificate Extensions) says the opposite:
"subjectKeyIdentifier NOT RECOMMENDED"

I think ZLint should implement a new `cabf_br` lint that emits a Warning if SKI is present, and that (when enabled) this new lint should override the existing RFC5280 lint in `rfc/lint_ext_subject_key_identifier_missing_sub_cert.go`.  Does that sound right?

robstradling:
The discussion in #762 is relevant here.

PR #788, opened 2024-01-22
SKI is now not recommended.

baloo:
Ballot SC62 appears to have marked Subject Key Identifier as NOT RECOMMENDED.

https://cabforum.org/2023/03/17/ballot-sc62v2-certificate-profiles-update/

See section 7.1.2.7.6 Subscriber Certificate Extensions of the BR: https://cabforum.org/wp-content/uploads/CA-Browser-Forum-TLS-BR-2.0.2.pdf#page=79

I believe zlint should stop issuing warnings for certificates issued after SC62 was effective.

### DIFF
--- a/v3/lints/rfc/lint_ext_subject_key_identifier_missing_sub_cert.go
+++ b/v3/lints/rfc/lint_ext_subject_key_identifier_missing_sub_cert.go
@@ -23,6 +23,7 @@ import (
 type subjectKeyIdMissingSubscriber struct{}
 
 /**********************************************************************
+RFC5280 suggested the addition of SKI extension:
    To facilitate certification path construction, this extension MUST
    appear in all conforming CA certificates, that is, all certificates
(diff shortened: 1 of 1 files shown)


baloo:
Fair warning:
I've never contributed here, nor interacted with the CABF before. Please take this as a suggestion and use of grain of salt are highly suggested.

XolphinMartijn:
Issue #749 ties into this. I'm not sure the proposed solution here is optional. 

The linter source is lint.RFC5280. That source still says the same about this SHOULD requirement.

SC62 went against RFC5280 in this regard. It seems to me like it's a case where the RFC5280 lint should still show a warning if the SKI is not present, where-as a similar lint for Source CABFBR, should print a warning is it is present. 

Anyone linting, could then exclude one of the two, depending on which requirement they'd like to follow more strictly

PR #798, opened 2024-02-16
Add lint for S/MIME subject directory attributes extension (7.1.2.3j)

bitlux:
I tried but failed to create a negative test case for this. The `x509.Certificate` type does not have subject directory attributes, and as far as I can tell, OpenSSL does not support writing them from the command line.

### DIFF
--- a/v3/lints/cabf_smime_br/lint_subject_dir_attr_test.go
+++ b/v3/lints/cabf_smime_br/lint_subject_dir_attr_test.go
@@ -1,0 +1,32 @@
+package cabf_smime_br
+
+import (
+	"testing"
+
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/test"
+)
+
+func TestSMIMESubjectDirAttributes(t *testing.T) {
+	testCases := []struct {
+		Name           string
+		InputFilename  string
+		ExpectedResult lint.LintStatus
+	}{
+		{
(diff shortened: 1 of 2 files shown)


v3/lints/cabf_smime_br/lint_subject_dir_attr_test.go
```diff
@@ -0,0 +1,32 @@
...
+		Name           string
+		InputFilename  string
+		ExpectedResult lint.LintStatus
+	}{
+		{
+			Name:           "pass - no subject dir attributes extension",
+			InputFilename:  "smime/mailboxValidatedStrictWithCommonName.pem",
+			ExpectedResult: lint.Pass,
+		},
+		// A negative test case is hard to construct because neither the x509 package
```
christopher-henderson, line 21: This does indeed happen and there a handful of such cases throughout this code base where producing a negative example is (thankfully?) quite difficult without facing the daunting task of constructing these certificates nearly by hand.

Fortunately, the [test corpus](https://github.com/zmap/zlint/tree/master/v3/integration) helps save us from ourselves by grapping hundreds-of-thousands of certificates from the wild to see if there are any examples that we could not catch ourselves.

None of the test corpus failed, which there are nearly [200k](https://github.com/zmap/zlint/pull/738) SMIME certs. So I guess it suppose the generating libraries just generally do a good job on this one.

christopher-henderson (approved):

Issue #795, opened 2024-02-12
lint_single_email_if_present.go Checks SANs instead of Subject


jdkasten:
I believe the [lint](https://github.com/zmap/zlint/blob/8deb02ba189e5f89d7f1e8a5bf2f75e86f81690e/v3/lints/cabf_smime_br/lint_single_email_if_present.go) 
Name:          "e_single_email_if_present",
Description:   "If present, the subject:emailAddress SHALL contain a single Mailbox Address",
Citation:      "7.1.4.2.h"

refers to this [requirement](https://github.com/cabforum/smime/blob/main/SBR.md#71422-subject-distinguished-name-fields) which is specifically about the subject's relative distingished names rather than subject alternative names.

The lint's code iterates over c.EmailAddresses, which is the the subject alternative name email addresses and not those potentially contained in the subject [[documentation](https://pkg.go.dev/crypto/x509#Certificate)]. There should only be one email address in the subject's distinguished name.

I also believe that the citation should be changed to `7.1.4.2.2.h` rather than its current `7.1.4.2.h`.

cardonator:
You're right about the reference, it should be 7.1.4.2.2.h.

You're also right that the target of the current check is the SANs and not the subject as specified in the BR. I can work on a patch for that. 

Please note that the referenced requirement **does not** state that only one Subject EmailAddress field is allowed in a compliant certificate, but that each Subject EmailAddress field contains only a single email address. Currently this lint is technically wrong per reference but not practically wrong in implementation: each Subject EmailAddress must appear as a SAN in the certificate, and therefore each SAN EmailAddress should also only contain a single email address.

mtgag:
@jdkasten and @cardonator is the discussion in this issue covered by PR #808? If yes, this issue could be closed.

PR #802, opened 2024-02-25
Update single email subject if present

mtgag:
This PR addresses an issue with lint_single_email_if_present.go. The lint takes a look into the SAN entries to check if there is an entry with two email addresses. The description and citation shows to the entries in the subjectDN. On the one hand taking a look into the SAN entries for two email addresses is meaningful, on the other hand the citation and implementation do not align. Therefore this PR adds a new lint to check the subjectDN entries, re-using the implementation. I believe following actions can be further taken. One is to keep the original lint and change the citation/description. A second is to incorporate the two implementations into one lint that checks both  the subjectDN and the SAN entries.

Best,
Vangelis


christopher-henderson:
Ah, I see the issue, thank you for catching that @mtgag.

I think that my take is that the later is preferred...

> A second is to incorporate the two implementations into one lint that checks both the subjectDN and the SAN entries.

We've done this before in the repo wherein a lint checks additional, related, structures as a part of one unified lint. I believe that that this helps keep citation implementations near each other as otherwise it would be easy to forget to update one if the implementation were split.

cardonator:
I agree that an additional lint is warranted here. I believe this will close #795 as well. Thanks for tacking that, @mtgag

v3/lints/cabf_smime_br/lint_single_email_subject_if_present.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func (l *singleEmailSubjectIfPresent) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, email := range c.Subject.EmailAddress {
+		_, err := mail.ParseAddress(email)
+		if err != nil {
+			return &lint.LintResult{
+				Status:       lint.Error,
+				Details:      fmt.Sprintf("subject:emailAddress was present and contained an invalid email address (%s)", email),
+				LintMetadata: lint.LintMetadata{},
```
bitlux, line 56: This field isn't needed, `lint.Metadata{}` is the default.

v3/lints/cabf_smime_br/lint_single_email_subject_if_present_test.go
```diff
@@ -0,0 +1,40 @@
...
+		InputFilename  string
+		ExpectedResult lint.LintStatus
+	}{
+		{
+			Name:           "Error - email address present in subjectDN with multiple values",
+			InputFilename:  "smime/twoEmailAddressesInSubjectDN.pem",
+			ExpectedResult: lint.Error,
+		},
+		{
+			Name:           "Error - email address present in subjectDN with one value",
```
bitlux, line 22: Should be `Pass`, not `Error`.

v3/lints/cabf_smime_br/lint_single_email_subject_if_present_test.go
```diff
@@ -0,0 +1,40 @@
...
+			InputFilename:  "smime/oneEmailAddressInSubjectDN.pem",
+			ExpectedResult: lint.Pass,
+		},
+		{
+			Name:           "Error - no email address present in subjectDN",
```
bitlux, line 27: Should be `NA`, not `Error`.

v3/lints/cabf_smime_br/lint_single_email_subject_if_present.go
```diff
@@ -0,0 +1,62 @@
...
+func init() {
+	lint.RegisterCertificateLint(&lint.CertificateLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "e_single_email_subject_if_present",
+			Description:   "If present, the subject:emailAddress SHALL contain a single Mailbox Address",
+			Citation:      "7.1.4.2.h",
+			Source:        lint.CABFSMIMEBaselineRequirements,
+			EffectiveDate: util.CABF_SMIME_BRs_1_0_0_Date,
+		},
+		Lint: func() lint.LintInterface { return &singleEmailSubjectIfPresent{} },
```
bitlux, line 35: Should probably be `Lint: NewSingleEmailSubjectIfPresent`?

v3/lints/cabf_smime_br/lint_single_email_subject_if_present.go
```diff
@@ -0,0 +1,62 @@
...
+	return &singleEmailSubjectIfPresent{}
+}
+
+func (l *singleEmailSubjectIfPresent) CheckApplies(c *x509.Certificate) bool {
+	return util.IsSubscriberCert(c) && c.Subject.EmailAddress != nil && util.IsSMIMEBRCertificate(c)
+}
+
+func (l *singleEmailSubjectIfPresent) Execute(c *x509.Certificate) *lint.LintResult {
+	for _, email := range c.Subject.EmailAddress {
+		_, err := mail.ParseAddress(email)
```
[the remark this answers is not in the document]
mtgag: 👍

v3/lints/cabf_smime_br/lint_single_email_subject_if_present.go
mtgag: 👍

v3/lints/cabf_smime_br/lint_single_email_subject_if_present.go
mtgag: 👍

v3/lints/cabf_smime_br/lint_single_email_subject_if_present_test.go
mtgag: 👍

mtgag:
> I agree that an additional lint is warranted here. I believe this will close #795 as well. Thanks for tackling that, @mtgag

Should I incorporate lint_single_email_if_present (and then delete it) into lint_single_subject_email_if_present, or keep both and slightly refactor it to reflect the discussion here and comments from #795?

cardonator:
My suggestion would be the latter.

mtgag:
> My suggestion would be the latter.

I will open a new PR for this. Then we have a stable pre-review and pre-merge state in lint_single_email_if_present addressing https://github.com/zmap/zlint/issues/795. We can still decide whether to go for one or two lints. I see benefits in both approaches.

christopher-henderson (approved):

PR #800, opened 2024-02-22
Add lint to enforce SMIME BRs: 7.1.4.2.1 requirement for mailbox addr…

toddgaunt-gs:
Hello, I'm going to be attempting to continue the work Rob Lee had been doing and would like to contribute lints from GlobalSign to the zlint repository. We mostly have smime lints to contribute developed in the past year and my goal is to continue contributing them one at a time. This is my first contribution beyond a documentation update so let me know if there is anything missing.

Description:
All mailbox addresses appearing in subjectDN or dirName must be repeated in san:rfc822Name or san:otherName. This lint does its best to detect mailbox address values in the subjectDN or dirName and if any are detected ensures they are repeated.

PR #806, opened 2024-02-28
Add lint for BR 7.1.4.2.2a mailbox-validated

bitlux:
The changes to san.go are copied from #800.

### DIFF
--- a/v3/lints/cabf_smime_br/lint_commonname_mailbox_validated.go
+++ b/v3/lints/cabf_smime_br/lint_commonname_mailbox_validated.go
@@ -15,9 +15,6 @@
 package cabf_smime_br
 
 import (
-	"fmt"
-
-	"github.com/zmap/zcrypto/encoding/asn1"
 	"github.com/zmap/zcrypto/x509"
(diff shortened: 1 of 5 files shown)


christopher-henderson (changes requested):
So the table of acceptable values is as follows...

|Certificate Type|Contents|
|---|---|
|Mailbox-validated | Mailbox Address
|Organization-validated | subject:organizationName or Mailbox Address
|Sponsor-validated | Personal Name, subject:pseudonym, or Mailbox Address
|Individual-validated | Personal Name, subject:pseudonym, or Mailbox Address

This lint, as written, appears to only be checking the first row (mailbox validated).

This is perfectly fine, although I suppose that we should state this explicitly in the citation.

Do you think that the following reasonable?

> If present, the commonName attribute of a mailbox validated certificate SHALL contain a mailbox address

bitlux:
Thanks for the review. 

Yes, I intended to check only the mailbox-validated certificates with this PR (as indicated by the title and lint filename). I updated the description.

christopher-henderson (approved):

PR #814, opened 2024-03-11
cRLIssuer MUST NOT be present

XolphinMartijn:
On the back of a recent incident report on bugzilla, this lint should help CAs identify this issue. 

I'm a bit unsure of the use of lint.Fatal and when it's appropriate to use. Open to feedback on this item


christopher-henderson (approved):
Thank you!

> I'm a bit unsure of the use of lint.Fatal and when it's appropriate to use. Open to feedback on this item

I believe that this is a reasonable use of the return code. It is typically used to indicate that the certificate in question is fundamentally not sane in some (which is usually an encoding issue, precisely as you used it).

Issue #818, opened 2024-03-28
Printable `BMPString` in subject DN fields marked as failing lint `e_subject_dn_not_printable_characters`


cwjnkins:
This concerns changes introduced by PR #249, specifically [lint](https://github.com/zmap/zlint/blob/master/v3/lints/rfc/lint_subject_dn_not_printable_characters.go) `e_subject_dn_not_printable_characters`. Some attribute values for subject DNs are declared as `DirectoryString` types, such as `X520CommonName`

    -- Naming attributes of type X520CommonName:
    --   X520CommonName ::= DirectoryName (SIZE (1..ub-common-name))
    --
    -- Expanded to avoid parameterized type:
    X520CommonName ::= CHOICE {
      teletexString     TeletexString   (SIZE (1..ub-common-name)),
      printableString   PrintableString (SIZE (1..ub-common-name)),
      universalString   UniversalString (SIZE (1..ub-common-name)),
      utf8String        UTF8String      (SIZE (1..ub-common-name)),
      bmpString         BMPString       (SIZE (1..ub-common-name)) }

which in particular includes the non-UTF8 encoded strings `BMPString` (a subset of UTF16) and `UniversalString`.

[...]

PR #819, opened 2024-03-28
Handle BMPString properly in e_subject_dn_not_printable_characters 

cardonator:
Looking at #818 , other string types should be allowable as printable in subject fields to conform to RFC 5280.

### DIFF
--- a/v3/lints/rfc/lint_subject_dn_not_printable_characters.go
+++ b/v3/lints/rfc/lint_subject_dn_not_printable_characters.go
@@ -57,7 +57,7 @@ func (l *subjectDNNotPrintableCharacters) Execute(c *x509.Certificate) *lint.Lin
 	for _, attrTypeAndValueSet := range rdnSequence {
 		for _, attrTypeAndValue := range attrTypeAndValueSet {
 			bytes := attrTypeAndValue.Value.Bytes
-			runes := []rune{}
+			var runes []rune
 			if attrTypeAndValue.Value.Tag == tagBMPString {
 				runestr, _ := util.ParseBMPString(bytes)
 				runes = []rune(runestr)
--- a/v3/lints/rfc/lint_subject_dn_not_printable_characters.go
+++ b/v3/lints/rfc/lint_subject_dn_not_printable_characters.go
@@ -15,8 +15,6 @@
 package rfc
 
 import (
-	"unicode/utf8"
-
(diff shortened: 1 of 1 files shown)


cardonator:
I need to get/create certs with matching field data types to add additional tests for this update. However, this does pass the cert referenced in #818 as well as the test certs we created for the original lint a few years ago.

cardonator:
I noticed in the Integration Test run that the number of certs in the test corpus that fail this lint has dropped from 541 to 34. That seems concerning to me but I'm not sure if it's something to be concerned about. One noteworthy comment here is that it increases the efficacy of this lint for RFC-compliant use cases, but decreases the efficacy for public PKI.

I'll look for feedback from others before deciding how to proceed.

zakird:
Part of me wonders if we should make it a "community best practice" lint? It doesn't seem to have tripped up any CAs in the last five years?

CBonnell:
SC-62 (the "profiles" ballot) added restrictions on the allowed encodings for `DirectoryString` attribute value types: https://cabforum.org/working-groups/server/baseline-requirements/requirements/#7142-subject-attribute-encoding.

Specifically, the deprecated `BMPString`, `UniversalString`, and `TeletexString` types are no longer allowed. It's likely more expedient to add this prohibition in a separate lint and forego trying to get the lints for these legacy types correct. Lints for BMPString and UniversalString might be not be too bad to implement, but there will be pain with TeletexString.

zakird:
@CBonnell's suggestion makes sense to me.

cardonator:
Does that mean we should add the restrictions and linting from the lint in question to the BR lints and remove it from the RFC lints?

zakird:
Yes, I think that moving them to a lint that points to the BRs is probably the right move (if one doesn't already exist.)

mtgag:
I could contribute on the BR lints about allowed encodings and also provide certificates with the disallowed encodings. If that is fine and you are not already on it, I could start working on it.

mtgag:
> I could contribute on the BR lints about allowed encodings and also provide certificates with the disallowed encodings. If that is fine and you are not already on it, I could start working on it.

This is the corresponding PR: https://github.com/zmap/zlint/pull/824

cardonator:
With #824, I'm thinking we should close this PR, right @christopher-henderson ?

PR #820, opened 2024-03-30
Lint for checking that organizationIdentifier Subject attribute and CABFOrganizationIdentifier extension are consistent as per EVG 9.2.8

defacto64:
Please add this lint for checking that the organizationIdentifier Subject attribute and the CABFOrganizationIdentifier extension are consistent as per EVG 9.2.8.


v3/lints/cabf_ev/lint_ev_orgid_inconsistent_subj_and_ext.go
```diff
@@ -0,0 +1,114 @@
...
+
+func ParseOrgId(orgIdString string, orgId *OrganizationIdentifier) error {
+
+	// This is according to the EVG (stricter than ETSI EN 319 412-1)
+	OrgIdPattern := `^[A-Z]{3}[A-Z]{2}(?:\+[A-Z]{2})?\-.+$`
+
+	compiledRegexp, err := regexp.Compile(OrgIdPattern)
+	if err != nil {
+		// This should neve occur, but one never knows....
+		panic(err)
```
christopher-henderson, line 60: [regexp.MustCompile](https://pkg.go.dev/regexp#MustCompile) will do what you're looking for here.

It is also common to place this regex compilation in the outer scope...

```go
var compiledRegexp = regexp.MustCompile(`^[A-Z]{3}[A-Z]{2}(?:\+[A-Z]{2})?\-.+$`)

func ParseOrgId(orgIdString string, orgId *OrganizationIdentifier) error {
    ....
}
```

Doing so makes it such that any panic that gets introduced would get triggered merely by starting up the program, thus reducing the likelihood of it making past CICD as code coverage on the line is guaranteed,

v3/lints/cabf_ev/lint_ev_orgid_inconsistent_subj_and_ext.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	if !compiledRegexp.MatchString(orgIdString) {
+		return errors.New("Cannot parse organizationIdentifier: it is probably invalid")
+	}
+
+	orgId.Scheme = orgIdString[0:3]
```
christopher-henderson, line 67: I'm always torn on this because

1. UTF8 is trickier than it seems
2. It at least _seems_ like all of our data is ASCII, despite not being declared as such,

For the first point, consider the [following example](https://go.dev/play/p/Nvci4QaO2kQ).

```go
package main

import (
	"fmt"
)

func main() {
	var darnit string = "Hello, 世界"
	// You would think it would be 世 but we get ä
	fmt.Println(string(darnit[7]))
}
```

This is because indexing into a string does a literal lookup to the byte at that location, however UTF8 is a variable length encoding.

For the second point, the EV BRs allows for this to be a `UTF8String`. However the contents seem to be entirely sources from ISO 3166, and those [seem to all be ASCII](https://en.wikipedia.org/wiki/List_of_ISO_3166_country_codes) without explicitly declaring as such. Additionally, the test corpus passes and it is fairly representative of what is out in the wild.

---

If you care to consider this then casting to a `[]rune` before indexing does the right thing.

```go
string([]rune(orgIdString)[6:8])
```

Note that, although it is not an issue in _this_ scenario, but do make a mental note that the above is linear time and may result in an allocation. That would be a nasty surprise sometime in the future.

v3/lints/cabf_ev/lint_ev_orgid_inconsistent_subj_and_ext.go
```diff
@@ -0,0 +1,114 @@
...
+	// It is actually mandatory that, if orgId is present, cabfOrgId be present as well,
+	// however this is already checked by another lint
+	return util.IsEV(c.PolicyIdentifiers) && (len(c.Subject.OrganizationIDs) > 0) &&
+		util.IsExtInCert(c, util.CabfExtensionOrganizationIdentifier)
+}
+
+func (l *orgIdInconsistentSubjAndExt) Execute(c *x509.Certificate) *lint.LintResult {
+	// It should be safe to assume there is only one element in OrganizationIDs
+	var orgId OrganizationIdentifier
+	err := ParseOrgId(c.Subject.OrganizationIDs[0], &orgId)
```
christopher-henderson, line 96: Please bounds check this as I'm not sure if the x509 package guarantees the presence of at least one org ID. It's not likely, but panics are a nasty surprise and incoming certs are assumed to be relatively wild.

christopher-henderson (changes requested):
Semantically it seems good to go.

However, we are in a release candidate cycle so only bug fixes are going in until the 3.6.2 release.

v3/lints/cabf_ev/lint_ev_orgid_inconsistent_subj_and_ext.go
CBonnell: I think using named capture groups would resolve the UTF-8 concern. It would also have the benefit of  removing the "magic string indexes", improving clarity of the code.

v3/lints/cabf_ev/lint_ev_orgid_inconsistent_subj_and_ext.go
defacto64: Chris, there is already a bounds check in CheckApplies() so I suppose it's no use to redoit here....am I wrong?

defacto64:
Thank you Chris and Corey for your suggestions. I have revised my code accordingly.

christopher-henderson (approved):

PR #824, opened 2024-04-05
Subject rdns correct encoding

mtgag:
This PR addresses the proposal in PR zmap#819, to have separate lints about the encoding of subject attributes as specified in BR 2.0.0.


christopher-henderson (commented):
This looks good, but I think we should wait until the release of 3.6.2 since CAs seems to have been operating for years just fine without this.

dadrian (approved):

mtgag:
> This looks good, but I think we should wait until the release of 3.6.2 since CAs seems to have been operating for years just fine without this.

Yes, we can put it on hold until and wait for the proper release.

PR #825, opened 2024-04-05
Limit scope for cn checking in SAN

mtgag:
This PR addresses issues of discussion in PR #809


v3/lints/cabf_smime_br/mailbox_address_from_san.go
```diff
@@ -90,12 +103,14 @@ func (l *MailboxAddressFromSAN) Execute(c *x509.Certificate) *lint.LintResult {
...
 }
 
-func getMailboxAddressesFromDistinguishedName(name pkix.Name) []string {
+func getMailboxAddressesFromDistinguishedName(name pkix.Name, includeCN bool) []string {
 	mailboxAddresses := []string{}
 
-	for _, commonName := range name.CommonNames {
-		if util.IsMailboxAddress(commonName) {
-			mailboxAddresses = append(mailboxAddresses, commonName)
+	if includeCN {
```
vanbroup, line 109:
```suggestion
	if len(name.EmailAddress) == 0 {
```
This would cover more scenario's and doesn't require additional function parameters.

mtgag: would not that, probably very unlikely, produce some false positives? E.g. for a sponsor-validated certificate with a subjectDN featuring a CN which has a pseudonym with a @ character and no emailAddress RDN in subject.

vanbroup: That might be possible indeed.

christopher-henderson:
Howdy @mtgag and @vanbroup, I'm just catching up on this. We cut a release candidate last week, but this seems like it would be a good call for an RC2, no?

vanbroup:
Agreed

mtgag:
> Howdy @mtgag and @vanbroup, I'm just catching up on this. We cut a release candidate last week, but this seems like it would be a good call for an RC2, no?

Yes, it looks much better now after the comments. Thanks for this.

christopher-henderson (approved):

PR #830, opened 2024-04-09
Add lint to check for a valid business category in EV certificates

defacto64:
Please add this lint to check for a valid business category in EV certificates, as per EVG 9.2.3


defacto64:
Cannot understand why "integration-test" fails....

XolphinMartijn:
> Cannot understand why "integration-test" fails....

It looks like 11306 certs in the test data-set are returning on error on this linter. If that number of certs hitting this error is indeed correct, config.json needs to be updated to reflect this expected error count

defacto64:
That's quite strange and unexpected. 
I have checked one of the offending certficates (whose fingerprint I found in the integration-test log) and it makes zlint **crash**:
https://crt.sh/?q=4712f1b2a94994b55626ecba2104bbf23d39c05e7a2751e5af8a923bac23fd8f
I mean, even the current official version of zlint....
I am quite perplexed....

defacto64:
Look at this: how come?
image

defacto64:
Thanks Rob, that's interesting, but I don't understand what this has to do with my lint...and the fact that "integration-test" fails....

XolphinMartijn:
@defacto64  I don't think it is. https://github.com/zmap/zlint/actions/runs/8613740876/job/23605682791 flags the same certificate as FATAL, but that is just a warning in the log, just as during the run on your PR.

I just ran the lint locally against a modified zlint version, which spits out the offending thumbprints. It seems to believe https://crt.sh/?q=1de8b31449944cdd6f5657af900a488c517246a474e6500832621dc058e9aea0 has a bad businessCategory

v3/lints/cabf_ev/lint_ev_invalid_business_category.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (l *invalidBusinessCategory) Execute(c *x509.Certificate) *lint.LintResult {
+
+	for _, v := range c.Subject.Names {
+		if util.BusinessOID.Equal(v.Type) {
+			businessCategory := v.Value
+			if (businessCategory == "Private Organization") ||
+				(businessCategory == "Government Entity") ||
+				(businessCategory == "Business Entity") ||
+				(businessCategory == "“Non‐Commercial Entity") {
```
XolphinMartijn, line 59:
```suggestion
				(businessCategory == "Non‐Commercial Entity") {
```

@defacto64  This would be why it's flagging some certs as bad, when they're not :)

XolphinMartijn: Note also, it seems to be doing a check against the same capitalization. Should this be regardless of which capitalization is used or not?

defacto64: Thank you Martijn, I really missed that one!

defacto64:
Hmmm... I fixed that typo, but there must still be something wrong somewhere.

XolphinMartijn:
> Hmmm... I fixed that typo, but there must still be something wrong somewhere.

Correct, I think my second comment went away. Some of the test certs have different capitalization on the strings. I'm not sure if that actually is a violation or not though

defacto64:
Oh, I assumed that capitalization is meaningful. The EVGs at 9.2.3 are quite clear on this, IMO:
`This field MUST contain one of the following strings: “Private Organization”, “Government Entity”, “Business Entity”, or “Non‐Commercial Entity`

defacto64:
And, if I am not mistaken, also pkilint only allows the case-sensitive versions of those four strings....

XolphinMartijn:
@defacto64  I suspect https://github.com/defacto64/zlint/pull/1 will solve this

cardonator:
I think the capitalization matters, considering that it matters in multiple other parts of the BRs and EVGs where sensitivity is not explicitly mentioned.

defacto64:
This morning I positively verified that pkilint reports ERROR on a businessCategory that has one of the values expected by EVG 9.2.3 but in all capital letters (e.g. "PRIVATE ORGANIZATION"), so I think we are quite confident in the interpretation of 9.2.3. 
At this point I suppose that in the corpus of test certificates used in the 'integration-test' there are EV certificates with businessCategory in capital letters and for which, however, an error is not expected. 
Right now I am not sure how to fix this problem, maybe I need a hand from Chris?
Any suggestions welcome.

XolphinMartijn:
> This morning I positively verified that pkilint reports ERROR on a businessCategory that has one of the values expected by EVG 9.2.3 but in all capital letters (e.g. "PRIVATE ORGANIZATION"), so I think we are quite confident in the interpretation of 9.2.3. At this point I suppose that in the corpus of test certificates used in the 'integration-test' there are EV certificates with businessCategory in capital letters and for which, however, an error is not expected. Right now I am not sure how to fix this problem, maybe I need a hand from Chris? Any suggestions welcome.

@defacto64  Please see https://github.com/defacto64/zlint/pull/1 which I filed against your repo. It should solve this I would imagine

defacto64:
Unfortunately, it seems not, Martijn.

v3/integration/config.json
```diff
@@ -426,6 +426,8 @@
       "ErrCount": 2
     },
     "e_ev_country_name_missing": {},
+    "e_ev_invalid_business_category": {
+      "ErrCount": 11306
```
XolphinMartijn, line 430:
```suggestion
      "ErrCount": 11306
      }
```

This time I forgot a character ;)

defacto64:
Εύρηκα!

v3/lints/cabf_ev/lint_ev_invalid_business_category.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterCertificateLint(&lint.CertificateLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "e_ev_invalid_business_category",
+			Description:   "Checks that businessCategory contains a valid value as per EV Guidelines 9.2.3",
+			Citation:      "EVGs 9.2.3",
```
aaomidi, line 33: I think this is supposed to be 7.1.4.2.3? https://github.com/cabforum/servercert/blob/main/docs/EVG.md#71423-subject-business-category-field

aaomidi: I see that this JUST changed though https://github.com/cabforum/servercert/commit/c4a34fe2292022e0a04ba66b5a85df75907ac2a2

defacto64: @aaomidi , the requirement is still there, identical, although its location within the EVG has just changed as a consequence of reformatting the EVG according to RFC3647 (the goal of ballot SC65). So the relevant section changed from 9.2.3 to 7.1.4.2.3. if you prefer, I could refer to this latter in the `Citation` field, but the requirements has always existed.... so I am unsure what's best.

aaomidi: Oh yeah I agree, I do think though that since the citation has changed, and this PR hasn't merged yet, it would make sense to cite the current location in the document?

v3/lints/cabf_ev/lint_ev_invalid_business_category.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (l *invalidBusinessCategory) Execute(c *x509.Certificate) *lint.LintResult {
+
+	for _, v := range c.Subject.Names {
+		if util.BusinessOID.Equal(v.Type) {
+			businessCategory := v.Value
+			if (businessCategory == "Private Organization") ||
+				(businessCategory == "Government Entity") ||
+				(businessCategory == "Business Entity") ||
+				(businessCategory == "Non‐Commercial Entity") {
```
XolphinMartijn, line 59: I've been running this lint against a large number of issued EV certs, and Non-Commercial Entities are being shown as invalid.  

I've done a little bit of digging, and it looks like the character used as Minus sign here, is the UTF-8 "HYPHEN" character (E2 80 90)

As such, please accept the following change to correct this to a "HYPHEN-MINUS" character:
```suggestion
				(businessCategory == "Non-Commercial Entity") {
```

defacto64: You are right @XolphinMartijn , but I am wondering how that happened as my source code in my own local repository (on my PC) has the plain MINUS character....

v3/lints/cabf_ev/lint_ev_invalid_business_category.go
defacto64: I changed the reference to 7.1.4.2.3

defacto64:
A recent incident on https://bugzilla.mozilla.org corroborates that businessCategory in EV certificates is a case-sensitive attribute at least from one major certificate consumer's point ot view, and based on what is by far the commonest practice in the industry, therefore I think it would be good to have this lint - to mitigate the risk that similar incidents re-occur.

v3/integration/config.json
```diff
@@ -426,6 +426,9 @@
       "ErrCount": 2
     },
     "e_ev_country_name_missing": {},
+    "e_ev_invalid_business_category": {
+      "ErrCount": 10957
```
christopher-henderson, line 430: This lint seems fine, but 10k+ failed certificates is a bit smelly (unless this is a requirement that you happen to know is commonly failed).

Off the top of my head, either these 10k truly are failures or they are doing something like using mixed casing for these strings (which may-or-may-not be valid, but it would be good to know).

Would you mind running the integration tests locally? My typical workflow is to simply take a random sampling of certs that failed and inspect them closely.

```bash
cd v3/
make integration INT_FLAGS="-lintSummary -fingerprintSummary -lintFilter='e_ev_invalid_business_category'"
```

This will give you a printout of the fingerprints of the certificates that failed. You can then hop on over to a tool such as [crt.sh](https://crt.sh/) and plug that fingerprint in to see a copy of the offending cert as well as run linting from other similar tools to see what their opinion is.

christopher-henderson (changes requested):

v3/integration/config.json
defacto64: I understand your perplexity. I was a little too laconic in proposing this lint. As I said, I took inspiration from a specific incident reported on Bugzilla (bug #1894560). As you can see there, at least two root programs (Mozilla and Chrome) have made it clear that, like it or not, businessCategory must be case-sensitive. Yes, there are numerous mis-issued EV certificates in the test corpus. Probably, until two months ago several CAs had a different idea on the matter and issued EV certificates with businessCategory in uppercase, honestly thinking that it was okay; but unfortunately, in light of bug #1894560, one must conclude that this is not the case. Hence this lint. 

We must also consider that this lint detects not only the businessCategory values ​​that do not have the right casing, but also those that are completely wrong such as the following very weird case: 
https://crt.sh/?q=fffc65f342a9cc71bed437fea03845ddfa94f403edf1bc3182c0ae4baa0ff6d8 
(there are many instances like this one). 

Furthermore, this lint detects errors that are almost invisibile at first sight, such as when the dash in "Non-Commercial Entity" is not the normal minus sign character (0x2D) but the Unicode HYPHEN character (0xE28090 in UTF-8). It seems we have a number of this cases in the testCorpus, unfortunately. Other linters, such as Pkilint, also flag it as an error.

christopher-henderson: Thank you for the rich context and history!

I figured that with so many errors that it had to be something as inane as casing or fiddly-bits such as extra whitespace or misunderstood character encodings. Plain text encoding for machine consumption truly is the devil.

christopher-henderson (approved):

PR #833, opened 2024-04-16
Add lint to cover TLS BR v2 EKU checks

vanbroup:
This introduces a new lint called `e_sub_cert_eku_check` and sets the ineffective date for `e_sub_cert_eku_server_auth_client_auth_missing` and `w_sub_cert_eku_extra_values`.

This update only covers subscriber certificates, the lints for CA certificates will also need to be reviewed.

The `util.IsServerAuthCert` did not consider certificates that attest the [CA/Browser Forum Reserved Certificate Policy Identifiers](https://cabforum.org/working-groups/server/baseline-requirements/requirements/#7161-reserved-certificate-policy-identifiers) as specified in section 7.1.6.1 of the BRs, but who did not include the `serverAuth` EKU. This has been addressed to cover the expectations of attesting a policy and to cover all test scenario's of this lint.


v3/lints/cabf_br/lint_sub_cert_eku_check.go
```diff
@@ -0,0 +1,83 @@
...
+		Lint: NewSubExtKeyUsageCheck,
+	})
+}
+
+func NewSubExtKeyUsageCheck() lint.LintInterface {
+	return &subExtKeyUsageCheck{}
+}
+
+func (l *subExtKeyUsageCheck) CheckApplies(c *x509.Certificate) bool {
+	return c.ExtKeyUsage != nil
```
aaomidi, line 45: I think this should just be `util.IsSubscriberCert()` or `util.IsSubscriberCert() && util.IsExtInCert(c, util.EkuSynOid)`

v3/lints/cabf_br/lint_sub_cert_eku_check.go
```diff
@@ -0,0 +1,83 @@
...
+
+	if len(c.UnknownExtKeyUsage) > 0 {
+		for _, eku := range c.UnknownExtKeyUsage {
+			if eku.Equal(util.PreCertificateSigningCertificateEKU) {
+				return &lint.LintResult{Status: lint.Error, Details: "Precertificate Signing Certificate extKeyUsage MUST NOT be present"}
+			}
+		}
+	}
+
+	if (len(c.ExtKeyUsage) > 2 && !hasClientAuthEKU) || len(c.UnknownExtKeyUsage) > 0 {
```
mathewhodson, line 78: Maybe you could set a variable in the `default` case in the `range c.ExtKeyUsage` loop and check that.

christopher-henderson: I may advocate for keeping a list of of OIDs found during the execution of the lint and cross referencing them at them end against the list of known OIDs.

```go
var known = map[string]struct{}{
	"1.3.6.1.5.5.7.3.1": {},
	"1.3.6.1.5.5.7.3.2": {},
	"1.3.6.1.5.5.7.3.3": {},
	"1.3.6.1.5.5.7.3.4": {},
	"1.3.6.1.5.5.7.3.8": {},
	"1.3.6.1.5.5.7.3.9": {},
	"2.5.29.37.0": {},
	"1.3.6.1.4.1.11129.2.4.4": {},
}

func lint() {
    found := []string
    for _, eku := range c.ExtKeyUsage {
        found = append(found, eku.String()
        ....
        ....
        ....
    }
    for _, eku := range.UnknownExtKeyUsage {
        found = append(found, eku.String()
        ....
        ....
        ....
    }
    for _, oid := range found {
        _, ok := known[oid]
        if !ok {
            return &lint.LintResult{Status: lint.Warn, Details: "any other value than id-kp-serverAuth and id-kp-clientAuth is NOT RECOMMENDED"}
        }
    }
}

```

v3/lints/cabf_br/lint_sub_cert_eku_check.go
```diff
@@ -0,0 +1,83 @@
...
+
+			return &lint.LintResult{Status: lint.Error, Details: fmt.Sprintf("%s MUST NOT be present", util.GetEKUString(eku))}
+		}
+	}
+
+	if !hasServerAuthEKU {
+		return &lint.LintResult{Status: lint.Error, Details: "id-kp-serverAuth MUST be present"}
+	}
+
+	if len(c.UnknownExtKeyUsage) > 0 {
```
vanbroup: This check is not needed

v3/lints/cabf_br/lint_sub_cert_eku_check.go
vanbroup: The ExtKeyUsage and UnknownExtKeyUsage are not of the same type:

```go
ExtKeyUsage        []ExtKeyUsage           // Sequence of extended key usages.
UnknownExtKeyUsage []asn1.ObjectIdentifier // Encountered extended key usages unknown to this package.
```

The lint currently checks for MUST, MUST NOT, MAY, and NOT RECOMMENDED, and returns the corresponding lint status and error.

We can create a map with each EKU and the requirement (i.e., MUST) and lookup the value while ranging over the two EKU lists, but I guess this would be less efficient

christopher-henderson: I reckon that this would be an efficiency modifier in the range of nanoseconds (possibly into ms) for each run of the lint. In general, legibility trumps such optimizations since correctness of the lint is paramount.

christopher-henderson (approved):

Issue #639, opened 2021-10-14
Add lint which checks for explicitly encoded critical default values


alex:
From ITU X.690:

```
11.5 Set and sequence components with default value

The encoding of a set value or sequence value shall not include an encoding for any component value which is equal to its default value.

```

From RFC 5280:

```
   Extension  ::=  SEQUENCE  {
        extnID      OBJECT IDENTIFIER,
        critical    BOOLEAN DEFAULT FALSE,
        extnValue   OCTET STRING
                    -- contains the DER encoding of an ASN.1 value
                    -- corresponding to the extension type identified
                    -- by extnID
        }
```

therefore `critical` should never be encoded with a false value.

Some issuing software does do this though, so a lint to capture it would be valuable.

(My proximate motivation: https://github.com/pyca/cryptography/issues/6368)

christopher-henderson:
I'm sorry, @alex, but I do not believe that this is quite feasible in our codebase without some interesting gymnastics. The reason is as follows...

Consider the definition of a `pkix.Extension` withing the Go standard library (and `zcyrpto` as well)

```go
// Extension represents the ASN.1 structure of the same name. See RFC
// 5280, section 4.2.
type Extension struct {
	Id       asn1.ObjectIdentifier
	Critical bool `asn1:"optional"`
	Value    []byte
}
```

The `Critical` data member here is a `bool` value. In Go, non-pointer values _may not_ be uninitialized. That is, if this field was not found within the source data structure then Go will initialize this field to `false` on our behalf. That is to say, we may encounter a `false` value, however we do not know whether the field was (correctly) missing and was thus initialized to `false` via Go's semantics **_or_** if the field actually was explicitly encoded with `false` (the issue that you encountered in `pyca`).

Take the following [Go playground sample](https://play.golang.org/p/U32Befr0Odk)...

```go
package main

import (
	"fmt"
	"crypto/x509/pkix"
)

func main() {
	fmt.Println(pkix.Extension{})
}
```

We could, conceivable, modify `zcrypto` with the following definition...

```go
type Extension struct {
	Id       asn1.ObjectIdentifier
	Critical *bool `asn1:"optional"`
	Value    []byte
}
```

That is, a pointer to a `bool` which we may then use to infer whether or not the field was explicitly encoded. However, this would be an extremely invasive change as all dependent code would be forced to account for `*bool` rather than a plain `bool`.

Alternatively, we could extend the struct to include this information at the time of deserialization.

```go
type Extension struct {
[...]

PR #839, opened 2024-05-01
Add lint for checking that the 'critical' field is properly DER-encoded in extensions

defacto64:
It is generally considered a requirement that certificates be DER encoded (meaning not BER), although RFC 5280 is not crystal clear on this. In any case, the non-properly-DER encoding of some certificate fields has historically been considered a compliance error, as it can be seen on https://bugzilla.mozilla.org, therefore it's useful to be able to detect it. 

Among other things, DER stipulates that "the encoding of a set value or sequence value shall not include an encoding for any component value which is equal to its default value" (see X.690). Therefore, when a certificate extension is not critical, its 'critical' field should be omitted. This lint checks just that.


christopher-henderson (approved):
Thank you! This one was indeed tricky as I was attempted to full deserialize the type in to a Go type, which had all of the complications and gotchas written up in #639

PR #840, opened 2024-05-04
Add lint to check that SubCA certificates do not have illegal values in their EKU extension

defacto64:
This lint verifies that Subordinate CA certificates do not have illegal values in their EKU extension according to the CABF BRs, and generates an error in case they do. This kind of problem has happened a few times in the past, as can be seen on https://bugzilla.mozilla.org, but apparently not in recent times; at any rate, it's better to prevent it.

This lint does not discriminate between normal SubCA certificates and cross-certificates, this being impossible in the absence of an extra input (which could be passed through a specific lint-configuration, but I find it preferable to do without it). However, this is not necessary in order to do a few essential checks that were missing until now. 

The lint's logic is as follows: in the CABF BR context, if the anyExtendedKeyUsage is present in a SubCA cert, then it can be inferred that it's actually an unrestricted cross certificate; if instead anyExtendedKeyUsage is not present, then it must be either a restricted cross certificate or a normal SubCA (non cross), and in both these cases the same rules apply to the EKU extension.


v3/lints/cabf_br/lint_ca_invalid_eku.go
```diff
@@ -0,0 +1,81 @@
...
+			Details: "anyExtendedKeyUsage MUST NOT be accompanied by any other value in the EKU extension",
+		}
+	}
+
+	// If we get here, it is necessarily a SubCA with serverAuth in the EKU
+	for _, eku := range c.ExtKeyUsage {
+		if eku == x509.ExtKeyUsageEmailProtection ||
+			eku == x509.ExtKeyUsageCodeSigning ||
+			eku == x509.ExtKeyUsageTimeStamping ||
+			eku == x509.ExtKeyUsageOcspSigning {
```
christopher-henderson, line 71: I believe that we are referring to sub-section g?

> For Subordinate CA Certificates that will be used to issue TLS certificates, the value
id-kp-serverAuth [RFC5280] MUST be present. The value id-kp-clientAuth
[RFC5280] MAY be present. The values id-kp-emailProtection [RFC5280], id-kpcodeSigning [RFC5280], id-kp-timeStamping [RFC5280], and anyExtendedKeyUsage
[RFC5280] MUST NOT be present. Other values SHOULD NOT be present.
>
>For Subordinate CA Certificates that are not used to issue TLS certificates, then the
value id-kp-serverAuth [RFC5280] MUST NOT be present. Other values MAY be
present, but SHOULD NOT combine multiple independent usages (e.g. including idkp-timeStamping [RFC5280] with id-kp-codeSigning [RFC5280]).

So unless my eyes are missing it, I think that since OCSP Signing is not explicitly enumerated that it would fall under the `Other values SHOULD NOT be present` clause, thus making it a warning?

v3/lints/cabf_br/lint_ca_invalid_eku.go
```diff
@@ -0,0 +1,81 @@
...
+	// This lint applies to any SubCA certificate to which the CABF BRs are applicable and which contains
+	// the EKU extension. Given that the lint source is lint.CABFBaselineRequirements, if we arrive here
+	// it's been already checked that the certificate falls within the purview of the CABF BRs.
+	return util.IsSubCA(c) && len(c.ExtKeyUsage) != 0
+}
+
+func (l *caInvalidEKU) Execute(c *x509.Certificate) *lint.LintResult {
+
+	// If the EKU contains anyExtendedKeyUsage, it's probably a cross-certicate
+	// In this case, the EKU must not contain any other value
```
christopher-henderson, line 58: I can see why you conditionally let `anyExtUsage` pass, however this does run contradictory to the clause...

> anyExtendedKeyUsage [RFC5280] MUST NOT be present. 

If the `the EKU must not contain any other value` is coming from CABF then that's problematic. However, if this come from a different doc I think that we should prefer the explicit failure spelled out by CABF here.

christopher-henderson (changes requested):

v3/lints/cabf_br/lint_ca_invalid_eku.go
mathewhodson: > So unless my eyes are missing it, I think that since OCSP Signing is not explicitly enumerated that it, so would fall under the `Other values SHOULD NOT be present` clause, thus making it a warning?

This list matches [7.1.2.10.6 CA Certificate Extended Key Usage](https://github.com/cabforum/servercert/blob/049237e096650fe01f67780b7c24bd5211ee3038/docs/BR.md#712106-ca-certificate-extended-key-usage).

christopher-henderson: @mathewhodson you are correct, it is enumerated in latest. This is where it gets mildly tricky since this lint is marked as being effective as of August 2020 via [CABF/BR 1.7.1](https://cabforum.org/uploads/CA-Browser-Forum-BR-1.7.1.pdf) where I don't think I see it marked as a `MUST NOT`.

@defacto64 did you have a particular reason for choosing 1.7.1?

v3/lints/cabf_br/lint_ca_invalid_eku.go
defacto64: Yes, I understand the doubts you may have regarding the reference version of the CABF BR for this lint. I myself was uncertain about it. The fact is, the ban on having certain other keyPurposes in the EKU of a TLS SubCA certificate was introduced in version 1.7.1 (section 7.1.2.2, subsection 'g'). However, the list of expressly prohibited keyPurposes has been extended in BRs 2.0.0 (section 7.1.2.10.6) by adding `ocspSigning`, hence the list of banned keyPurposes you see in my lint. So the requirement goes back "mostly" to BR 1.7.1, so to speak. In fact you can find EKU related incidents on https://bugzilla.mozilla.org in SubCA certificates that mention BR v1.7.1.

And as regards the first check you find in my lint, where we don't know with mathematical certainty but we can reasonably guess that we are analyzing a cross-certificate, there I refer to section 7.1.2.2.4 of the BR since v2.0.0 (Cross‑ Certified Subordinate CA Extended Key Usage - Unrestricted).

So how do you see it?

v3/lints/cabf_br/lint_ca_invalid_eku.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+type caInvalidEKU struct{}
+
+func NewCaInvalidEKU() lint.LintInterface {
+	return &caInvalidEKU{}
+}
+
+func (l *caInvalidEKU) CheckApplies(c *x509.Certificate) bool {
+	// This lint applies to any SubCA certificate to which the CABF BRs are applicable and which contains
```
toddgaunt-gs, line 49: Nit: This comment may be better above the function signature rather than inside the function so it shows up more easily using LSP tools found in most code editors.

defacto64: Okay.

christopher-henderson (approved):

PR #841, opened 2024-05-11
Add lint to check that precertificates do not contain an SCT list

defacto64:
This lint checks that the SCT (SignedCertificateTimestampList) extension is not present in precertificates, as this is not permitted by RFC 6962. This type of error has actually happened at least once as can be seen on https://bugzilla.mozilla.org.


v3/lint/source.go
```diff
@@ -32,6 +32,7 @@ const (
 	RFC5280                       LintSource = "RFC5280"
 	RFC5480                       LintSource = "RFC5480"
 	RFC5891                       LintSource = "RFC5891"
+	RFC6962                       LintSource = "RFC6962"
```
christopher-henderson, line 35: (I really wish Go had exhaustive pattern matching, because adding a new enum value usually results in usage hunt throughout most code bases...)

Scroll down to [source.go#L91](https://github.com/zmap/zlint/blob/208af03b5346578ba252fed89c93ceda0d6dc62e/v3/lint/source.go#L91) and add the following case

```go
	case RFC6962:
		*s = RFC6962
```

christopher-henderson: Also visit [source.go#L54](https://github.com/zmap/zlint/blob/208af03b5346578ba252fed89c93ceda0d6dc62e/v3/lint/source.go#L54) and add `RFC6962` to the long case statement.

```go
case RFC5280, RFC5480, RFC5891, CABFBaselineRequirements, CABFEVGuidelines, CABFSMIMEBaselineRequirements, MozillaRootStorePolicy, AppleRootStorePolicy, Community, EtsiEsi, RFC6962:
```

christopher-henderson: Add the following test case to [registration_test#L133](https://github.com/zmap/zlint/blob/208af03b5346578ba252fed89c93ceda0d6dc62e/v3/lint/registration_test.go#L133)

```go
		{
			name: "new lint source category",
			lint: &Lint{
				Name:   "sct",
				Lint:   func() LintInterface { return &mockLint{} },
				Source: RFC6962,
			},
			registry:      dupeReg,
			expectNames:   []string{egLint.Name, "sct"},
			expectSources: SourceList{egLint.Source, RFC6962},
		},
```

christopher-henderson (changes requested):
Just some code nuances since Go enums are so rich in functionality.

v3/lint/source.go
defacto64: Fixed the first two items, but the last one is causing some trouble in the "Go / Build and Unit Test " phase that right now I am not clear how to solve....

defacto64: I have slightly modified the additional test that you asked me to insert in registration_test.go, and the problem I had this morning is now solved, hopefully in the correct way.

christopher-henderson (approved):

PR #848, opened 2024-05-28
cabfOrganizationIdentifier extension for VAT and PSD based organizationIdentifiers cannot have referenceStateOrProvince

XolphinMartijn:


v3/lints/cabf_ev/lint_cabf_org_identifier_psd_vat_has_state.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterCertificateLint(&lint.CertificateLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "e_cabf_org_identifier_psd_vat_has_state",
+			Description:   "The cabfOrganizationIdentifier field for PSD org VAT Registration Schemes cannot include the referenceStateOrProvince field.",
+			Citation:      "Current PSD and VAT based identifiers are issued from Country level only. No state or province information is included.",
```
christopher-henderson, line 28: The citation is usually the section/paragraph/clause number of the given requirement (E.G `7.1.1.2a`).

Would you mind finding that and attaching it here? Its purpose is because the `Description` field is usually a paraphrasing and it is important for others to be able to quickly lookup the requirement in case a dispute arises.

Indeed, I tried to grep around in [CABF EV 1.7.0](https://cabforum.org/uploads/CA-Browser-Forum-EV-Guidelines-v1.7.0.pdf) in order to independently verify this lint but I couldn't find it's source (I assume that it is there, but it's important that these are fast and easy to look up).

christopher-henderson (changes requested):

v3/lints/cabf_ev/lint_cabf_org_identifier_psd_vat_has_state.go
XolphinMartijn: I've instead added the section, which is 7.1.4.2.8. 

The problem with this one is that it's not very clearly outlined. Section 7.1.2.2 is what this lint is based on, yet the definition of what goes where, is indirectly in 7.1.4.2.8, which shows that only NTR based orgIdentifiers have the state value. 

Let me know if you see a better way of showing this within the code

christopher-henderson: I apologize deeply, but I'm a bit confused about the citation numbers.

I certainly see this requirement in the [Ballot SC17](https://cabforum.org/2019/05/21/ballot-sc17-version-7-alternative-registration-numbers-for-ev-certificates/#motion-begins).

However, the entirety of section 7 in the EV guidelines is just two pages of document preamble.
image

I am quite sure that I'm staring something blank in the face, but I'm not seeing it.

XolphinMartijn: @christopher-henderson I just now see you're on V1.7.0. The EVGs were recently converted to the RFC3647 format, hence, I've used the current version's section numbers as reference:

https://github.com/cabforum/servercert/blob/main/docs/EVG.md#71428-subject-organization-identifier-field

For version 1.7.0, the relevant section number is 9.2.8. 

(This has me wondering, which numbering should be used, the section-at-time-of-ballot, or section-at-time-of-lint-creation?)

christopher-henderson: Thank you for pointing me in the right direction @XolphinMartijn!

> The EVGs were recently converted to the RFC3647 format

This may end up being problematic should CABF halt publishing they way that they have been all these years. However, ZLint generally makes references to the older format, so if you would not mind then I would appreciate that we cite `9.2.8`.

---

Since this lint is paraphrasing, I presume that this is the requirement that you are citing:

> For the NTR Registration Scheme identifier, if required under [Section 7.1.4.2.4](https://github.com/cabforum/servercert/blob/main/docs/EVG.md#71424-subject-jurisdiction-of-incorporation-or-registration-field), a 2 character ISO 3166-2 identifier for the subdivision (state or province) of the nation in which the Registration Scheme is operated, preceded by plus "+" (0x2B (ASCII), U+002B (UTF-8));

That is, NTR is the only scheme that explicitly allows the `State` filed. Thus, this lint derives the inverse that a `State` is not legal for either `VAT` nor `PSD` schemes, correct?

XolphinMartijn: >This may end up being problematic should CABF halt publishing they way that they have been all these years. 

Should the question arrise, I will advocate the need to keep old versions around ;)

>However, ZLint generally makes references to the older format, so if you would not mind then I would appreciate that we cite 9.2.8.

Sure. Done!

>That is, NTR is the only scheme that explicitly allows the State filed. Thus, this lint derives the inverse that a State is not legal for either VAT nor PSD schemes, correct?

That is correct indeed.

christopher-henderson (approved):

PR #860, opened 2024-06-20
Subscriber aia lints

mtgag:
This pull request adds a few lints from Section 7.1.2.7.7 of BR. This work has been done in close cooperation with D-Trust CA. We would be grateful if you could incorporate this PR into the main project.

Also while looking what is already implemented to avoid duplications, a few corrections have been made and a test has been added to an older lint (e_sub_cert_aia_does_not_contain_ocsp_url) to show the difference between the old lint which is a Pass and the new one (e_aia_ocsp_must_have_http_only) which is an Error for the same certificate.


v3/lints/cabf_br/lint_aia_must_contain_permitted_access_method.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/zmap/zcrypto/encoding/asn1"
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+type bRAIAAccessMethodAllowed struct{}
+
+/************************************************************************
+7.1.2.7.7 Subscriber Certificate Authority Information Access
```
christopher-henderson, line 29: :pray: thank for putting the full citation in comments. Perhaps I should start asking of this more regularly since I often spend a good chunk of my time hunting these down so that I can double check the interpretation.

christopher-henderson (approved):

PR #864, opened 2024-06-30
Add lint to check that Root CA and TLS SubCA certificates do not contain the OU subject attribute

defacto64:
Starting with v2.0, the CABF BRs prohibit the organizationalUnitName (OU) attribute in the Subject of Root CA and TLS SubCA certificates (they are tolerated, although not recommended, in other types of CA certs such as cross-certificates and non-TLS SubCA certificates). Zlint does not check this requirement as of today, so I propose this lint to fill the gap. 

This lint is configurable, thus allowing to safely distinguish a cross-certificate from a normal SubCA certificate. Non-TLS CA certificates and other uncommon types of CA certificates (to which the OU ban does not apply) are ignored via the pre-flight checks. To signal this lint that the certificate under examination is a cross-certificate, pass to zlint a TOML file containing the following directive:
```
[e_subj_orgunit_in_ca_cert]
CrossCert = true
```


christopher-henderson (commented):
Thank you! Especially for the effective use of configurable lints. There are a handful of features over the years that the community seemed interested in, but then rarely got used. So I am happy to see it used here.

I had only had a minor `nit` to update the description to remind operators that there is a flag that they can flip should they hit an error on a cross-signed cert.

defacto64:
I am fine with the retouches you suggested.

christopher-henderson (approved):

PR #874, opened 2024-08-07
Update main.go to have CRL linting lint on provided registry

joshschoenberg:
Currently, linting a CRL will run all lint sources. The proposed change makes it so that only the provided lint sources or lint names will be run when a CRL is linted.


christopher-henderson (approved):
Good catch! Thank you very much.

PR #884, opened 2024-10-03
Add lint for checking compliance with §7.1.2.10.5 of the BRs (CA Certificate Policies)

defacto64:
After reading incident https://bugzilla.mozilla.org/show_bug.cgi?id=1921597 and doing some quick checks, I realized that Zlint lacks a lint related to §7.1.2.10.5 of the BRs. So please add this lint to fill the gap. This lint verifies that a SubCA certificate contains either the AnyPolicy OID (alone) or at least one of the CAB Forum's Reserved Policy OIDs.


defacto64:
I really can't figure out what golangci-lint is complaining about, because the source was actually `gofmt`-ed with `-s`
Can someone give me a hand please?

v3/lints/cabf_br/lint_invalid_ca_certificate_policies.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+)
+
+func init() {
+	lint.RegisterCertificateLint(&lint.CertificateLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "e_invalid_ca_certificate_policies",
+			Description:   "Checks compliance with CABF section 7.1.2.10.5",
```
zakird, line 31: Can you put in a description in English of what's actually being checked rather than a reference to the BR?

v3/lints/cabf_br/lint_invalid_ca_certificate_policies.go
```diff
@@ -0,0 +1,93 @@
...
+func NewInvalidCACertificatePolicies() lint.LintInterface {
+	return &invalidCACertificatePolicies{}
+}
+
+func (l *invalidCACertificatePolicies) CheckApplies(c *x509.Certificate) bool {
+	return util.IsCACert(c) && !util.IsRootCA(c)
+}
+
+func (l *invalidCACertificatePolicies) Execute(c *x509.Certificate) *lint.LintResult {
+	if !util.IsExtInCert(c, util.CertPolicyOID) {
```
zakird, line 51: It doesn't look like this requirement is in `7.1.2.10.5` from what I can tell. Can you paste the requirement you're seeing?

v3/lints/cabf_br/lint_invalid_ca_certificate_policies.go
```diff
@@ -0,0 +1,93 @@
...
+		return &lint.LintResult{
+			Status:  lint.Error,
+			Details: "In a non-Root CA certificate, the CertificatePolicies extension is mandatory",
+		}
+	}
+
+	if len(c.PolicyIdentifiers) == 1 {
```
zakird, line 58: If I'm reading this BR text correctly, it's not forbidden to have only the `A Reserved Certificate Policy Identifier`. One `MAY` have additional policy identifiers.

v3/lints/cabf_br/lint_invalid_ca_certificate_policies.go
```diff
@@ -0,0 +1,93 @@
...
+		if !c.PolicyIdentifiers[0].Equal(util.AnyPolicyOID) {
+			return &lint.LintResult{
+				Status:  lint.Error,
+				Details: "If a CA cert contains only one policy OID, then it MUST be AnyPolicy",
```
zakird, line 62: I'm not seeing this specific text in the BR. Do you have a pointer?

v3/lints/cabf_br/lint_invalid_ca_certificate_policies.go
CBonnell: From TLS BR 7.1.2.10.5:

> Regardless of the order of `PolicyInformation` values, the Certificate Policies extension MUST contain exactly one Reserved Certificate Policy Identifier.

https://github.com/cabforum/servercert/blob/main/docs/BR.md?plain=1#L2974

This will be relaxed for cross certificates as part of SC-79 (assuming it passes).

v3/lints/cabf_br/lint_invalid_ca_certificate_policies.go
defacto64: Sure. 

First of all, 7.1.2.10.5 reads like follows in the case of un-restricted CAs (No Policy Restrictions):
> When the Issuing CA wishes to express that there are no policy restrictions, the Subordinate CA MUST be an Affiliate of the Issuing CA. The Certificate Policies extension MUST contain only a **single** PolicyInformation value, which MUST contain the **anyPolicy** Policy Identifier.

In the case of Policy-Restricted CAs, instead, the same section reads:

> The CA MUST include **at least one Reserved** Certificate Policy Identifier (see Section 7.1.6.1) associated with the given Subscriber Certificate type (see Section 7.1.2.7.1) directly or transitively issued by this Certificate ... [and] ...
> The **anyPolicy** Policy Identifier **MUST NOT** be present.

This amounts to saying, if I am not mistaken, that if CertificatePolicies contains just one OID then it must be anyPolicy; otherwise, _at least one_ of the included Policy OIDs must be a CABF reserved OID and _none_ of them can be anyPolicy (because this latter must be alone per the first rule in §7.1.2.10.5).

defacto64: I realized that my conclusion is wrong, because in fact it is allowed that the CP extension contains only one Policy OID as long as, in that case, it is a CABF reserved one. I was a bit too hasty, but in these things one must be more thoughtful :) 

I rewrote and retested my code, also adding more comments and one more test case.

defacto64:
I am closing this PR and opening a new one for better clarity and to facilitate review.

PR #890, opened 2024-10-24
Add lint to check that CRL does not have duplicates in RevokedCertificates

kowshikRoy:
Add lint to check that CRL does not have duplicates in RevokedCertificates

Though it's not specifically mentioned in the RFC(https://datatracker.ietf.org/doc/html/rfc5280#section-5.1.2.6), nonetheless, it's a good check to have so that CRL doesn't contain any redundant certs.


v3/lints/community/lint_crl_unique_revoked_certificate.go
```diff
@@ -0,0 +1,59 @@
...
+func (l *uniqueRevokedCertificate) CheckApplies(c *x509.RevocationList) bool {
+	return true
+}
+
+func (l *uniqueRevokedCertificate) Execute(c *x509.RevocationList) *lint.LintResult {
+	serials := make(map[string]bool)
+	for _, rc := range c.RevokedCertificates {
+		if serials[rc.SerialNumber.String()] {
+			return &lint.LintResult{
+				Status:  lint.Error,
```
christopher-henderson, line 52:
```suggestion
				Status:  lint.Warn,
```

v3/lints/community/lint_crl_unique_revoked_certificate_test.go
```diff
@@ -0,0 +1,57 @@
...
+	testCases := []struct {
+		name       string
+		path       string
+		want       lint.LintStatus
+		wantSubStr string
+	}{
+		{
+			name: "CRL with duplicates in revoked certificates list",
+			path: "crlWithDuplicatesInRevokedCertificateList.pem",
+			want: lint.Error,
```
christopher-henderson, line 36:
```suggestion
			want: lint.Warn,
```

christopher-henderson (changes requested):
Thank you for your contribution!

I agree with your premise. Although, since it is grounded in common sense rather than a direct citation, I have submitted a request to change this from an error to a warning.

christopher-henderson (approved):

PR #893, opened 2024-10-29
Add functionality to generate CRL in asn1 encoding

kowshikRoy:
Add functionality to generate CRL for testing purposes in ASN1 encoding.


christopher-henderson (approved):
Thank you for your contributions to the repo's tooling @kowshikRoy!

Both this and the certificate generation script are quick-and-dirty skeletons, so I'm not going to belabor code style or second guess byte literal constants like I typically would in lints. If there is some control flow or code cleanup that is possible then it will likely become evident when using the tooling more often.

PR #901, opened 2024-12-01
Add lint to check for a valid Registration Scheme in the Subject.organizationIdentifier of EV certificates

defacto64:
This lint verifies that the `organizationIdentifier` Subject attribute in an EV certificate, if present, meets the requirement that...

> Registration Schemes listed in Appendix H are currently recognized as valid under these guidelines.

This requirement was first introduced in section 9.2.8 of EVGL 1.7.0 and subsequently moved to section 7.1.4.2 when the EVGL were restructured.

Appendix H only allows three registration schemes: VAT, NTR, PSD. If the actual registration scheme found in the `organizationIdentifier` attribute is not one of those, this lint returns an Error.

Note: There are no such cases in the Zlint test corpus, however a search on Censys shows that some cases do exist.


v3/lints/cabf_ev/lint_invalid_orgid_reg_scheme.go
```diff
@@ -0,0 +1,68 @@
...
+	return util.IsEV(c.PolicyIdentifiers) && util.IsSubscriberCert(c) && c.Subject.OrganizationIDs != nil
+}
+
+func (l *invalidOrgIDRegistrationScheme) Execute(c *x509.Certificate) *lint.LintResult {
+
+	// Let's assume there is just one OrganizationID; if not so, it's not this lint's business to raise an alarm
+	orgId := c.Subject.OrganizationIDs[0]
+
+	runes := []rune(orgId)
+	firstThreeRunes := runes[:3]
```
christopher-henderson, line 56: Unfortunately, Go slicing is not particularly safe as this can panic the program (instead of doing something fancier like returning an zero-value or whatever).

[Go Playground example](https://go.dev/play/p/SS27SUTbr6L)

```go
func main() {
	fmt.Println("Hello, 世界")
	a := []rune("")
	fmt.Println(a[:3])
}
```

If you don't mind, could you protect against this case? Simply because ZLint doesn't recover from panics originated individual lints (although perhaps it should).

I reckon that this check would result in an error, since if the rune slice is smaller than 3 runes then it most certainly is not `NTR`/`VAT`/`PSD`.

v3/lints/cabf_ev/lint_invalid_orgid_reg_scheme.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (l *invalidOrgIDRegistrationScheme) CheckApplies(c *x509.Certificate) bool {
+	return util.IsEV(c.PolicyIdentifiers) && util.IsSubscriberCert(c) && c.Subject.OrganizationIDs != nil
+}
+
+func (l *invalidOrgIDRegistrationScheme) Execute(c *x509.Certificate) *lint.LintResult {
+
+	// Let's assume there is just one OrganizationID; if not so, it's not this lint's business to raise an alarm
+	orgId := c.Subject.OrganizationIDs[0]
```
christopher-henderson, line 53: Would you mind length checking this index as well? I know it's unlikely, but it's the wild west out.

christopher-henderson (changes requested):

v3/lints/cabf_ev/lint_invalid_orgid_reg_scheme.go
defacto64: I agree.

v3/lints/cabf_ev/lint_invalid_orgid_reg_scheme.go
defacto64: I am not sure what you mean, here. The `CheckApplies` function already checks that `c.Subject.OrganizationIDs != nil`, and based on zcrypto/x509/pkix/pkix.go it seems to me that this implies that OrganizationIDs contains at least one element, so I can always take its element [0] ... no?

christopher-henderson: Unfortunately, no. A nil-check protects against a null dereference (except not really because of...Go)**, but it does not protect against indexing into an empty slice.

```go
func main() {
	org := []byte{}
	if org != nil {
		// Explosion
		fmt.Println(org[0])
	}
}
```

** When derefencing a nil pointer to a slice, Go will automagically dereference a zero value instead.

```go
func main() {
	var org []byte = nil
	// true 
	fmt.Println(org == nil)
	// []
	fmt.Println(org)
}
```

Neat, but also confusing.

christopher-henderson: Thank you! I know it's `nitty`, but it always comforts me to know that local code protects itself and is not reliant upon safety hidden within another codebase.

christopher-henderson (approved):

PR #902, opened 2024-12-07
Add lint to check that EV certificates contain only allowed attributes in the Subject

defacto64:
I realized that Zlint currently does not check the correct composition of the Subject of EV certificates, so I propose this lint to fill the gap. This lint verifies that the subject of an EV certificate does not contain other attributes than those expressly allowed in the EV Guidelines (requirement introduced by the SC16 ballot which came into force on 16 April 2019).


v3/lints/cabf_ev/lint_extra_subject_attribs.go
```diff
@@ -0,0 +1,100 @@
...
+ * We also include the OU attribute here, even though it is now banned, because this lint
+ * deals with a more general requirement that came into force long before the OU ban,
+ * and there is already another lint that deals with the OU attribute specifically.
+ *
+ * The organizationIdentifier attribute is only permitted starting from 21-may-2019 (EVGL 1.7.0),
+ * which is slightly after SC16 came into force, however any certificates that contain this
+ * attribute and were issued before that date have long since expired, so it makes no difference.
+ */
+var allowedAttribs = map[string]bool{
+	"1.3.6.1.4.1.311.60.2.1.1": true, // joiLocalityName
```
christopher-henderson, line 64: Thank you for taking the time to clearly enumerate these. It must have been at least mildly tedious.

christopher-henderson (approved):

PR #907, opened 2024-12-20
Add lint to detect HTML entities in Subject attributes

defacto64:
Due to some strange programming error, sometimes it happens that the value of some Subject attribute ends up containing HTML entities, for example "\&amp\;" instead of the ampersand character ("&"). This has happened several times, mostly in the past but also recently (see Bugzilla). This lint looks for HTML entities in Subject attributes that might contain them, and raises an error if it finds any. 

Since a false positive cannot be completely ruled out (in some jurisdictions it may be allowed to register a company with a name that would trigger this lint), it is possible to bypass this lint through configuration. To do so, pass the path of a TOML file to zlint (via the -config option) with this content:
```
[e_subj_contains_html_entities]
Skip = true

```


v3/lints/community/lint_subj_contains_html_entities_test.go
```diff
@@ -0,0 +1,85 @@
...
+			want:  lint.Pass,
+		},
+		{
+			input: "html_entity_ok2.pem",
+			want:  lint.Pass,
+		},
+		{
+			input: "html_entity_ok3.pem",
+			config: `
+                [e_subj_contains_html_entities]
```
christopher-henderson, line 57: Thank for exercising this.

christopher-henderson (approved):
1. Thank you for your patience over the holidays.
2. This is a silly thing that some code somewhere has done.
3. I was trying to think of something better than a regex, like perhaps HTML decoding the string and seeing if it still matches the original. But that has its own complications too.
4. I do not find it likely that people have company names with [HTML entities](https://mateam.net/html-escape-characters/) in it, so I agree that this is likely fine.
5. Thank you for using the configuration infrastructure!

Issue #912, opened 2025-01-29
RSA-PSS Warning


Hagelkruys:
Hello, 

if parsing a certificate with RSASSA-PSS Signing Algorithm the zlint gives a warning "e_signature_algorithm_not_supported"

In "\v3\lints\cabf_br\lint_signature_algorithm_not_supported.go" it states the following.

	// The BRs do not forbid the use of RSA-PSS as a signature scheme in
	// certificates but it is not broadly supported by user-agents. Since
	// the BRs do not forbid the practice we return a warning result.
	// NOTE: The Mozilla root program policy *does* forbid their use since v2.7.
	// This should be covered by a lint scoped to the Mozilla source instead of in
	// this CABF lint.

That isn't correct anymore. 
* The baseline requirements allow RSASSA-PSS, see BR 7.1.3.2.1
* The Mozilla root program allows RSASSA-PSS, see https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/ 5.1.1

On the contrary, you should not use RSA with PKCS1v1.5 anymore, see https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile&v=6 - 1.5  Dealing with Legacy Algorithms - RSA with PKCS1v1.5 padding

zakird:
@Hagelkruys would you be willing to submit a PR to address this?

Hagelkruys:
I looked into creating an PR for this and found why the current implementation gives a warning for RSA-PSS.

In the referenced mozilla root program (v2.7) there was a change that RSA-PSS is not allowed as id in the SubjectPublicKeyInfo field, thats also in the current mozilla root program policy and BR.

> CAs MUST NOT use the id-RSASSA-PSS OID (1.2.840.113549.1.1.10) within a SubjectPublicKeyInfo to represent an RSA key.

The linting file above checks for the SigningAlgorithm, that is another field. so there seams to be a mixup of SigningAlgorithm and  SubjectPublicKeyInfo  identifier

Hagelkruys:
I removed the warning and added the signature algorithm to the pass list. 
PR ist here: https://github.com/zmap/zlint/pull/913

PR #913, opened 2025-01-31
update lingT_signature_algorithm_not_supported - support rsa-pss

Hagelkruys:
see https://github.com/zmap/zlint/issues/912

### DIFF
--- a/v3/lints/cabf_br/lint_signature_algorithm_not_supported.go
+++ b/v3/lints/cabf_br/lint_signature_algorithm_not_supported.go
@@ -38,17 +38,16 @@ var (
 		x509.SHA1WithRSA:   true,
 		x509.DSAWithSHA1:   true,
 		x509.ECDSAWithSHA1: true,
-	}
-	// The BRs do not forbid the use of RSA-PSS as a signature scheme in
-	// certificates but it is not broadly supported by user-agents. Since
-	// the BRs do not forbid the practice we return a warning result.
-	// NOTE: The Mozilla root program policy *does* forbid their use since v2.7.
-	// This should be covered by a lint scoped to the Mozilla source instead of in
-	// this CABF lint.
-	warnSigAlgs = map[x509.SignatureAlgorithm]bool{
+		// The BRs v2.1.2 allow the use of RSA-PSS as a signature scheme, see 
+		// 7.1.3.2.1 RSA 
+		// The Mozilla root program policy v2.9 allow RSA-PSS as a signature 
+		// scheme, see 5.1.1 RSA
 		x509.SHA256WithRSAPSS: true,
 		x509.SHA384WithRSAPSS: true,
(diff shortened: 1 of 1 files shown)


v3/lints/cabf_br/lint_signature_algorithm_not_supported.go
```diff
@@ -38,17 +38,16 @@ var (
...
-	// The BRs do not forbid the use of RSA-PSS as a signature scheme in
-	// certificates but it is not broadly supported by user-agents. Since
-	// the BRs do not forbid the practice we return a warning result.
-	// NOTE: The Mozilla root program policy *does* forbid their use since v2.7.
-	// This should be covered by a lint scoped to the Mozilla source instead of in
-	// this CABF lint.
-	warnSigAlgs = map[x509.SignatureAlgorithm]bool{
+		// The BRs v2.1.2 allow the use of RSA-PSS as a signature scheme, see 
+		// 7.1.3.2.1 RSA 
+		// The Mozilla root program policy v2.9 allow RSA-PSS as a signature
```
christopher-henderson, line 43: Just double checking these references.

From Mozilla 5.1.1.

> When a root or intermediate certificate's RSA key is used to produce a signature, only the following algorithms MAY be used, and with the following encoding requirements:
>
> ...
>
> RSASSA-PSS with SHA-256, MGF-1 with SHA-256, and a salt length of 32 bytes.
> The encoded AlgorithmIdentifier MUST match the following hex-encoded bytes:
>
> 304106092a864886f70d01010a3034a00f300d06096086480165030402010500a11c301a06092a864886f70d010108300d06096086480165030402010500a203020120

From CABF/Br 7.1.3.2.1

> The CA SHALL use one of the following signature algorithms and encodings. When encoded, the
AlgorithmIdentifier MUST be byte‐for‐byte identical with the specified hex‐encoded bytes.
>
> ...
>
> RSASSA‐PSS with SHA‐256, MGF‐1 with SHA‐256, and a salt length of 32 bytes:
> Encoding:
>
> 304106092a864886f70d01010a3034a00f300d06096086480165030402010500a11c301a06092a864886f70d010108300d06096086480165030402010500a203020120

christopher-henderson (changes requested):
Thank you for the patch! I double checked the citations and I believe this is correct. Now we need to get CICD passing.

1. We need to edit the unit test at [lint_signature_algorithm_not_supported_test.go#L44](https://github.com/zmap/zlint/blob/62639dffec8a0c6a4b82686bbc55ef535281b18a/v3/lints/cabf_br/lint_signature_algorithm_not_supported_test.go#L44) to be from `expected := lint.Warn` to `expected := lint.Pass`.
2. The linter wants us to run (from the repository root) `gofmt -w v3/lints/cabf_br/lint_signature_algorithm_not_supported.go`

PR #911, opened 2025-01-18
Correct effective date for e_ev_extra_subject_attribs

mathewhodson:
Update the comment to explain which version of the EVGs the list comes from. The comment wasn't accurate after commit cd7321158ebe64e35b7da0b1d2a1f16cc62f5a08

PR #903, opened 2024-12-09
fix: organizationUnitName is prohibited

XolphinMartijn:
A fix for #902, which had the prohibited OU field included in the allow list

PR #915, opened 2025-02-08
OU (2.5.4.11) is incorrectly omitted from the allow list in e_ev_extra_subject_attribs

christopher-henderson:
The [relevant date](https://cabforum.org/working-groups/server/extended-validation/guidelines/#relevant-dates) table of the EV CABF guidelines specifically references that §9.2.7 regarding the prohibition of `subject:organizationalUnitName (OID: 2.5.4.11)` takes effect on 2022-09-01 rather than inheriting the effective date of the parent revision.

| Compliance Date | Section(s) | Summary Description |
|----------------|-----------|---------------------|
| 2022-09-01    | [9.2.7]   | CAs MUST NOT include the `organizationalUnitName` field in the Subject. |

The lint's in-code documentation correctly identifies that it should not have an opinion on `subject:organizationalUnitName (OID: 2.5.4.11)` as that lint covers all other subject attributes and leaves OU to its own lint. However, we removed OU from the allow list of this lint in #903. For my part, I read the CABF guideline and agreed to the change because I had forgotten that there was wording elsewhere that overrides the parent document's effective date.

@XolphinMartijn and @defacto64 I would appreciate it if you would chime in on whether-or-not you may an argument against the above. If not then I will move forward with the change.

Thank you @mathewhodson for your PR at #911 which highlighted this issue.

### DIFF
--- a/v3/integration/config.json
+++ b/v3/integration/config.json
@@ -978,7 +978,7 @@
         "ErrCount": 1303
     },
     "e_ev_extra_subject_attribs": {
-        "ErrCount": 12279
+        "ErrCount": 63
     },
     "e_subj_contains_html_entities": {
         "ErrCount": 14
--- a/v3/lints/cabf_ev/lint_extra_subject_attribs.go
+++ b/v3/lints/cabf_ev/lint_extra_subject_attribs.go
@@ -51,15 +51,6 @@ func (l *extraSubjectAttribs) CheckApplies(c *x509.Certificate) bool {
 	return util.IsEV(c.PolicyIdentifiers) && util.IsSubscriberCert(c)
 }
 
-/*
- * We also include the OU attribute here, even though it is now banned, because this lint
- * deals with a more general requirement that came into force long before the OU ban,
- * and there is already another lint that deals with the OU attribute specifically.
- *
- * The organizationIdentifier attribute is only permitted starting from 21-may-2019 (EVGL 1.7.0),
- * which is slightly after SC16 came into force, however any certificates that contain this
- * attribute and were issued before that date have long since expired, so it makes no difference.
- */
 var allowedAttribs = map[string]bool{
 	"1.3.6.1.4.1.311.60.2.1.1": true, // joiLocalityName
 	"1.3.6.1.4.1.311.60.2.1.2": true, // joiStateOrProvinceName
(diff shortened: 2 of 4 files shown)


defacto64:
As it's reminded into the lint's code itself....

>  We also include the OU attribute here, even though it is now banned, because this lint
>  deals with a more general requirement that came into force long before the OU ban,
>  and there is already another lint that deals with the OU attribute specifically.

Since the lint has an effective date of `util.SC16EffectiveDate` (2019-04-16), when the OU was still allowed (it only became prohibited on 2022-09-01), I am not clear why @XolphinMartijn proposed to remove it from this lint's table of allowed subject attributes, and why the PR#903 was approved. Therefore I am in favour of restoring it.

As I pointed out in the in-code comments, there is already another lint that specifically deals with the OU prohibition (`lint_organizational_unit_name_prohibited.go`) based on its own and correct effective date (`util.CABFBRs_OU_Prohibited_Date`).

XolphinMartijn:
>Since the lint has an effective date of util.SC16EffectiveDate (2019-04-16), when the OU was still allowed (it only became prohibited on 2022-09-01), I am not clear why @XolphinMartijn asked to remove it from this lint's table of allowed subject attributes. I am entirely in favour of restoring it.
>@XolphinMartijn and @defacto64 I would appreciate it if you would chime in on whether-or-not you may an argument against the above. If not then I will move forward with the change.

I'm still in favor of not having OU in the allow list here. But I can live with it if you take a different direction ;)

I will expand on why I personally find it the better approach, which boils down to two items:

1. Confusion. Yes, there's a comment in the code that it isn't allowed by the BRs. But, we also know people don't always read the manual!.
2. Design Consistency. Now granted, I don't currently know what the design intent is here, and also, in this case it doesn't have too much effect. But IIUC, there's no such thing as a "EffectiveEndDate" in zlint. Now imagine there's a lint for an existing "MUST" requirement, but CABF changes that to a "MUST NOT". I would in such a case expect that zlint has to either update the existing lint, or remove it and add a new lint for the new requirement. Any such a change would not take into account any certificates issued prior to the change. 
Again, in this case that's not as much of an issue, but it is why I took that thought and applied it to this lint, especially as it was a newly added lint

defacto64:
> But IIUC, there's no such thing as a "EffectiveEndDate" in zlint

Actually, the `Lint` struct includes an `IneffectiveDate` field (very seldom used).

defacto64:
Apart from that, let's take for example this EV certificate:
https://search.censys.io/certificates/399d72aaffa8e653567aed2cb99311160dccf54ca8b4a589b8099e9b402fcb04

Since it was issued on 2022-06-23, when the OU was still allowed, IMO it's wrong to raise an error about the OU attribute being present in the Subject of this cert. This would not have happened if it was not for PR#903.

My understanding, and it seems completely logical to me, is that Zlint should detect if the certificate has any features that break the _rules that were in place at the time when the certificate was issued_. If a certificate that was good when it was issued later "becomes" bad (as the rules change over time), it would not be correct to dispute something with the issuing CA.

christopher-henderson:
> Yes, there's a comment in the code that it isn't allowed by the BRs. But, we also know people don't always read the manual!.

For my part, I actually spent well over an hour combing over the BRs trying to understand why the heck is in the allow list to begin with. All that work instead of scrolling up in the code file :face_with_head_bandage:

PR #916, opened 2025-02-10
Add lint to check CRLs for a valid nextUpdate as per CABF BRs

defacto64:
Already in the CABF TLS BRs v1.8.7 there was a requirement about the maximum lifespan (the time span between `thisUpdate `and `nextUpdate`) of CRLs, but until now there was no corresponding check in Zlint. The lint I am proposing here returns an error if this limit is exceeded in the CRL under examination. 

It should be noted that the BRs set different lifespan limits depending on whether the CRL covers Subscriber certs (max 10 days) vs. CA certs (max 12 months), but since it is not possible to distinguish the two cases without a context, I implemented the "Configure" interface in this lint. By default, this lint assumes that the input CRL is one that covers Subscriber certs; if you want to specify that it covers CA certs instead, you need to pass Zlint (via the -config option) a TOML file containing the following configuration:
```
[e_crl_next_update_invalid]
SubscriberCRL = false
```


v3/lints/cabf_br/lint_crl_next_update_invalid.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"fmt"
+)
+
+func init() {
+	lint.RegisterRevocationListLint(&lint.RevocationListLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "e_crl_next_update_invalid",
+			Description:   "For CRLs covering (EE|CA) certificates, nextUpdate must be at most (10 days|12 months) beyond thisUpdate",
+			Citation:      "Section 4.9.7 of BRs v1.8.7 (then section 7.2 since BRs v2.0.0)",
```
christopher-henderson, line 30: `(then section 7.2 since BRs v2.0.0)` Thank you :sob: it is super painful sometimes to go hunting for wording that may have changed just enough to sabotage a raw text search.

v3/lints/cabf_br/lint_crl_next_update_invalid_test.go
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+func TestCrlNextUpdateInvalid(t *testing.T) {
+
+	type Data struct {
+		input  string
+		config string
+		want   lint.LintStatus
+	}
+	data := []Data{
```
christopher-henderson, line 40: As always, thank you for the comprehensive test suite. These check out against my own expectations.

christopher-henderson (approved):

PR #931, opened 2025-03-17
Add lint to detect wrongly encoded diacritics due to UTF-8 mistaken for Latin-1

defacto64:
This lint addresses the error where a UTF8String field in a certificate (e.g. Subject:organizationName) contains, instead of a certain diacritic, the Latin-1 (or Windows-1252) equivalent of the individual bytes that make up the UTF-8 encoding of that diacritic. For example, instead of the single letter "ü" whose UTF-8 encoding is 0xC3 0xBC, we find the two characters "Ã" and "¼" (which correspond to 0xC3 and 0xBC in the Latin-1 character set, respectively).

This mix-up is likely caused by processing UTF-8 strings with obsolete and/or buggy software that mistakenly assumes them to be Latin-1 or Windows-1252 strings (or, at any rate, that they are made up of 1-byte characters).

Numerous "ever-trusted" certificates affected by this error can be found on Censys, most quite old but some issued as late as December 2023.

Even though it's a problem that can hardly go unnoticed, I think it is useful to introduce a lint for it.

This lint is based on the fact that the two-character sequences resulting from the wrong encoding are highly unlikely in the real names of organizations, localities, persons, etc., so their occurrence in some field of the Subject is an almost certain signal that the mix-up has occurred.


christopher-henderson (approved):
Naivety regarding encoding is perhaps right up there with off-by-one errors. The ASCII notion that `1 byte == 1 character` is both elegant and a death trap.

Issue #925, opened 2025-03-13
Possible bug:  unknown lint source in list: CABF_CS_BR


harrison314:
Hello, I've come across a potential bug when using the `-excludeSources` switch with `CABF_CS_BR`.
When using `zlint.exe -excludeSources=CABF_CS_BR certificate.der` I get an error:

```
time="2025-03-13T10:16:26+01:00" level=fatal msg="invalid -excludeSources: unknown lint source in list: \"CABF_CS_BR\""
```

But when listing the sources, CABF_CS_BR is available (`zlint.exe -list-lints-source`):
```
    Apple
    CABF_BR
    CABF_BR
    CABF_CS_BR
    CABF_EV
    CABF_SMIME_BR
    Community
    Community
    ETSI_ESI
    Mozilla
    RFC3279
    RFC5280
    RFC5280
    RFC5480
    RFC5891
    RFC6962
    RFC8813
```

**ZLint version 3.6.5**

christopher-henderson:
Thank you for the report @harrison314! I have corrected this in #934 and it will be fixed in the next release of ZLint,

PR #934, opened 2025-04-05
Corrected an issue with not all lint sources being considered correcty during filtering

christopher-henderson:
Addresses #925.

Additionally, the `list-sources` flag was printing duplicates because the same source is present across multiple lint types (cert, crl, and ocsp).

### DIFF
--- a/v3/lint/source.go
+++ b/v3/lint/source.go
@@ -53,7 +53,20 @@ func (s *LintSource) UnmarshalJSON(data []byte) error {
 	}
 
 	switch LintSource(throwAway) {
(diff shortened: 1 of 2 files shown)


v3/lint/source.go
```diff
@@ -71,28 +84,32 @@ func (s *LintSource) FromString(src string) {
 	// Trim space and try to match a known value
 	src = strings.TrimSpace(src)
 	switch LintSource(src) {
+	case RFC3279:
```
christopher-henderson, line 87: I moved these around to match the ordering of the enum itself.

PR #940, opened 2025-04-23
chore: add arm64 architecture support in goreleaser configuration

takumin:


takumin:
I apologize for creating a PR without any explanation.
I was worried about security issues if I continued using the old version, but unless an external source such as Dependabot points out an issue, it may not be necessary to take action.
I will close this PR as it will just be noise.
Thank you very much for providing us with such great software.

PR #958, opened 2025-05-30
Add lint to check for mandatory individual identity subject attributes in non-legacy IV and SV S/MIME certificates

defacto64:
CAB Forum's Baseline Requirements for S/MIME Certificates include the following provision:

> **Multipurpose and Strict Generation profiles SHALL include either subject:givenName and/or subject:surname, or the subject:pseudonym.**

...for both Individual-Validated and Sponsor-Validated certificates (see sections 7.1.4.2.5 and 7.1.4.2.6 of CABF SMIME BRs). 

I realized that Zlint was lacking a check for this, so I am proposing this lint to fill the gap.


christopher-henderson (approved):
Very clean lint, thank you.

PR #959, opened 2025-06-07
Add lint to check for multiple Reserved Policy Identifiers in Subordinate CA certificates

defacto64:
Made suspicious by a recent incident (see https://bugzilla.mozilla.org/show_bug.cgi?id=1963663), I discovered that Zlint is lacking a uniqueness check for reserved policy identifiers within non-policy-restricted TLS SubCA certificates. It is in fact required by the CABF BRs that there is exactly one Reserved Policy Identifier in such CA certificates. I am therefore proposing this lint to fill the gap.

Although Ballot SC083 (implemented in BRs v2.1.3) made it absolutely clear, IMO it was already quite clear in BRs v2.0.0 that in a TLS SubCA certificate, if it is a policy restricted SubCA, "_Regardless of the order of PolicyInformation values, the Certificate Policies extension MUST contain exactly one Reserved Certificate Policy Identifier_" (quoted from the BRs, section 7.1.2.10.5). Therefore I set the EffectiveDate of this lint to the publication date of BRs v2.0.0.


christopher-henderson (approved):
Great lint, thank you very much!

PR #964, opened 2025-06-11
Add lint to check CRL Number range

kowshikRoy:


v3/lints/cabf_br/lint_crl_number_range.go
```diff
@@ -0,0 +1,72 @@
...
+			Details: "CRL number extension is missing",
+		}
+	}
+	if c.Number.Cmp(big.NewInt(0)) < 0 {
+		return &lint.LintResult{
+			Status:  lint.Error,
+			Details: fmt.Sprintf("CRL number is negative: %v", c.Number),
+		}
+	}
+	crlNumberUpperBound := new(big.Int).Lsh(big.NewInt(1), 159)
```
christopher-henderson, line 64:
```suggestion
	crlNumberUpperBound := new(big.Int).Exp(big.NewInt(2), big.NewInt(159), big.NewInt(0))
```

If you wouldn't mind, I find this a bit more readable due being a closer mapping to the literal language. Were it not for the extra `modulus` parameter then it would be a direct mapping.

This [Go Playground](https://go.dev/play/p/-TDOMScmJ19) indicates that this generates the same number.

christopher-henderson (changes requested):
Just a small request on a `big` API usage.

PR #965, opened 2025-06-11
Add lint to check that revoked certificates in a CRL has revocation time before or equal to thisUpdate.

kowshikRoy:
…ime before or equal to thisUpdate.


christopher-henderson (changes requested):
Thank you! This looks good, although you have a merge conflict that needs to be cleared up.

PR #968, opened 2025-06-21
Add lint to check for HTTP URLs in the AIA extension of Subordinate CA certificates

defacto64:
This lint checks that, in the AIA extension of a Subordinate CA certificate, the URL (accessLocation) associated with the 'caIssuers' and 'ocsp' access methods be an HTTP URL (other schemes are not allowed by the CABF BRs). 

A similar lint already exists in Zlint, but - apart from having a rather misleading name - it is very old and consequently it just generates a _warning_ that might be ignored by some CA systems (depending on how they use Zlint). This lint, on the other hand, refers to CABF BR 2.0+ which only allow the "http" scheme, and therefore generates an _error_ in case the certificate fails to meet the requirement. This type of error has been made in the generation of some fairly recent CA certificates (see Bugzilla).


christopher-henderson (approved):
Thank you for the simple lint and for the thoughtful consideration towards backwards compatibility.

PR #972, opened 2025-07-01
set IneffectiveDate for w_sub_ca_aia_does_not_contain_issuing_ca_url

### LINKED ISSUE #956: w_sub_ca_aia_does_not_contain_issuing_ca_url is outdated
`w_sub_ca_aia_does_not_contain_issuing_ca_url` currently checks for "Subordinate CA Certificate: authorityInformationAccess SHOULD also contain the HTTP URL of the Issuing CA's certificate." and emits a warning if the aforementioned Access Method is not present. 

The presence of this Access Method is not a SHOULD anymore as observed in the TBRs v2.1.5, "7.1.2.10.3 CA Certificate Authority Information Access".

mhyder13:
This change is intended to fix #956. It sets an `IneffectiveDate` for this lint for when this requirement was removed from the BRs to maintain backwards compatibility. I've also included a test with a certificate with a notBefore one second after the `IneffectiveDate`. That test cert is a sample cert modified to trigger this lint. I ran manual before-and-after tests with it to verify that it does trigger the warning without the patch. I'm confident has sufficient fidelity for this use, but let me know if you'd like to clean up any any component to make it a bit more like a real CA cert.

I traced back the changes, and it looks like this requirement was removed from the BRs in [v2.0.0](https://cabforum.org/uploads/CA-Browser-Forum-BR-v2.0.0.pdf) as part of the profile rework. You can see the new MAY requirement that replaced it in § 7.1.2.10.3 and the old SHOULD in the previous [v1.8.7](https://cabforum.org/uploads/CA-Browser-Forum-BR-1.8.7.pdf) §7.1.2.2(c). (That section was also rewritten from the language cited in the lint in [v1.7.1](https://cabforum.org/uploads/CA-Browser-Forum-BR-1.7.1-redlined.pdf), but that revision kept the SHOULD requirement, now phrased as `It SHOULD contain the HTTP URL of the Issuing CA’s certificate (accessMethod = 1.3.6.1.5.5.7.48.2).`) Note that 2.0.0 is the revision immediately following 1.8.7.

### DIFF
--- a/v3/lints/cabf_br/lint_sub_ca_aia_does_not_contain_issuing_ca_url.go
+++ b/v3/lints/cabf_br/lint_sub_ca_aia_does_not_contain_issuing_ca_url.go
@@ -35,11 +35,12 @@ It SHOULD contain the HTTP URL of the Issuing CA’s certificate (accessMethod =
 func init() {
 	lint.RegisterCertificateLint(&lint.CertificateLint{
 		LintMetadata: lint.LintMetadata{
-			Name:          "w_sub_ca_aia_does_not_contain_issuing_ca_url",
-			Description:   "Subordinate CA Certificate: authorityInformationAccess SHOULD also contain the HTTP URL of the Issuing CA's certificate.",
-			Citation:      "BRs: 7.1.2.2",
-			Source:        lint.CABFBaselineRequirements,
-			EffectiveDate: util.CABEffectiveDate,
+			Name:            "w_sub_ca_aia_does_not_contain_issuing_ca_url",
+			Description:     "Subordinate CA Certificate: authorityInformationAccess SHOULD also contain the HTTP URL of the Issuing CA's certificate.",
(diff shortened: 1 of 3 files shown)


christopher-henderson (approved):
Thank you @mhyder13!

1. Thank you for tackling this issue.
2. Thank you for your patience over the holiday
3. Thank you for hunting down the precise diffs. Indeed, this activity is 90% of what I do in this repo and it's not the most fun thing in the world.

I agree that a deprecation of the lint on the `2.0.0` suffices for this change. That is to say, no new lint is necessary as the `MAY` language would result in a `PASS` in either scenario where the URL either does-or-does-not exist.

> ...make it a bit more like a real CA cert.

Thank you for considering this! The test certs are intended to be incredibl specific to the one requirement that they are exercising. As such, the rest of the cert can of any shape-and-content that is reasonable. For my part, I like to leave little easter eggs such as song lyrics or quotes from old cartoons.

PR #974, opened 2025-07-14
Replace CRL Distribution Points oid(2.5.29.31) with Issuing Distribution Point oid(2.5.29.28) when checking crl extension validity

whtung:
@kowshikRoy

Hi Repon,

May I have this change?
Replace CRL Distribution Points oid(2.5.29.31) with Issuing Distribution Point oid(2.5.29.28) when checking crl extension validity


kowshikRoy:
@christopher-henderson  Could you please review this CL?

christopher-henderson (approved):

Issue #973, opened 2025-07-14
e_ca_common_name_missing and Outdated Citations


mhyder13:
I've been reading through some of the older lints for the TLS BRs, and it looks like a bunch of them now have out of date section citations in their metadata. A lot of this is from the big refactor in 2.0.0, but for some of the original ones, they can be seen referencing not the current requirement, nor the original version of the requirement, but the section that was relevant when the lint was written. At this point, some of these (like `lint_ca_country_name_invalid`) are now referencing a section from an unclear prior version of the Requirements that is neither current nor original.

To take a very simple example from a newer lint, `e_ca_common_name_missing` was added for a provision added in v1.4.8. The existing metadata still lists the citation as "BRs: 7.1.4.3.1", which was the correct citation from v1.4.8. As a part of the certificate profile update in v2.0.0, this requirement was moved to §7.1.2.10.2 for all CA types except that cross-signs of existing CA keys were exempted under some circumstances.

The question is, what exactly _should_ the citation metadata now be for this lint? I think it would be most useful for the citation to be as up-to-date as possible, and for the prior citation to be moved to a comment in the lint file or documentation somewhere. This would both allow researchers and CAs that hit a lint failure to have a better starting off point when researching the issue while preserving the historical information for investigating older certificates.

More importantly, regardless of the end format, **is the project interested in receiving pull requests to update the citations** or would we rather leave these as-is?

(For `e_ca_common_name_missing`, I did check the redline for v1.4.8 and verified that this was the original version of the requirement. It was added as a new requirement and the existing subject name requirements for subordinate and root CAs were consolidated from §7.1.2.1(e) and §7.1.2.2(h) into the new section with commonName added as subsection a.)

christopher-henderson:
> is the project interested in receiving pull requests to update the citations 

Oh absolutely. As it stands, this would be an overwhelming task for one individual to accomplish.

Actually, now that I think about it, this might not be a bad use of LLMs.

1. Feed it the context of every version of BRs.
2. Ask it to find where in the latest BRs this older citation exists.
  2a. If it does not exist, when was it removed (so that we can add an `Inneffective` date to the lint).

Of course, every outcome would require scrutiny. But it is far more feasible than performing this task by hand.

mhyder13:
[...]

PR #976, opened 2025-07-24
update citation for e_ca_common_name_missing

mhyder13:
This is related to #973. This requirement comes in after most of the early reformatting of the BRs, so the citation string is only one movement out of date. I've added the old citation and the versions it was included in the comments. I've also added the requirements language this lint is (now) based on to the comment, which is similar to how some other lints have it. ([example](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_ca_country_name_invalid.go#L27))

One sticking point is that in the new version, there is a "Cross-Certified Subordinate CA Certificate Profile" intended to allow new cross-signs of existing compliant CAs even after the name rules are update with their original names. This profile allows certain CAs without commonName in the subjects of their existing certificates issued before 2017-06-08 to receive new certificates which are not subject to this requirement, provided that "the encoded subject name MUST be byte-for-byte identical to the
encoded subject name of the existing CA Certificate." (It also allows other deviations from the normal CA naming requirements.) The profile otherwise looks similar to the other CA profiles.

Unfortunately, there doesn't seem to be any definitive, distinctive element to this profile that can be used to detect it in zlint. Additionally, the factors that allow its usage like "the existing CA Certificate complied with the Baseline Requirements in force at time of issuance" aren't detectable from the certificate itself. For now, I'm proposing to leave the implementation of the lint as it is with a note about this profile in the comment. CAs attempting to use it should already be aware that they are using it to circumvent the normal naming rules. Research users will unfortunately have to live with this limitation or follow up to determine if a certificate should have been eligible for this profile. If anyone has an idea how to accurately filter these certificates out, we might want to split this lint to separate the two forms of the requirement. Failing that, maybe the configuration system can be leveraged to indicate when a certificate is using this profile?

### DIFF
--- a/v3/lints/cabf_br/lint_ca_common_name_missing.go
+++ b/v3/lints/cabf_br/lint_ca_common_name_missing.go
@@ -22,12 +22,44 @@ import (
 
 type caCommonNameMissing struct{}
 
+/*
+--- Citation History of this Requirement ---
+v1.4.8 to v1.8.7: 7.1.4.3.1a
+v2.0.0 to v2.1.6: 7.1.2.10.2
+
+--- Version Notes ---
+As of v2.0.0, this requirement no longer applies to CA certificates that conform to the
+Cross-Certified Subordinate CA Certificate Profile. This lint does not implement this exemption
+because it is impossible to identify certificates to which it applies from only the certificate.
+
+This requirement was baselined at v2.1.6 and is current.
+
+--- Requirements Language ---
+BRs: 7.1.2
+If the CA asserts compliance with these Baseline Requirements, all certificates that it issues MUST
+comply with one of the following certificate profiles
+
+[Each of the CA profiles, excepting the Cross-Certified Subordinate CA Certificate Profile,
+specifies the subject follows 7.1.2.10.2]
+
+BRs: 7.1.2.10.2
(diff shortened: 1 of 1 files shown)


christopher-henderson (approved):
Thank you so much @mhyder13! This sort of work is sorely needed, and I admit that I have just enough time to keep the repo alive and humming from a code-and-CICD stand point. So your work is here **_very_** much appreciated :pray: 

> Unfortunately, there doesn't seem to be any definitive, distinctive element to this profile that can be used to detect it in zlint.

You may find [ZLint's configurable lints infrastructure](https://github.com/zmap/zlint/blob/master/CONTRIBUTING.md#making-your-lint-configurable) to be useful.

```go
type MyLint struct {
    FrobbleTheWidgets bool
}

func (m *MyLint) Configure() interface{} {
	return m
}

func (m *MyLinkt) Execute(c *x509.Certificate) *lint.LintResult {
	if m.FrobbleTheWidgets {
	} else {
		return &lint.LintResult{Status: lint.Error, Details: "Widgets MUST be frobbled."}
	}
}

```

The operator can then define a `TOML` file as the following...

```toml
[MyLint]
FrobbleTheWidgets = true
```

And invoke ZLint vis-a-vis this configuration...

```bash
zlint -config myConfig.toml mycert.pem
```

PR #979, opened 2025-08-02
update CA countryName lints' citations

mhyder13:
This history on these two lints goes back quite a bit further than the last one, but I went ahead and included the full history. The early BR versions inexplicably express this requirement as being on the issuer field, at least until I guess someone remembered that the issuer field should match the subject of the CA certificate. Given that the fix in v1.4.8 just rephrases the existing wording against the CA's subject, I think that language was always intended to be interpreted as CA subject restriction and I've left that in the history.

I know the project favors many smaller CRs over large ones, but I've combined these two together because they have always been derived from the same language in the BRs and their history is identical.

These two lints, and also `e_ca_organization_name_missing` are both affected by the issue with the cross-signed CA profile I mentioned in #976 and this change does not address that. I'll open a new issue to track that problem as separate from the rest of the requirements language change.

### DIFF
--- a/v3/lints/cabf_br/lint_ca_country_name_invalid.go
+++ b/v3/lints/cabf_br/lint_ca_country_name_invalid.go
@@ -23,11 +23,37 @@ import (
 type caCountryNameInvalid struct{}
 
 /************************************************
-BRs: 7.1.2.1e
-The	Certificate	Subject	MUST contain the following:
-‐	countryName	(OID 2.5.4.6).
-This field MUST	contain	the	two‐letter	ISO	3166‐1 country code	for	the country
-in which the CA’s place	of business	is located.
+--- Citation History of this Requirement ---
+v1.0 to v1.2.4:   9.1.4
+v1.2.5:           Appendix B 1E, 2H
+v1.3.0 to v1.4.7: 7.1.2.1e, 7.1.2.2h
+v1.4.8 to v1.8.7: 7.1.4.3.1c
+v2.0.0 to v2.1.6: 7.1.2.10.2
+
+--- Version Notes ---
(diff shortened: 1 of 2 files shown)


mhyder13:
Side note, I know the readme says the original tests were based on v1.4.8, and that would make sense for the timing of the paper, but the original citation for these two lints were definitely from v1.4.7 or earlier. I went back and looked and these citations match the last version of the code before it was imported to this repo, and that predates the publication of 1.4.8 by close to a year. I also pulled up as much of the notes as I can find from the time and I can't find exactly what version of the BRs was use for the initial read-through. I also don't see the the CA common name lint in the original project list of lints. That lint was added [here](https://github.com/zmap/zlint/commit/6aae9006637ed745e65a7acedbad2cc1cdfbaa60), after the import.

I guess this brings up what "Historically ZLint was focused on only [RFC 5280](https://www.ietf.org/rfc/rfc5280.txt) and [v1.4.8](https://cabforum.org/wp-content/uploads/CA-Browser-Forum-BR-1.4.8.pdf)" really means in practice. I think it was written against earlier versions (probably 1.3.9, with 1.4.0 published during the first pass of zlint being worked on) and 1.4.8 was added about a year later once it was already moved to this public repo. I guess that's probably the revision it was updated to when the paper for it was written. My only real point here is that the citations in the lint files of the original lints are potentially citing to one of several of older versions. So even for the really old lints, we may not know the exact version because they weren't all 1.4.8 and while I'm sure someone checked that the requirements were still in force on each update, the citations were clearly not all updated for each change during the project.

There's no real practical implications of this, but it threw me for a loop when tracing how this lint ended up with a citation that was older than v1.4.8 and I thought I'd share.

christopher-henderson (approved):

PR #980, opened 2025-08-03
Update cab_iv_requires_personal_name lint to only require Personal Name

digirenpeter:
Just a small change for this lint, as of CABF BR v2.0.0, organizationName is a NOT RECOMMENDED Subject attribute, and so the old language no longer applies:

    Such Certificates MUST also include either organizationName or both givenName and
    surname, localityName (to the extent such field is required under Section 7.1.4.2.2),
    stateOrProvinceName (to the extent required under Section 7.1.4.2.2), and countryName
    in the Subject field.

This has been updated with a new table in BR v2.0.0:

    The following table details the acceptable AttributeTypes that may appear within the
    type field of an AttributeTypeAndValue, as well as the contents permitted within the
    value field.
    Attribute Name | Presence | Verification
    ------------------ | ---------------- | ----------------------
    organizationName | NOT RECOMMENDED | Section 3.2.3, pg. 111
    surname | MUST | Section 3.2.3
    givenName | MUST | Section 3.2.3

I'm only showing the relevant part of the table for these changes.


v3/lints/cabf_br/lint_cab_iv_requires_personal_name.go
```diff
@@ -23,24 +23,30 @@ import (
...
 	lint.RegisterCertificateLint(&lint.CertificateLint{
 		LintMetadata: lint.LintMetadata{
 			Name:          "e_cab_iv_requires_personal_name",
-			Description:   "If certificate policy 2.23.140.1.2.3 is included, either organizationName or givenName and surname MUST be included in subject",
-			Citation:      "BRs: 7.1.6.4",
+			Description:   "If certificate policy 2.23.140.1.2.3 is included givenName and surname MUST be included in subject",
+			Citation:      "BRs: 7.1.2.7.3",
 			Source:        lint.CABFBaselineRequirements,
-			EffectiveDate: util.CABV131Date,
+			EffectiveDate: util.CABFBRs_2_0_0_Date,
```
mhyder13, line 49: This is going to break zlint output for certificates issued before 2023-09-15 which might affect some use cases like crt.sh. I think it might make more sense here to mark the existing lint ineffective at `CABFBRs_2_0_0_Date` and copy this code into a new lint that enforces only the new, unmixed version of the requirement.

christopher-henderson: > Is there a plan or policy for how long to keep ineffective lints around before cleaning them up or is the plan to always be backwards compatible with old requirements?

I suppose that that depends on...

1. What is out there in the wild
2. The audience of ZLint

For #1, as certs expire their sunsetted requirements will naturally become irrelevant. That'll be much easier over the course of the next decade as shorter-and-shorter lifecycles not only become the norm, but are enforced. In the meantime, I would not be surprised to find many 10 year+ certificates.

For #2, it depends on whether the audience of ZLint cares to scrutinize old certificates with such long life cycles. I would hazard a guess that most industry users are more concerned with the now-rather-than-the-then. Although researchers may have an interest in the "wild side" of the older web PKI.

christopher-henderson (approved):
`lgtm`. Thank you to @mhyder13 for catching the backwards compatibility, I'll let you chime-in again before merging.

PR #981, opened 2025-08-10
update language and citations for e_ca_organization_name_missing

mhyder13:
This is the last of the lints for the MUST include requirements for CA subjects. Like the other three, this PR only fixes the citation and the documentation in the comments. I'll open an issue to track fixing the cross-signed CA profile exemption separately.

The history of the organizationName requirement follows essentially the same history as the countryName requirements in #979, usually occurring in the same or an adjacent section. I've put it in a separate PR to highlight that most of the references are one character off so it doesn't get lost in the mix.

Summary:  
* added full requirement history for all previous versions
* replaced old governing language in the comments with the current table and language
* updated the Citation string to the current section
* fixed a typo in the Description string
* added a note about the limitations in handling cross-signed CAs with legacy subjects

### DIFF
--- a/v3/lints/cabf_br/lint_ca_organization_name_missing.go
+++ b/v3/lints/cabf_br/lint_ca_organization_name_missing.go
@@ -23,16 +23,51 @@ import (
 type caOrganizationNameMissing struct{}
 
 /************************************************
-BRs: 7.1.2.1e
-The Certificate Subject MUST contain the following: organizationName (OID 2.5.4.10): This field MUST be present and the contents MUST contain either the Subject CA’s name or DBA as verified under Section 3.2.2.2.
+--- Citation History of this Requirement ---
+v1.0 to v1.2.4:   9.1.3
+v1.2.5:           Appendix B 1E, 2H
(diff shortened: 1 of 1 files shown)


v3/lints/cabf_br/lint_ca_organization_name_missing.go
```diff
@@ -23,16 +23,51 @@ import (
 type caOrganizationNameMissing struct{}
 
 /************************************************
-BRs: 7.1.2.1e
-The Certificate Subject MUST contain the following: organizationName (OID 2.5.4.10): This field MUST be present and the contents MUST contain either the Subject CA’s name or DBA as verified under Section 3.2.2.2.
+--- Citation History of this Requirement ---
+v1.0 to v1.2.4:   9.1.3
+v1.2.5:           Appendix B 1E, 2H
```
christopher-henderson, line 28: I am VERY impressed that you have managed to track down not only the latest citation, but what is essentially the migration history as well.

christopher-henderson (approved):

Issue #983, opened 2025-08-12
Missing oid 2.5.29.60 (expiredCertsOnCRL) in allowed crl extension


certigna:
Hello,

I think OID 2.5.29.60 (expiredCertsOnCRL) is missing from the list of authorized extensions in the CRL.
We use this extension for a French authority.
Furthermore, adding extensions to the CRL is not prohibited in BRs, it's just not recommended.

Best regards
Jérôme

christopher-henderson:
Resolved in #984

PR #984, opened 2025-08-13
(#983) Fix CRL extensions lint

XolphinMartijn:
This update includes:
- Removal of crl_with_wrong_critical_for_crl_distribution.pem and the accompanying test case
  - This test case appears to be incorrect. It tests the criticality of the CRL Distribution Point, however the BRs do not state any requirements on the criticality of this extension. I expect this was intended to test the Issuing Distribution Point criticality, but the test CRL does not include that extension
- Mark any other not explicitly allowed extension as a NOT RECOMMENDED with accompanying WARNING response from zlint. The current code returns an Error response, however the BRs do not disallow other extensions


certigna:
Thanks a lot

christopher-henderson (approved):
Thank you @XolphinMartijn 

@kowshikRoy a heads-up that I agree with this change. At the heart of the issue is that the lint explicitly enumerated a list of discouraged extensions and (properly) issued a warning if they were present. However, the BRs states that this should be the behavior for `Any other value`, not just those specific values that were explicitly outlined.

In the test case changed in this PR, `cRLReason` changes from being an error to being a warning.

aarongable (commented):
This test case should not have been removed; it instead should have been replaced with a CRL that has the proper `issuingDistributionPoint` (OID 2.5.29.31) extension, instead of the improper `crlDistributionPoints` (OID 2.5.29.28) extension, which is used in Certificates.

This test fix was missed in https://github.com/zmap/zlint/pull/974, which only updated the happy-path test case.

PR #986, opened 2025-08-23
 Add lint to check encoding of SubjectPublicKeyInfo.AlgorithmIdentifier in S/MIME certificates

defacto64:
This lint checks that the DER encoding of the AlgorithmIdentifier (within SubjectPublicKeyInfo) is allowed, namely that it complies with section 7.1.3 of CABF SMIME BRs, also taking into account the allowance for PQC algorithms introduced on August 22 (see CABF SMIME BRs version 1.0.11).


v3/lints/cabf_smime_br/lint_invalid_spki_algoid.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// Let's first try with the legacy algorithms
+	for _, algoId := range allowedAlgoIds {
+		if bytes.Equal(spki.Algorithm.FullBytes, algoId) {
+			return &lint.LintResult{Status: lint.Pass}
+		}
+	}
+
+	// Let's now try with the PQC algorithms
+	if !c.NotBefore.Before(pqcEffectiveDate) {
```
christopher-henderson, line 113: Having two efficacy dates within one lint hints to me that perhaps this should be two lints. It's a bit of a pain-in-the-butt, but would you mind splitting the PQC algorithms into a separate lint?

My reasoning is that if/when the legacy algorithms are marked `Ineffective` that we will cease to check for PCQ algorithms that were created after the sunsetting of the legacy algorithms.

It's tough to visualize, but this is what I mean,

* PQC accidentally gets sunsetted along with the legacy algorithms.
```
          |------| 
legacy --> pqc --> ineffectiveLegacy ---> .....
```

* As a separate lint, PQC keeps going after quantum computing and AGI makes this all a pointless as we are now either in a utopia or in the Matrix.
```
          |---------------------------------------------| 
legacy --> pqc --> ineffectiveLegacy ---> .....
```

christopher-henderson (changes requested):
I went down this whoooole rabbit hole about how we have another [lint](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_subject_public_key_info_improper_algorithm_object_identifier_encoding.go) that checks for this, and how I can't find any reference to the Edward curves. and also where are you seeing ANY reference at all to the PQC algorithims...and...and....

> Oh, this is the SMIME BRs :melting_face: 

Thank you! I do have one request, though, to split the PQC algorithms into a separate lint as to avoid accidentally sunsetting them with the legacy algorithms.

v3/lints/cabf_smime_br/lint_invalid_spki_algoid.go
defacto64: I am a bit puzzled. The logic behind this lint is to take the binary encoding (DER) of the SPKI algorithm and see if it appears in a list of allowed encodings (a white list). If I split the PQC algorithms into a separate lint, as you suggest, it's not clear to me how can I _avoid_ to return an error — from this separate lint — if the algo's binary encoding doesn't appear in the relevant white list but nonetheless ie legit (as in the case of a legacy algo). And similarly, in reverse, for the first lint, which is supposed to only handle legacy algos. In short, it's not clear to me how to do what you're asking.

defacto64:
OK. So, first, I've modified this lint to have the effective date of SMIME BR 1.0.11, which introduced the PQC algorithms, and thus removed the additional check for the certificate issuance date I'd previously added. I've also trimmed down the test data a bit, skipping some unnecessary tests, and at this point I think this lint meets your expectations for the current stage (post-SMIME BR 1.0.11 release). If you're okay with it, I'll later submit a PR for a second lint called "... _legacy" that will only address the algos that were allowed in the previous stage.

defacto64:
@christopher-henderson While I was updating this PR, I didn't realize you'd decided to take the initiative and edit my lint yourself (and then submit a separate PR). It's all the same to me; go ahead as you prefer. If you like, I'll close this PR.

christopher-henderson:
Thank you @defacto64! I've merged #989 so I'll close this one out.

PR #987, opened 2025-08-27
Add cross-sign configuration for CA name tests

### LINKED ISSUE #982: Some CA subject lints improperly handle cross-signed CAs
In the major certificate profile rework added to the TLS BRs in v2.0.0, a new category of CA certificate was created that is exempt from the current subject naming requirements for CA certs. These "Cross-Certified Subordinate CA Certificates" (see § 7.1.2.2) are intended to allow new cross-signs to be issued for existing CA keys where the original name no longer follows the current naming rules. They instead fall under the requirement "the encoded subject name MUST be byte-for-byte identical to the encoded subject name of the existing CA Certificate."

Zlint is designed to do static analysis on a single signed artifact and doesn't have the context required to determine if a certificate is using the same key as an existing CA or if the subject matches the existing certificates for that key. One of the affected lints, [e_subj_orgunit_in_ca_cert](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_subj_orgunit_in_ca_cert.go), implements this exemption by using the configuration system to let the user provide the context that the supplied certificate is intended as a cross-signed CA. (Though that lint marks cross-signs as passed, I'd argue it should probably be `NA`.) I also found [e_ca_multiple_reserved_policy_oids](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_ca_multiple_reserved_policy_oids.go) which implements a similar exemption in a similar fashion.

I've found four more lints from before v2.0.0 that enforce older CA subject restrictions that have not yet been updated to handle this exemption and so incorrectly fail these certificates:
* [e_ca_common_name_missing](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_ca_common_name_missing.go)
* [e_ca_country_name_missing](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_ca_country_name_missing.go)
* [e_ca_country_name_invalid](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_ca_country_name_invalid.go)
* [e_ca_organization_name_missing](https://github.com/zmap/zlint/blob/master/v3/lints/cabf_br/lint_ca_organization_name_missing.go)

I'm currently thinking we can probably fix these lints along the same lines, but they have the added complication that the exemption only applies to new cross-signed CAs issued after 2023-09-15 whereas `e_subj_orgunit_in_ca_cert` has always had the exemption. Questions for the group: How have we typically handled requirements that have been narrowed after a date? Would having the `CheckApplies()` function also enforce the start of the exemption period when checking the configuration be acceptable or is that too hacky? Given that even new cross-signs MAY set their notBefore as early as "the earliest notBefore date of the existing CA Certificate(s)" does it even make sense to check the date? I suppose the issuance date could also asserted in the config to handle this.

mhyder13:
## Notes
This adds a new source level configuration to allow the user to specify that the input is a "Cross-Certified Subordinate CA Certificate" under the profile in § 7.1.2.2 to support proper evaluation of the subjects of those certificates. Unlike all the other CA profiles, these certificates are exempt from the CA naming rules in § 7.1.2.10.2 and instead have their own naming rules in § 7.1.2.2.2 that require only that they be byte-for-byte identical to the subject of the previous CA certificate. The only restriction is that the previous certificate must have been "issued in compliance with the then-current version of the Baseline Requirements." There's no way for zlint to check this or what naming requirements existed at the time, so instead it treats the CA naming rules as not applicable.

The reason this configuration is required is that there is no way to detect a cross-signed CA from the certificate itself. The only way to determine this would be a global search for matching CA certificates, which is clearly outside the scope of a static analysis tool like zlint. Instead, this allows the user (which might be the CA organization which issued it) to specify which rule set should apply based on outside context.

* Added a new source level configuration for the (TLS) Baseline Requirements to allow the user to specify that the input is a "Cross-Certified Subordinate CA Certificate" under the profile in § 7.1.2.2
  * This is not a global config: I've limited the scope to only the TLS BRs where this concept comes from. 
  * I made this a boolean because TOML and the zlint configuration system don't really support enumerated types or another similar way to specific which profile applied. Each lint independently evaluates the configuration based on which parts it wants to consider, so enforcing consistency in a string specifying the profile would be difficult.
  * This is the first source-level config to be added to zlint, but this applies to at least these 4 lints and probably at least two more which have this configuration at the individual level and could probably use to be set in a single place. This profile also ignores the normal rules for the certificatePolicies and extendedKeyUsage extensions and the encoding rules for name types, so it's likely it will be used in more places in the future.
* Removed a redundant entry for `CABFBaselineRequirementsConfig` from the global config list.
  * This looks like a bug that was included in the original configuration system commits that was being covered by the deduplication logic of the global configs and the fact that nothing was actually using this feature yet.
* Added new logic to `e_ca_common_name_missing`, `e_ca_country_name_invalid`, `e_ca_country_name_missing`, and `e_ca_organization_name_missing` to exempt CA certificates where the user indicates that they are cross-signed CAs
  * I think it makes more sense to mark these as `NA` than `PASS` because they don't actually pass the checks,  the section those checks are based on just doesn't apply to them, similar to how it doesn't apply to subscriber certificates.
* Updated the comments in that affected lints to indicate that cross-signed CA cert are now supported via a config setting and making note of the limitation on checking whether the certificates we issued during the period where such certs were/are allowed.
  * As noted in the comment, there's no way to tell when the certs were actually signed because CAs are permitted to back-date the notBefore field.
  * I decided against trying to add any way to specify when a certificate was actually issued because it would either require a massive change across zlint in how it processes issuance time, or else would be misleading/confusing in what it actually applies to. The limited value gained here doesn't seem to justify a change of that scale.
  * This is a two-way door that shouldn't constrain future development either way.
* Added some extra testing to the countryName lints, including a test where countryName is actually missing from the subject
  * For some reason, the name missing lint was being tested on a certificate that had a countryName in its subject sequence but where the name was an empty string. This seems more like an invalid name than a missing name to me so I've added a new test case that tests a certificate that actually has no countryName in its subject. I've also added the blank name cert as a test case to the `e_ca_country_name_invalid` test cases as an empty string is clearly an invalid name.
  * These tests aren't related to the rest of the change, but I wanted a more clearly appropriate certificate for the new tests I was adding for `e_ca_country_name_missing` anyway, so I included them here.

## Testing
* Updated the unit tests for the global config printer
* Updated tests for each of the affected lints with three test cases:
  * The config is present but set to false when the CA cert fails the test
  * The config is present and set to true which makes the test not applicable when it would otherwise fail
  * The config is present and set to true which makes the test not applicable when it would otherwise pass
  * The cases where the config is not present are retained from before this change.
* Added a new test certificate which is actually missing a RelativeDistinguishedName with a countryName AttributeTypeAndValue in the subject sequence because this is closer to what `e_ca_country_name_missing` is intended to test for
  * (It was modified from the existing test certificate, so the values are from that.)

## Document References
(Direct section fragment links work in Chrome, YMMV.)
* I developed these updates based on the then-current version 2.1.6, but 2.1.7 (released yesterday) doesn't make any changes to the certificate profiles and should not affect the correctness of these changes.
* See "Cross-Certified Subordinate CA Certificate Profile" in [§ 7.1.2.2](https://cabforum.org/working-groups/server/baseline-requirements/documents/CA-Browser-Forum-TLS-BR-2.1.7.pdf#page=93&zoom=100,92,509)
* See also, "Cross-Certified Subordinate CA Naming" in [§ 7.1.2.2.2](https://cabforum.org/working-groups/server/baseline-requirements/documents/CA-Browser-Forum-TLS-BR-2.1.7.pdf#page=94&zoom=100,92,465)
* The normal CA naming rules are in [§ 7.1.2.10.2](https://cabforum.org/working-groups/server/baseline-requirements/documents/CA-Browser-Forum-TLS-BR-2.1.7.pdf#page=129&zoom=100,92,96)

## Related Items
* This fixes #982

### DIFF
--- a/v3/lint/configuration_test.go
+++ b/v3/lint/configuration_test.go
@@ -934,6 +934,7 @@ func TestPrintConfiguration(t *testing.T) {
 [AppleRootStorePolicyConfig]
 
 [CABFBaselineRequirementsConfig]
+CrossSignedCa = false
 
 [CABFEVGuidelinesConfig]
 
--- a/v3/lint/global_configurations.go
+++ b/v3/lint/global_configurations.go
@@ -66,9 +66,11 @@ func (r RFC5891Config) namespace() string {
 // CABFBaselineRequirementsConfig is the higher scoped configuration which services as the deserialization target for...
 //
 // [CABFBaselineRequirementsConfig]
+// CrossSignedCa = false # Used to indicate that the certificate is a Cross-Certified Subordinate CA
 // ...
-// ...
-type CABFBaselineRequirementsConfig struct{}
+type CABFBaselineRequirementsConfig struct {
+	CrossSignedCa bool
+}
 
 func (c CABFBaselineRequirementsConfig) namespace() string {
 	return "CABFBaselineRequirementsConfig"
@@ -143,7 +145,6 @@ type GlobalConfiguration interface {
 // out a TOML document that is the full default configuration for ZLint.
 var defaultGlobals = []GlobalConfiguration{
 	&Global{},
-	&CABFBaselineRequirementsConfig{},
 	&RFC5280Config{},
 	&RFC5480Config{},
 	&RFC5891Config{},
--- a/v3/lints/cabf_br/lint_ca_common_name_missing.go
+++ b/v3/lints/cabf_br/lint_ca_common_name_missing.go
@@ -20,17 +20,17 @@ import (
 	"github.com/zmap/zlint/v3/util"
 )
 
-type caCommonNameMissing struct{}
-
 /*
 --- Citation History of this Requirement ---
 v1.4.8 to v1.8.7: 7.1.4.3.1a
 v2.0.0 to v2.1.6: 7.1.2.10.2
 
 --- Version Notes ---
-As of v2.0.0, this requirement no longer applies to CA certificates that conform to the
-Cross-Certified Subordinate CA Certificate Profile. This lint does not implement this exemption
-because it is impossible to identify certificates to which it applies from only the certificate.
+As of v2.0.0, this requirement no longer applies to CA certificates that conform to the Cross-Certified
+Subordinate CA Certificate Profile. This lint uses the global CrossSignedCa setting to determine if the
+certificate under test should be treated as an exempt cross-signed CA. It does not attempt to determine
+if the issuance date is after CABFBRs_2_0_0_Date because CAs are permitted to back-date the notBefore
+to that of the earliest existing certificate under section 7.1.2.2.1
 
 This requirement was baselined at v2.1.6 and is current.
 
@@ -54,6 +54,10 @@ field of an AttributeTypeAndValue, as well as the contents permitted within the 
 +----------------+----------+----------------------------------------------------------+----------------+
 */
 
+type caCommonNameMissing struct {
+	TlsBrConfig *lint.CABFBaselineRequirementsConfig
+}
+
 func init() {
 	lint.RegisterCertificateLint(&lint.CertificateLint{
 		LintMetadata: lint.LintMetadata{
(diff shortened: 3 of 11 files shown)


christopher-henderson (approved):
This is a great use of the configuration framework!

And thank you for the extremely thorough due diligence and explanation. It is a refreshing change of pace for me as, in my duties elsewhere, I have grown numb to change lists which feature completely empty PR bodies and zero code comments or rationale.

PR #992, opened 2025-09-18
Allowed subjectdn attributes

mtgag:
Deutsche Telekom Security brought to my attention an issue with e_subject_country_name, suggested providing a fix, and also sponsored the development. The lint currently checks the value of countryName only for mailbox-validated certificates, whereas it should be checked for all other S/MIME certificates except these. For mailbox-validated certificates, the country attribute must not be present. This PR fixes the issue by ensuring the check is applied to the other certificate types and adds a new lint to verify the presence of certain attributes in mailbox-validated certificates.


christopher-henderson (approved):
Thank you @mtgag and Deutsche Telekom Security for the clean and reasonable lint!

PR #996, opened 2025-10-04
add requirements comment to e_ca_aia_non_http_url

mhyder13:
## Notes
I was fixing some citations in other lints (will be posted soon) and noticed that this one doesn't have a comment with the language that it's based on like most of them do and figured I'd offer one up in passing. The citation string in the metadata is still correct, and there's no real history to this requirement yet, so this is pretty straightforward. If you don't think this is useful, feel free to discard it.

## Doc Reference
[v2.1.7 §7.1.2.10.3](https://cabforum.org/working-groups/server/baseline-requirements/documents/CA-Browser-Forum-TLS-BR-2.1.7.pdf#page=130&zoom=100,92,320)


christopher-henderson (approved):
Thank you!

I, personally, appreciate touches such as this. The wide majority of the effort I have to put into this repo is double checking that the lint, as presented, is what is actually in the requirements. Having at least SOME of the original, verbatim, text is appreciated so that I can simply grep through documents. Otherwise, I am stuck internalizing entire sections (possibly across multiple versions) just to run a basic smoke check

PR #1002, opened 2025-11-02
Add some lints for OCSP Responder certificates

defacto64:
These that I propose here are some of the various lints that can (should) be developed to check that OCSP Responder certificates comply with CABF TLS BRs section 7.1.2.8 (OCSP Responder Certificate Profile).

There are no cases in the Zlint test corpus that trigger these lints, however some can be found on Censys, and in any case every now and then some incidents do happen related to defective OCSP Responder certificates (as recently seen on Bugzilla).


christopher-henderson (approved):
Very clean and easy, thank you!

And I see what you meant by "micro lints". It's often tempting to apply DRY and centralize logic, but sometimes the acrobatics just aren't worth it :shrug:

Issue #1003, opened 2025-11-12
Lint e_qcstatem_qctype_web throws an error for legitimate e-signature and e-seal qualified certificates


defacto64:
I believe this issue was discussed a few years ago, but remained unresolved, and I believe it needs to be addressed at this point.

The lint `e_qcstatem_qctype_web` is triggered by _any_ qualified certificate, even if it's an electronic signature or seal certificate, and an error is invariably thrown with the message "etsi Type does not indicate certificate as a 'web' certificate"

This is obviously incorrect, because Zlint doesn't assume that the certificate being linted is necessarily one for TLS server authentication.

The problem arises, to my understanding, because the `util.IsServerAuthCert()` utility function generates an incorrect `true` result for certificates that do NOT have the EKU extension but are clearly not intended for TLS server authentication.

What approach do you suggest, folks, to solve this problem?

defacto64:
Back to my point: let's take for example a qualified e-seal certificate: per ETSI EN 319 412-5 (section 5), it shall contain the `QcType` QcStatements specifying that it is an electronic seal certificate. So, for the very fact that the certificates contains such QcType, it is obviously NOT a 'web' certificate, and therefore it is absurd and misleading to complain about the lack if the 'web' QcType.

christopher-henderson:
Addressed in #1004

PR #1004, opened 2025-11-16
Lint e_qcstatem_qctype_web throws an error for legitimate e-signature and e-seal qualified certificates

christopher-henderson:
Addresses #1003 

The heart of the issue was that this particular lint was using `util.IsServerAuth` which plays quite fast-and-lose with its definition of what signifies a server auth certificate. Notably, it claims that a certificate with no EKUs at all is a server auth cert. This works well enough for CABF, but it falls on its face in this ETSI lint.

Note that this is the only ETSI lint that makes reference to `util.IsServerAuth`. So rather than add no `util` facilities it made sense to make the  most simplistic change via a simple `util.HasEku` call.

### DIFF
--- a/v3/lints/etsi/lint_qcstatem_qctype_web.go
+++ b/v3/lints/etsi/lint_qcstatem_qctype_web.go
@@ -49,7 +49,6 @@ func (l *qcStatemQctypeWeb) CheckApplies(c *x509.Certificate) bool {
 		return false
 	}
 	if util.ParseQcStatem(util.GetExtFromCert(c, util.QcStateOid).Value, *l.getStatementOid()).IsPresent() {
-		//return util.IsServerAuthCert(c)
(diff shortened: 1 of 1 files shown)


v3/lints/etsi/lint_qcstatem_qctype_web_test.go
```diff
@@ -28,7 +28,8 @@ func TestEtsiQcTypeWeb(t *testing.T) {
 		"QcStmtEtsiNoQcStatmentsCert22.pem": lint.NA,
 		"qcSmimeNatural.pem":                lint.NA,
 		"qcSmimeLegal.pem":                  lint.NA,
-		"qcLegal.pem":                       lint.Error,
+		"qcLegal.pem":                       lint.NA,
```
christopher-henderson: This one stopped reaching its desired error because it did not have the server auth EKU.

PR #1011, opened 2025-12-17
Add lint to check for a reserved policy identifier in S/MIME certificates

defacto64:
I've noticed that ZLint doesn't raise any errors when linting an S/MIME certificate which lacks a reserved policy identifier among those required by the S/MIME BRs (see section 7.1.2.3), which seems _counterintuitive_ to me. I am therefore proposing this trivial lint to resolve the issue.

(It could/should also be verified that there is only one such policy OID in the certificate, but I'll leave that to some other volunteer!)


christopher-henderson (changes requested):

christopher-henderson:
> but sometimes I get the impression that you don't much like lints that perform multiple checks at once, while other times it seems like you actually like them :)

Well in my own mental model this is not checking two things. `There must be exactly one reserved policy` is a singular clause that evaluates to one boolean expression. If this were `there must be exactly one reserved policy AND it must be....` then I would break it up.

> especially to create 22 test certificates. I plan to do so in the next few days, barring any impediments :)

This is why I wrote the function to accept a list of OIDs rather than certificates, it is much easier to unit test. I would be fine leaning on the unit test provided and keeping the current certs for a smoke check.

defacto64:
So, first I added your `ContainsExactlyOneSMIMEPolicy` function to the `smime_policies.go` source (package `util`), then I called this function in my lint. Then, in the test source of my lint, I integrated your test suite for the function itself, but I also added a certificate containing two CABF SMIME policy OIDs to the test data — for completeness. Finally, I changed the lint's name because "missing" would be misleading in some cases. How does it seem to you?

PR #1022, opened 2026-02-14
Add lint to check that the notBefore timestamp is not too early compared to the SCTs

defacto64:
Following a recent incident, as seen on bugzilla, I thought it would be useful to develop a simple lint that verifies compliance with §7.1.2.7 of the CABF TLS BRs, which requires that... 
>     ...notBefore [must be] a value within 48 hours of the certificate signing operation
In general, there is no way to determine, by just examining a certificate, when it was signed. However, since the certificate signing time cannot be earlier than the timestamps included in any SCTs, verification is possible when the certificate includes SCTs, as is done in this lint.


christopher-henderson (approved):
Clean and easy lint, thank you!

PR #1027, opened 2026-03-05
Fix IV countryName lint checkApplies, add IV personal name lint history

mhyder13:
## Summary
* Fixed a bug in `e_cert_policy_iv_requires_country` where IV-issuing policy constrained CAs were inadvertently linted
* Added unit tests to `e_cert_policy_iv_requires_country` to cover `checkApplies()` and effective date
* Updated the citation string in `e_cert_policy_iv_requires_country` to the current location
* Updated the policy language and added a citation history for `e_cert_policy_iv_requires_country`
* Added a citation history for `e_cab_iv_requires_personal_name`
* Minor cleanup of citation comment in `e_cab_iv_requires_personal_name_strict`

## Explanation
`e_cert_policy_iv_requires_country` was written to enforce the country requirement of Subscriber Certificates issued with Individual Validation. While it's true that Subordinate CAs are _also_ required to include the countryName in their subject, that requirement has always been separate from the requirement enforced and cited by this lint. In v1.3.1 (when IV was added) the requirement for IV was in §7.1.6.1 while the CA subject requirement was in §7.1.2.2h. In the current version v2.2.5, these are still separate as §7.1.2.7.3 and §7.1.2.10.2 respectively. Additionally, the other IV name lints already exclude CAs, which shows this was likely unintentional. Lastly, we already have an implementation of the CA requirement in `e_ca_country_name_missing`, so this lint is only adding an incorrect additional error when it triggers on CA certs.

It's not abnormal for Subordinate CA certs to assert the 2.23.140.1.2.3 IV policy. In the current version of the BRs, TLS Subordinate CAs are required to include the certificate policies extension, and Subordinates which are not operated by affiliates of the issuer must be "Policy Restricted." To be Policy Restricted, exactly one Reserved Certificate Policy, such as 2.23.140.1.2.3, must be included. The rules at the time of v1.3.1 did not explicitly require the use of a Reserved Certificate Policy and did not yet use the term "Policy Restricted," but it did require CAs issued to non-affiliates to include "one or more explicit policy identifiers that indicates the Subordinate CA’s adherence to and compliance with these Requirements" and allowed the use of "the CA/Browser Forum reserved identifiers" for this purpose. As such, this lint should never have applied to CAs because the appearance of this policy in a CA has a different meaning than in a Subscriber Certificate, and this has been true for the entire lifetime of this requirement. See v1.3.1 §7.1.6.3.1 and v2.2.5 §7.1.2.10.5.

#### v1.3.1
image

#### v2.2.5
image

## Doc History
All three lints follow the same section history, but `e_cab_iv_requires_personal_name` is replaced with the strict variant in v2.0.0.

#### v1.3.1
image

#### v1.7.3 (edit shown from redline)
image
image

#### [v2.2.5](https://cabforum.org/working-groups/server/baseline-requirements/documents/CA-Browser-Forum-TLS-BR-2.2.5.pdf#page=87&zoom=100,48,97) (link to section, screenshot only shows relevant table sections)
image
image
image

### DIFF
--- a/v3/lints/cabf_br/lint_arpa_domain_not_allowed.go
+++ b/v3/lints/cabf_br/lint_arpa_domain_not_allowed.go
@@ -1,0 +1,64 @@
+/*
+ * ZLint Copyright 2024 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
+ * of the License at http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package cabf_br
+
+import (
+	"github.com/zmap/zcrypto/x509"
+	"github.com/zmap/zlint/v3/lint"
+	"github.com/zmap/zlint/v3/util"
+
+	"strings"
+)
+
+func init() {
+
+	lint.RegisterCertificateLint(&lint.CertificateLint{
+		LintMetadata: lint.LintMetadata{
+			Name:          "e_arpa_domain_not_allowed",
+			Description:   "CAs SHALL NOT issue Certificates containing Domain Names that end in an IP Reverse Zone Suffix",
+			Citation:      "CABF TLS BRs section 4.2.2",
+			Source:        lint.CABFBaselineRequirements,
+			EffectiveDate: util.CABF_SC086_EffectiveDate,
+		},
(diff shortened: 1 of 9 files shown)


v3/lints/cabf_br/lint_cert_policy_iv_requires_country_test.go
```diff
@@ -38,3 +38,30 @@ func TestCertPolicyIvNoCountry(t *testing.T) {
 		t.Errorf("%s: expected %s, got %s", inputPath, expected, out.Status)
 	}
 }
+
```
christopher-henderson: This is perfectly fine, but just a heads up that this codebase leverages table driven testing quite extensively (plus the Golang ecosystem tends to prefer it as well).

mhyder13: Given that every test in this packages boils down to "run on this file and match this result" that makes sense. I was just following the style of the existing tests in this file, but I'll rewrite my next PR in that style if you prefer.

Issue #1016, opened 2026-01-28
Introducing support for additional browser-vendor-specific sources of lints


defacto64:
Since we already have the `Mozilla` and `Apple` lint sources in ZLint, I propose adding support for additional lint sources dedicated to other browser vendors (or "certificate consumers" if you prefer).  At the very least, I'd suggest supporting a `Google` source , given that some specific requirements on TLS certificates already exist within the Chrome Root Program Policy.
How about this?

christopher-henderson:
It would not be uncalled for.

I'm working in the background to setup a bit more automation in this repo to help us scrape and identify linting opportunities. This is the [Chrome Root Program Policy](https://googlechrome.github.io/chromerootprogram/) that you're referring, I presume?

defacto64:
Agree.

PR #1031, opened 2026-03-28
Add support for Chrome Root Program Policy-based lints, plus a first such lint addressing clientAuth deprecation

defacto64:
Following up on my proposal on https://github.com/zmap/zlint/issues/1016, which I understand was approved, this PR introduces what's needed to support Chrome Root Program Policy-based lints and includes a first lint based on the new `ChromeRootStorePolicy` lint source. Specifically, the lint proposed here verifies that TLS Server certificates issued after March 15, 2027, do not have clientAuth in their EKU, as required by the CRP [1]. This check is a bit ahead of its time, if you like, but I believe it won't hurt to introduce it now.

[1] https://googlechrome.github.io/chromerootprogram/#132-promote-use-of-dedicated-tls-server-authentication-pki-hierarchies


christopher-henderson (approved):
Thank you for your patience. My primary duties have been a living nightmare for the past month, completely monopolizing my life.

---

This looks good! Since it's been a long time since we have last time anyone has touched overall infrastructure I gave this a few extra smoke checks

* `chrome` is enumerated in the CLI help
* The `chrome` lint is reachable by the CLI
  * This includes include/exclude filtering.
* The library usage (actually the more common, I believe) is finding the new lint category just fine.

Honestly, I'm a bit surprised (pleasantly so) that it was as simple as adding a few extra enums and filesystem paths.

Thank you very much for this addition to ZLint!

PR #1034, opened 2026-04-07
Add serial number lint for code signing

digirenpeter:
Added a lint to check for serial number based on the latest CABF code signing requirements:
`CAs SHALL generate non-sequential Certificate serial numbers greater than zero (0) containing at least 64 bits of output from a CSPRNG.`


digirenpeter:
Per the conversation on this old PR:
https://github.com/zmap/zlint/pull/857

I'm closing this PR for now, I'll try to revisit it a bit later.

PR #1040, opened 2026-04-30
Add lint to check for NULL parameters in the AlgorithmIdentifier element of CRLs, when mandatory per RFC 4055

defacto64:
Suspicious of a recent incident, I noticed that Zlint lacks a certain check on the signature algorithm identifier of CRLs. Specifically, it lacks verification that — when the signature algorithm is one of the four `shaXXXWIthRSAEncryption` algorithms (with XXX being one of 224, 256, 384, 512) — the mandatory `parameters` element is present and set to NULL, as required by RFC 4055 Section 6 (which is referenced by RFC 5280). A similar check already exists for certificates, but not for CRLs, so here's my proposal. In my lint, I perform this check on the outer AlgorithmIdentifier. I think examining the inner or the outer one is equivalent, since the two are supposed be identical (and if they aren't, it's a different problem that IMO should be addressed by a separate lint).


v3/lints/rfc/lint_crl_sigalgo_missing_null_params.go
```diff
@@ -0,0 +1,128 @@
...
+
+	if bytes.Equal(algoBytes, sha224WithRSAEncryption) ||
+		bytes.Equal(algoBytes, sha256WithRSAEncryption) ||
+		bytes.Equal(algoBytes, sha384WithRSAEncryption) ||
+		bytes.Equal(algoBytes, sha512WithRSAEncryption) {
+
+		// Attempt to read the parameters element
+		var nullBytes cryptobyte.String
+		var nullFound bool
+		if !signatureAlgorithmBytes.ReadOptionalASN1(&nullBytes, &nullFound, asn1.NULL) {
```
christopher-henderson, line 106: -- Pointless Commentary --

`cryptobyte` is both very useful **_and_** very clearly a C library pretending to be Go library 😆 

I have to reckon that these are 1:1 translations, minus the fact these are methods on a `*String` rather than functions that take in a `*String`.

christopher-henderson (approved):
Thank you for this lint!

It's not easy when `cryptobyte` has to be whipped so that you can manually walk through the structure. So your work is very much appreciated.

PR #1048, opened 2026-05-22
Add telekom security

mtgag:
We would be grateful if you could add Deutsche Telekom Security GmbH to the list of zlint users, as they have been actively using it and have supported its development for many years.


christopher-henderson (approved):
Of course! And thank you to Deutsche Telekom Security for its support throughout the years.

PR #1069, opened 2026-07-28
Add lint to check for a valid commonName in Organization Validated S/MIME certificates

defacto64:
Section 7.1.4.2.2 (Subject Distinguished Name Fields) of the CABF S/MIME BRs requires that, for **Organization Validated** S/MIME certificates, the `commonName` attribute contain either an email address or the same value as the `organizationName` attribute. 

I realized that a lint for this requirement was still missing, so here is my proposal.

### DIFF
--- a/v3/lints/cabf_smime_br/lint_org_validated_invalid_cn.go
+++ b/v3/lints/cabf_smime_br/lint_org_validated_invalid_cn.go
@@ -1,0 +1,67 @@
+/*
+ * ZLint Copyright 2024 Regents of the University of Michigan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy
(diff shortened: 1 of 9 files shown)


christopher-henderson (approved):

robstradling:
@defacto64 @christopher-henderson This lint appears to require that the Subject CN field is always present (or if it's absent, then it behaves as if the CN is present as an empty string), which is problematic.  The S/MIME BRs are clear that this field is optional.

defacto64:
@robstradling I don't understand what you mean. This lint doesn't assume that the Subject CN is always present. And where the Subject CN is absent, which is very common, this lint doesn't return any errors.

robstradling:
@defacto64 Not sure how you tested it, but this PR really does return errors when the Subject CN is absent.  I've created #1071 to fix the problem.  (If you try linting the testcase I've added in #1071 using the code from #1069, you'll see what I mean).

PR #1070, opened 2026-08-01
Add lint to detect UTF8 Replacement Character in the Subject DN (for any kind of certificates)

defacto64:
Inspired by a recent incident, I put together this lint that checks for the UTF8 Replacement Character (Unicode 0xFFFD) in any Subject attribute, which signals that something went wrong when the CA assigned a value to that attribute. For example, a first name like Müller might end up in the givenName as "M\U+FFFDller" due to some Latin-1 to UTF8 conversion error. There are a couple of occurrences in the test corpus, in very old certificates, but there are likely more in the wild.

This lint scans all subject attributes that are actually tagged as UTF8String, regardless of whether this tag is legit for each particular attribute (it's not this lint's job to check that).


v3/lints/community/lint_utf8_replac_char_in_subj.go
```diff
@@ -0,0 +1,112 @@
...
+		return name
+	}
+	return "Subject attribute with OID " + oidStr
+}
+
+func (l *UTF8ReplacementCharInSubject) Execute(c *x509.Certificate) *lint.LintResult {
+
+	var rdnSeq []asn1.RawValue // RDNSequence ::= SEQUENCE OF RDN
+	if _, err := asn1.Unmarshal(c.RawSubject, &rdnSeq); err != nil {
+		panic(err)
```
christopher-henderson: Panics from lints are [caught-and-recovered](https://github.com/zmap/zlint/blob/2d3f694e393dc675f9830c5d02f7bd8d15a34785/v3/lint/base.go#L244). However, if you don't mind, I believe that should prefer propagating a `lint.Fatal` with the `Detail` being the error.

v3/lints/community/lint_utf8_replac_char_in_subj.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	for _, rdn := range rdnSeq {
+		var atvs []attributeTypeAndValue // RDN ::= SET OF AttributeTypeAndValue
+		if _, err := asn1.UnmarshalWithParams(rdn.FullBytes, &atvs, "set"); err != nil {
+			panic(err)
```
christopher-henderson: Same as above.

christopher-henderson (changes requested):
This is a cool lint! I only hold exception to the `panic`s.

v3/lints/community/lint_utf8_replac_char_in_subj.go
defacto64: Sure, done.

christopher-henderson (approved):

PR #1071, opened 2026-08-04
PR #1069 follow-up

robstradling:
The `lint_org_validated_invalid_cn` lint added by PR #1069 mishandles OV S/MIME certificates that omit the Subject CN field.  Instead of treating the lint as not applicable due to the absent CN, it treats the empty string as an invalid email address and a mismatch for the Subject O.

This PR fixes the problem and adds a suitable test case.


robstradling:
I also thought about how to handle Subject CNs where the attribute is present but its value is zero-length.  It seems that the `e_subject_contains_noninformational_value` lint already handles this, meaning that `lint_org_validated_invalid_cn` doesn't need to.

defacto64:
@robstradling You are right: I forgot to consider the case when the Subject CN is absent, which is allowed. As to the case when it is empty, that's a different problem. At any rate, I am fine with your proposed adjustments.

christopher-henderson (approved):

PR #1083, opened 2026-08-26
Improve robustness of the "e_org_validated_invalid_cn" lint for S/MIME OV certificates

defacto64:
Added a check for the organizationName to actually appear in the certificate Subject, which is mandatory in S/MIME OV certs but shouldn't be taken for granted. This will prevent a panic when referencing element #0 of the `c.Subject.Organization` array. Thanks to Wayne for suggesting this; I missed it.


defacto64:
I am not clear why the unit test fails...

PR #1084, opened 2026-08-31
Add lint to check for suspicious characters in localityName and/or stateOrProvinceName

defacto64:
After a recent incident seen on Bugzilla, I thought this lint would be good to have. This lint checks whether the Subject's localityName and stateOrProvinceName attributes contain suspicious characters such as comma, colon, parentheses, etc. The presence of such characters almost certainly indicates an incorrect value for the containing attribute. A very common case is when localityName contains two or even three address elements, separated by commas, which is obviously incorrect. For example, there are quite a few certificates in the test corpus where localityName contains not only the locality but also the street address or the state/province. But there are also stranger cases.

Since I am not sure that the presence of such characters is _always_ an error, this lint only returns a warning upon a hit. But if people here believe it's better, it's fine with me to change it to an error. I refined the set of suspect characters (as currently used in this lint) heuristically, so I may have missed some other characters that would be worth including; I'm of course open to changes.


christopher-henderson:
@defacto64 so this is clearly a work in progress with some question marks and @TODOs in places such as the description.

So rather than try to ideate a back-and-forth, I whipped up this [other PR](https://github.com/zmap/zlint/pull/1086) to take my own crack at it. Please give it a peek to see if you agree with it.

defacto64:
That's fine with me, so I am closing this PR.

PR #1086, opened 2026-09-06
Add n_suspicious_chars_in_address lint

christopher-henderson:
Adds `n_suspicious_chars_in_address`, which checks whether Subject:localityName or Subject:stateOrProvinceName contain characters (comma, colon, equals sign, quote, or parenthesis) suggesting the attribute improperly concatenates multiple address components, such as a locality plus a street address (a Community heuristic informed by real-world incident reports; not codified in a formal specification, hence `Notice` rather than `Warn`/`Error`).


defacto64:
I am OK with this.

PR #1080, opened 2026-08-10
feat: Confirm JoiC matches organizationIdentifier Country / cabfOrganizationIdentifierCountry

XolphinMartijn:
In response to https://bugzilla.mozilla.org/show_bug.cgi?id=2062202, this lint should catch any such potential misissuance prior to certificate issuance.

PR #1089, opened 2026-09-11
Add lint to check for consistent jurisdictionCountry and organizationIdentifier (NTR scheme) in EV certificates

defacto64:
When an EV certificate includes the `Subject::organizationIdentifier` attribute carrying a Registration Reference with the "NTR" scheme (see Appendix H in the CABF EVGL), then...

> ...the country code used in the Registration Scheme identifier shall match that of the subject’s jurisdiction as specified in Section 7.1.4.2.4." 

This lint checks that this requirement is met, returning an error otherwise. 

There are no hits in the test corpus, but a case was recently reported on Bugzilla.


XolphinMartijn:
@defacto64 I believe this is covered by the recently added https://github.com/zmap/zlint/pull/1080 ?

defacto64:
No @XolphinMartijn , nothing to do with that.

defacto64:
https://github.com/zmap/zlint/pull/1080 deals with the `cabfOrganizationIdentifier `extension, while this lint checks the `jurisdictionCountry `with the Subject's `organizationIdentifier `attribute (in the specific case of the NTR scheme) according to Appendix H of the EVGL.

XolphinMartijn:
@defacto64  You are correct that 1080 deals with the cabfOrganizationIdentifier, however, this also incorporated the organizationIdentifier already. Specifically, the linter in 1080 combined with lint_ev_orgid_inconsistent_subj_and_ext, which has existed for a while, already make sure that an error is thrown if there is an error between the values

defacto64:
@XolphinMartijn I should look into what you're saying in more detail. However, as it stands, ZLint does not currently detect any issues in a certificate containing inconsistent `orgId` and `joiCountry` values—such as the one I’ve attached here.
[incons_orgid_joic_example.txt](https://github.com/user-attachments/files/32101895/incons_orgid_joic_example.txt)

CBonnell:
While I remember that there are some rare scenarios where a jurisC can legitimately be different from the orgId's attribute value under the NTR scheme, this lint is probably fine for the vast majority of cases.

The lint in this PR takes the right approach by restricting the check to only NTR scheme identifiers. The approach taken in #1080 is flawed in that non-NTR orgIds will be incorrectly flagged. For example, it is allowed to use EL as the country code for VAT scheme, but the jurisC must still be GR. The current logic in #1080 will incorrectly flag this as a mismatch.

PR #1090, opened 2026-09-18
Prefer CRL IDP scope over SubscriberCRL config in e_crl_next_update_invalid

MincongZhou:
## Background

`e_crl_next_update_invalid` (added in #916) enforces the CABF BRs 4.9.7 limit on
the gap between `thisUpdate` and `nextUpdate`. The BRs set two different limits:
10 days for CRLs covering subscriber certificates, and 12 months for CRLs
covering CA certificates.

As noted in #916, an arbitrary CRL does not say which of the two applies, so the
lint falls back to the `SubscriberCRL` config option (default `true`, i.e. the
10-day limit). The practical effect is that CA CRLs — legitimately allowed up to
12 months — are reported as errors. Other consumers hit this too; e.g.
letsencrypt/boulder#8368 had to grow the ability to skip this lint.

## Why this is worth doing

The project that reported the problem also emits the in-band signal this PR uses.
Boulder (the CA software behind Let's Encrypt) attaches an Issuing Distribution
Point to its CRLs through its `crl/idp` package:

* `idp.MakeUserCertsExt(...)` asserts `onlyContainsUserCerts = TRUE`
* `idp.MakeCACertsExt()` asserts `onlyContainsCACerts = TRUE`

Both are critical, and they exist precisely to separate the subscriber CRL from
the CA CRL. With this change the lint picks the correct limit for those CRLs on
its own, instead of depending on the `SubscriberCRL` option or on a downstream
skip-lint mechanism.

## Change

RFC 5280 Section 5.2.5 provides the signal: when a CRL carries an Issuing
Distribution Point, `onlyContainsUserCerts` / `onlyContainsCACerts` state exactly
which kind of certificate the CRL covers.

This PR consults that extension first:

* `onlyContainsCACerts = TRUE`   -> 12-month limit
* `onlyContainsUserCerts = TRUE` -> 10-day limit
* no IDP / unparseable IDP / neither bit set -> fall back to `SubscriberCRL`

The IDP is only consulted when present, so existing users and CRLs without an IDP
see no behavior change. Note that the IDP is optional and not every CA emits it;
for those CRLs the `SubscriberCRL` option remains the way to select the limit.

## Tests

* `crl_nextupdate_idp_ca_eff1.pem` – IDP scoped to CA, 100-day span -> Pass
* `crl_nextupdate_idp_ca_beyond12m_eff1.pem` – IDP scoped to CA, >12-month span -> Error
* `crl_nextupdate_idp_ee_eff1.pem` – IDP scoped to subscriber, 100-day span -> Error
* `crl_nextupdate_idp_ee_eff1.pem` + `SubscriberCRL = false` – IDP wins -> Error


v3/lints/cabf_br/lint_crl_next_update_invalid.go
```diff
@@ -82,3 +97,61 @@ func (l *CrlNextUpdateInvalid) Execute(c *x509.RevocationList) *lint.LintResult
...
+// RFC 5280, Section 5.2.5:
+//
+//	IssuingDistributionPoint ::= SEQUENCE {
+//	     distributionPoint          [0] DistributionPointName OPTIONAL,
+//	     onlyContainsUserCerts      [1] BOOLEAN DEFAULT FALSE,
+//	     onlyContainsCACerts        [2] BOOLEAN DEFAULT FALSE,
+//	     onlySomeReasons            [3] ReasonFlags OPTIONAL,
+//	     indirectCRL                [4] BOOLEAN DEFAULT FALSE,
+//	     onlyContainsAttributeCerts [5] BOOLEAN DEFAULT FALSE }
+func idpCRLScope(c *x509.RevocationList) (coversCA bool, ok bool) {
```
christopher-henderson: There is a much easier, and correct, way to do this by leveraging defining the target structure decoratively rather than attempting to walk it manually.

Something like this.

```go
var errAmbiguousIDPScope = errors.New("IDP does not unambiguously scope the CRL")

// IssuingDistributionPoint ::= SEQUENCE {
//      distributionPoint          [0] DistributionPointName OPTIONAL,
//      onlyContainsUserCerts      [1] BOOLEAN DEFAULT FALSE,
//      onlyContainsCACerts        [2] BOOLEAN DEFAULT FALSE,
//      onlySomeReasons            [3] ReasonFlags OPTIONAL,
//      indirectCRL                [4] BOOLEAN DEFAULT FALSE,
//      onlyContainsAttributeCerts [5] BOOLEAN DEFAULT FALSE }
type issuingDistPoint struct {
	DistributionPoint     asn1.RawValue  `asn1:"optional,tag:0"`
	OnlyContainsUserCerts bool           `asn1:"optional,tag:1"`
	OnlyContainsCACerts   bool           `asn1:"optional,tag:2"`
	OnlySomeReasons       asn1.BitString `asn1:"optional,tag:3"`
	IndirectCRL           bool           `asn1:"optional,tag:4"`
	OnlyContainsAttrCerts bool           `asn1:"optional,tag:5"`
}

type crlScope int

const (
	crlScopeUnknown crlScope = iota
	crlScopeSubscriber
	crlScopeCA
)

// idpCRLScope inspects the CRL's Issuing Distribution Point extension, if
// present, and reports whether it unambiguously scopes the CRL to CA or
// subscriber certificates. It returns (crlScopeUnknown, nil) when there is
// no IDP extension, or when it's present but doesn't assert either scope
// boolean. It returns a non-nil error when the extension can't be parsed,
// or when it asserts both scope booleans (which BR §7.2.2.1 prohibits).
func idpCRLScope(c *x509.RevocationList) (crlScope, error) {
	for _, ext := range c.Extensions {
		if !ext.Id.Equal(util.IssuingDistOID) {
			continue
		}

		var idp issuingDistPoint
		if _, err := asn1.Unmarshal(ext.Value, &idp); err != nil {
			return crlScopeUnknown, err
		}

		switch {
		case idp.OnlyContainsCACerts && idp.OnlyContainsUserCerts:
			return crlScopeUnknown, errAmbiguousIDPScope
		case idp.OnlyContainsCACerts:
			return crlScopeCA, nil
		case idp.OnlyContainsUserCerts:
			return crlScopeSubscriber, nil
		default:
			return crlScopeUnknown, nil
		}
	}
	return crlScopeUnknown, nil
}
```

christopher-henderson (changes requested):

MincongZhou:
Agreed, declaring the structure is the right call. Pushed `ef5ba39`:

- `issuingDistPoint` is now a struct with `optional,tag:N` fields parsed by
  `asn1.Unmarshal`; the manual walk is gone.
- `idpCRLScope` returns `(crlScope, error)`: `crlScopeCA`, `crlScopeSubscriber` or
  `crlScopeUnknown`.
- A parse failure, or both scope bits set, now becomes `lint.Error` instead of
  falling back to `SubscriberCRL`. BR 7.2.2.1 forbids both bits, and the IDP is
  critical when present, so the fallback would hide a broken extension. Say so if
  you'd rather have the fallback and I'll switch it.
- An IDP with neither bit set (present only to carry a distribution point) still
  falls back to `SubscriberCRL`.

`TestIDPCRLScope` marshals the IDP in the test itself, which covers both-bits and
unparseable IDPs that openssl will not produce as fixtures. The existing
`crl_nextupdate_idp_*` cases still pass. `gofmt` / `go vet` / `go test ./...` clean
on top of `5ffe609`.

